PR #1, opened 2021-02-11
Added basecase for "other" component

JanMaartenvanDoorn:
Added basecase for "other" component

### DIFF
--- a/stf/model/prediction/prediction.py
+++ b/stf/model/prediction/prediction.py
@@ -117,9 +117,9 @@ class AbstractPredictionModel(ABC):
         # rename
         basecase_forecast = basecase_forecast.rename(columns=dict(load="forecast"))
 
-        # Also make basecase for component "other". We do this to make sure the sum of
-        # the basecase of components equals the basecase of the net load
-        # (basecase of wind and pv components is zero)
+        # Also make a basecase forecast for the forecast_other component. This will make a
+        # simple basecase components forecast available and ensures that the sum of
+        # the components (other, wind and solar) is equal to the normal basecase forecast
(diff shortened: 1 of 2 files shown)


stf/model/prediction/prediction.py
```diff
@@ -116,6 +116,12 @@ def make_basecase_forecast(pj, historic_load, overwrite_delay_hours=48):
         basecase_forecast.dropna(inplace=True)
         # rename
         basecase_forecast = basecase_forecast.rename(columns=dict(load="forecast"))
+
+        # Also make basecase for component "other". We do this to make sure the sum of
+        # the basecase of components equals the basecase of the net load
+        # (basecase of wind and pv components is zero)
+        basecase_forecast['forecast_other'] = basecase_forecast['forecast']
```
bremme, line 123: Good to add a comment here! Perhaps you can rephrase this a little bit to make it more clear perhaps. What do you think about something along the line of:

```python
# Also make a basecase forecast for the forecast_other component. This will make a
# simple basecase components forecast available and ensures that the sum of 
# the components (other, wind and solar) is equal to the normal basecase forecast
```

bremme:
Looks good! Nice to see that we only need to make a very small change to have a correct medium term components forecast. I added one comment with a suggestion to rephrase your comment a bit perhaps.

bremme:
If we want to use the feature branches for the changelog it would help if you would use a more descriptive name like you actually did in the title:

* title of PR: Added basecase for "other" component
* branch name: `feature-add-basecase-for-other-component`

By doing so we could automate the changelog generation and it would look like this:

### Changelog
* Add basecase for other component -> (replace `-` with ` `, capitalize first word)

PR #2, opened 2021-02-12
Feature KTP-1547 fix load project config from package root

bremme:
This feature is only about a small required change to fix loading of the configuration. The only things that has changed to make this works is: `stf/tasks/utils/taskcontext.py`.

When we use our `ktpbase` implementation to load the configuration we needed to know the base/root path of the project. This logic has been taken out and moved to our own config repo (`icarus-forecasts`)

Furthermore (I should have switched branch, sorry for that). I made a start with the Github Actions CI/CD. But more importantly I made various fixes to the unit tests (which were/are broken). Perhaps you can mainly have a look a `test_more_train.py` since I rewrote that test. Does it still cover the same ground?

Gonna switch branch now before continuing working on other features :)


test/unit/model/test_more_train.py
```diff
@@ -53,49 +54,62 @@
...
+model, model_file = serializer.load(123, TestData.TRAINED_MODELS_FOLDER)
+model_path = TestData.TRAINED_MODELS_FOLDER / "<pid>/20191119120000/model.bin"
+
+
+def get_serializer_mock():
+    serializer_mock = MagicMock()
+    serializer_mock.load.return_value = model, model_path
+
+
+# @patch("stf.model.serializer.creator.XGBModelSerializer", get_serializer_mock())
```
JanMaartenvanDoorn, line 66: Is this commented out code functional or can we remove it?

stf/model/trainer/xgboost/xgboost.py
```diff
@@ -191,7 +192,7 @@ def better_than_old_model(self, test_data, penalty_factor=1.2):
                     ntree_limit=self.trained_model.best_ntree_limit,
                 )
             except Exception as e:
-                print("Could not get prediction from new model:", str(e)[:20])
+                self.logger.error("Could not get prediction from new model:", str(e)[:20])
                 return False
```
JanMaartenvanDoorn, line 197: Good idea to use the logger for this!

JanMaartenvanDoorn (changes requested):
Nice improvements, one question about commented out code which I think we can remove.

test/unit/model/test_more_train.py
bremme: Good point. I have already removed this on a newer branch.

bremme: To be sure, I removed the comment on this branch as well.

JanMaartenvanDoorn (approved):

PR #4, opened 2021-02-15
Feature KTP-1548 setup GitHub actions CI/CD

bremme:
I have implemented a basis CI/CD workflow using Github Actions. The following steps will be performed:

* Checkout code
* Setup Python (We can easily test with different Python versions if we want)
* Restore pip cache (speed up the whole process if dependencies have not change I think?)
* Install dependencies
* Lint code with flake8
* Test code with pytest

I have updates our `setup.cfg` file to to set a maximum complexity (10 at present) and some errors that we will ignore (for now). The following errors will be ignored:

```shell
# E501    line too long (made an issue to fix this)
# W292    no newline at end of file
# W391    blank line at end of file
# W503    line break before binary operator
# W504    line break after binary operator
```

In order to get here I had to fix many errors. Can you have a look at the following modules which had a too high complexity:

* stf.model.general -> combine_forecasts
* stf.feature_engineering.weather
* stf.tasks.utils.predictionjobloop

Futhermore I also had to change quite a bit to get the unit tests going again. The solution I came up with is the following. I added a dummy `ktpbase` package with just empty files and folders.  Futhermore I patched the logger and config manager on the ` BaseTestCase` level. So every unit test will have those things patched. Is this clear the way I implemented this or should I change something?

image


ktpbase/config/config.py
```diff
@@ -0,0 +1,2 @@
+class ConfigManager:
```
JanMaartenvanDoorn, line 1: Maybe add a comment to these files that they are nescesarry at the moment for unit tests.

setup.cfg
```diff
@@ -6,4 +6,13 @@
 max_line_length = 88
 [flake8]
 max_line_length = 88
-max_complexity = 10
\ No newline at end of file
+max_complexity = 10
+ignore = E501,W292,W391,W503,W504
+
+# E501    line too long
```
JanMaartenvanDoorn, line 12: Nicely documented!

stf/data_validation/data_validation.py
```diff
@@ -31,7 +31,7 @@ def clean(data):
     len_original = len(data)
     # TODO Look into this
     # Remove where load is NA # # df.dropna?
-    data = data.loc[np.isnan(data.iloc[:, 0]) != True, :]
+    data = data.loc[np.isnan(data.iloc[:, 0]) != True, :]   # noqa E712
```
JanMaartenvanDoorn, line 34: What does # noqa E712 mean?

stf/tasks/utils/predictionjobloop.py
```diff
@@ -12,9 +12,9 @@ def __init__(
             context,
             stop_on_exception=False,
             random_order=True,
-            on_exception=None,
-            on_successful=None,
-            on_end=None,
+            on_exception_cb=None,
```
JanMaartenvanDoorn, line 15: what does "_cb" mean?

test/utils/base.py
```diff
@@ -3,13 +3,27 @@
...
+        for patcher in self.patchers:
+            patcher.start()
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        for patcher in self.patchers:
+            patcher.stop()
+
     # TODO for some reason this does not work : has not attribute addTypeEqualitty
     # def __init__(self, *args, **kwargs):
```
JanMaartenvanDoorn, line 28: Can we remove this commented out code?

JanMaartenvanDoorn (commented):
Nice additions, few things that are unclear to me.

stf/tasks/utils/predictionjobloop.py
bremme: cb is an abbreviation of callback, Should I use `on_exception_callback`  instead to make it even more clear?

stf/data_validation/data_validation.py
bremme: This make flake8 ignore the E712 violation here. In this case we need to use `!= True` and we can't use ` is not True` which is normally the preferred option.

image

setup.cfg
bremme: thanks. I always forget some the error code, so this makes it a bit easier to find the meaning of the codes.

ktpbase/config/config.py
bremme: Good point. This should be solved by #a1978c5c487b6f603f9b0b8916bcb4f7e57e0905

test/utils/base.py
bremme: I used an old depreciated method, should have used `addTypeEqualityFunc()`  I have fixes this now so no commented code anymore, good point! See 5b287377ee04b280600dbc491ed44f044ecdf4e3.

stf/tasks/utils/predictionjobloop.py
JanMaartenvanDoorn: ah, yes `on_exception_callback` is more clear.

bremme: 8cd32c13b5672dcc77497f0c54968bdad0ce8b0f fixed.

PR #11, opened 2021-02-16
Feature KTP-1552 build PyPi package

bremme:
I've added a very basic workflow to build a PyPi published Python package.

* Made an account at PyPi using our korte.termijn.prognoses@alliander.com email adres (credentials are in our password manager)
* Added the PyPi credentials as project secrets
* Added a `python-upload-package.yaml`  workflow
* Minor tweaks to `setup.py`

The workflow should be triggered if we make a manual release. Perhaps it would be nice to always release tagged master commits. What do you think is this good enough for now or should I automate more stuff?

Another discussion is if we want to make `alpha`  (feature branches) and or `beta`  (develop branch) releases?

> **NOTE:** If have not tested this all since a release has to be made from the `main` branch. So once this PR gets accepted I can start testing this.


FrankKr (approved):
Looks good! Looking forward to see the first publication!

PR #17, opened 2021-02-17
Fix teams dependency by checking if a valid url is given

bremme:
Solves the issues in the most basic way possible. Just check if a Teams url is configured and if it's not don't try to send messages to Teams but log an warning instead.


FrankKr (commented):
Looks good, nice comments as well +1

PR #18, opened 2021-02-17
Made holiday features more universal

JanMaartenvanDoorn:


bremme:
Just a note. It seems you branched of `main` for this feature instead of `develop`. That's (one) of the reasons the CI/CD fails.

stf/feature_engineering/feature_free_days.py
```diff
@@ -6,11 +6,13 @@
 
 import numpy as np
 import pandas as pd
+from datetime import date
+import holidays
 
 HOLIDAY_CSV_PATH = os.path.dirname(__file__) + "/dutch_holidays_2020-2022.csv"
```
bremme, line 12: This is not used anymore.

stf/feature_engineering/feature_free_days.py
```diff
@@ -6,11 +6,13 @@
 
 import numpy as np
 import pandas as pd
+from datetime import date
+import holidays
```
bremme, line 10: `datetime` is a build in package so should precede the third-party imports.

stf/feature_engineering/feature_free_days.py
```diff
@@ -6,11 +6,13 @@
...
 
 HOLIDAY_CSV_PATH = os.path.dirname(__file__) + "/dutch_holidays_2020-2022.csv"
 
 
-def create_holiday_functions():
+def create_holiday_functions(country="FR", years=None):
```
bremme, line 15: Probably temporarily, but I don't think it makes sense to use one country as default.

stf/feature_engineering/feature_free_days.py
```diff
@@ -6,11 +6,13 @@
...
     """
     This function provides functions for creating holiday feature.
     This improves forecast accuracy. Examples of features that are added are:
```
bremme, line 18: Still work in progress, but don't forget to document the arguments. Probably nice to refer to the country code standard that is used and so on.

stf/feature_engineering/feature_free_days.py
```diff
@@ -44,27 +46,31 @@ def create_holiday_functions():
                 consist of "Is" + the_name_of_the_holiday_to_be_checked
     """
 
-    # Manully generated csv including all dutch holidays for different regions
-    df_holidays = pd.read_csv(HOLIDAY_CSV_PATH, index_col=None)
-    df_holidays["datum"] = pd.to_datetime(df_holidays.datum).apply(lambda x: x.date())
+    if years is None:
+        years = [2020, 2021]
```
bremme, line 50: Don't use hard coded years here. If you think we need a default option its probably better to use the current year. Something along the line of:

```python
if years is None:
  now = datetime.now()
  years = [now.year, now.year + 1]
```

stf/feature_engineering/feature_free_days.py
```diff
@@ -6,11 +6,13 @@
```
bremme, line 6: Wouldn't it make more sense to call this module just `create_holiday_functions` or `holiday_function`. Perhaps it nice to be a bit more consistent with naming.

stf/feature_engineering/feature_free_days.py
bremme: I think actually both `date` and `pd` are not used anymore.

stf/feature_engineering/feature_free_days.py
```diff
@@ -44,27 +46,31 @@ def create_holiday_functions():
                 consist of "Is" + the_name_of_the_holiday_to_be_checked
     """
 
-    # Manully generated csv including all dutch holidays for different regions
-    df_holidays = pd.read_csv(HOLIDAY_CSV_PATH, index_col=None)
-    df_holidays["datum"] = pd.to_datetime(df_holidays.datum).apply(lambda x: x.date())
+    if years is None:
+        years = [2020, 2021]
+
+    holidays_object = holidays.CountryHoliday(country, years=years)
```
bremme, line 52: Would `country_holidays` not be more suitable here. Its just a regular dict with the dates and names of the holidays for a certain country.

I think in general it makes more sense to name things according to which data they contain, not how that data is stored. Also in Python everything is an object of course:)

stf/feature_engineering/feature_free_days.py
```diff
@@ -44,27 +46,31 @@ def create_holiday_functions():
...
     holiday_functions.update(
-        {"IsFeestdag": lambda x: np.isin(x.index.date, df_holidays.datum.values)}
+        {"IsHoliday": lambda x: np.isin(x.index.py_datetime().date, holidays)}
     )
 
     # Loop over list of holidays names
-    for holidayname in list(set(df_holidays.name)):
+    for date, holidayname in sorted(holidays_object.items()):
         # Define function explicitely to mitigate 'late binding' problem
         def make_holiday_func(holidayname=holidayname):
```
bremme, line 64: It looks like the `holidayname` argument is not used in this function.

stf/feature_engineering/feature_free_days.py
```diff
@@ -44,27 +46,31 @@ def create_holiday_functions():
...
     holiday_functions = {}
     # Add check function that includes all holidays of the provided csv
     holiday_functions.update(
-        {"IsFeestdag": lambda x: np.isin(x.index.date, df_holidays.datum.values)}
+        {"IsHoliday": lambda x: np.isin(x.index.py_datetime().date, holidays)}
     )
 
     # Loop over list of holidays names
-    for holidayname in list(set(df_holidays.name)):
+    for date, holidayname in sorted(holidays_object.items()):
```
bremme, line 62: Would `holiday_name` not be more readable?

stf/feature_engineering/feature_free_days.py
bremme, line 64: Why is this function definition in the loop? It does not need to be redefined every iteration I guess?

stf/feature_engineering/feature_free_days.py
```diff
@@ -44,27 +46,31 @@ def create_holiday_functions():
...
         # Define function explicitely to mitigate 'late binding' problem
         def make_holiday_func(holidayname=holidayname):
-            return lambda x: np.isin(
-                x.index.date, df_holidays.datum[df_holidays.name == holidayname].values
-            )
+            return lambda x: x.index.py_datetime().date is date
 
         # Create lag function for each holiday
         holiday_functions.update(
             {"Is" + holidayname: make_holiday_func(holidayname=holidayname)}
```
bremme, line 69: I prefer the `snake_case` convention for Python. But that's too much of a hassle right now. I would at least stick to the `CamelCase` convention otherwise. You probably need to format the holiday feature name properly, for example:

```python
# First capatilize every word and next remove space
# "Eerste paasdag" becomes "IsEerstePaasdag"
feature_name = f'Is{holiday_name.title().replace(" ","")}'
holiday_function[feature_name] = make_holiday_function()
```

stf/feature_engineering/feature_free_days.py
```diff
@@ -44,27 +46,31 @@ def create_holiday_functions():
...
-    for holidayname in list(set(df_holidays.name)):
+    for date, holidayname in sorted(holidays_object.items()):
         # Define function explicitely to mitigate 'late binding' problem
         def make_holiday_func(holidayname=holidayname):
-            return lambda x: np.isin(
-                x.index.date, df_holidays.datum[df_holidays.name == holidayname].values
-            )
+            return lambda x: x.index.py_datetime().date is date
 
         # Create lag function for each holiday
```
bremme, line 67: You don't create "lag" function here do you?

bremme:
Since I have a day off tomorrow I already had a quick loop on this PR. I like the idea of using a package which works for various countries. But it seems very limited with only the public holidays.

If we go this route we should probably have a `create_public_holiday_functions()` and a `create_school_holidays_functions()`. I'm also not totally sure if this is data retrieval (ktpbase / Database) stuff or feature engineering. I guess its a bit of both. We should try not to mix these responsibilities if possible. Not sure how big of a deal it is in this case.

stf/feature_engineering/feature_free_days.py
```diff
@@ -43,27 +48,83 @@ def create_holiday_functions():
...
+                (date + timedelta(days=2)).weekday() == 5:
+
+            # If tomorow is not a national holiday or a weekend day make it a bridgeday
+            if not (date + timedelta(days=1)) in country_holidays \
+                    and not (date + timedelta(days=1)).weekday() in [5, 6]:
+
+                # Create feature function for each holiday
+                holiday_functions.update(
+                    {"is_bridgeday" + holiday_name.replace(" ", "_").lower(): make_holiday_func((date + timedelta(days=1)))}
+                )
```
FrankKr, line 88: Bridgedays: Nice addition 
Can 'bridgeday' without the holidayname be a more significant feature? 

We also have the generic 'is_holiday' feature. Can we add 'is_bridgeday' as well? (or perhaps even say that is_holiday=True for bridgedays)

FrankKr (commented):
Bridgedays: Nice addition 
Can 'bridgeday' without the holidayname be a more significant feature? 

We also have the generic 'is_holiday' feature. Can we add 'is_bridgeday' as well? (or perhaps even say that is_holiday=True for bridgedays)

JanMaartenvanDoorn:
> Bridgedays: Nice addition
> Can 'bridgeday' without the holidayname be a more significant feature?
> 
> We also have the generic 'is_holiday' feature. Can we add 'is_bridgeday' as well? (or perhaps even say that is_holiday=True for bridgedays)

I now added a general 'is_bridgeday' feature

bremme:
@JanMaartenvanDoorn can you resolve or comment on my comments? That makes it more clear which point you have addressed and which you didn't.

bremme:
I know this is almost done, but every PR should have a clear description. I also notice that 3 new code smells where introduced, you are probably working on them, but these should be fixed before merging.

FrankKr (approved):

Issue #25, opened 2021-02-19
auto-black [Feature] 


FrankKr:
**Is your feature request related to a problem? Please describe.**
The Flake8 check in the ci/cd pipeline was more strict than we liked. Now it has been (temporarily) been disabled. 
A nice option would be that for every PR, a github action applies black-formatting and generates a new commit for this automatically. 

**Describe the solution you'd like**
Implement a github action which runs the black code formatter for every PR and adds a commit to make it compliant.

**Describe alternatives you've considered**
Teach everyone how to write perfectly linted code

FrankKr:
After discussing with JM, do we want to use Black, or Autopep8. It's equally easy to implement.
https://github.com/marketplace/actions/autopep8

@bremme let's discuss next week.

PR #26, opened 2021-02-19
Feature #25 auto black

FrankKr:
automatically 'blacken' every PR. 
If PR is already compliant, this does nothing, otherwise, it generates an additional commit to the PR


FrankKr:
iironically, the build fails because even after blackify'ing the repo, flake8 isn't happy.

FrankKr:
Proposed solution: merging the other PR's to develop where this is dealt with, then merging develop in this PR.

bremme:
I'm not totally convinced on the need for such a tool. But I do think its a waste of time and energy to keep discussing syntax and formatting errors. If we think sticking to PEP8 by our self (using local linting / formatter) is too hard this might be a viable alternative.

Perhaps we should just give it a try. I guess the three most well known Python formatters are: `autopep8`, `black`  and `yapf`. `autopep8`  is very basic, `black` is advanced but very opinionated and non configurable while `yapf` is advanced and very configurable. To get started quickly and still have advanced features `black` is the best choice. I guess there are only a couple things I dislike about `black`:

1. if arguments / list, tuple elements etc. don't fit on a single line `black` uses one element per line
  1. For example a 30 element list becomes 30 lines, an 8 argument function signature becomes 8 lines etc.
2. new lines before binairy operators (I like them after)
3. white space before `:`

Especially 1. can make files really long. I'm not sure how we should deal with this. I guess it might be a code smell indication if it happens to function/method signatures.

I made a couple changes on this branch to finetune this integration:

* Added (almost) non restrictive Flake8 linting step
 * Only fails on very serious errors (things that would cause runtime errors for example)
 * Check and warn on everything else
*  Update flake8 configuration to not warn on black formatting in setup.cfg
* Make workflows a bit more consistent

What do you think? Good additions?

FrankKr:
@bremme we can also use yapf if that is preferred. Implementation-wise it's almost identical: https://github.com/diegovalenzuelaiturra/yapf-action

bremme:
Just for reference. Decided to test how well `black` works for now. We might switch to `yapf` in the future if we need more customization.

PR #27, opened 2021-02-19
Patch to fix reuse compliance

Jonasvdbo:
This patch suggest a fix for the reuse compliance in the main branch that broke due to https://github.com/alliander-opensource/short-term-forecasting/pull/22


FrankKr (approved):
Looks good, I'll add the additions!

PR #34, opened 2021-02-26
Feature ktps 461 sanity check energy split coefficients

gfsdekoning:
I added the sanity check for coefficients for energy splitting with variable threshold `COEF_MAX_PCT_DIFF`. In case the check fails, a message is sent to Teams with a manual insert sql statement.

I tried to keep the Alliander-specific code in `teams.py` as much as possible.

### DIFF
--- a/openstf/model/split_energy.py
+++ b/openstf/model/split_energy.py
@@ -53,13 +53,9 @@ def split_energy(pid):
     if not invalid_coefs.empty:
         # If coefs not valid, do not update the coefs in the db and send teams
         # message that something strange is happening
-        df_to_str = [
-            f"\ncoefficient: {row.coef_name}, new coefficient: {row.coef_value_new}, last coefficient: {row.coef_value_last}"
-            for row in invalid_coefs.iterrows()
-        ]
         monitoring.post_teams_alert(
-            f"New splitting coefficient(s) {list(invalid_coefs.coef_name)} for pid {pj['id']} deviate strongly from previously stored coefficients."
-            + df_to_str,
+            f"New splitting coefficient(s) {list(invalid_coefs.coef_name)} for pid {pj['id']} deviate strongly from previously stored coefficients.",
+            invalid_coefs=invalid_coefs,
             coefsdf=coefsdf,
         )
         # Use the last known coefficients for further processing
--- a/openstf/model/split_energy.py
+++ b/openstf/model/split_energy.py
@@ -12,7 +12,7 @@ import scipy.optimize
 
 import openstf.monitoring.teams as monitoring
 
-COEF_MAX_PCT_DIFF = 0.3
+COEF_MAX_FRACTION_DIFF = 0.3
 
 
 def split_energy(pid):
@@ -46,36 +46,37 @@ def split_energy(pid):
     coefdict.update({"MAE": mae})
     coefsdf = convert_coefdict_to_coefsdf(pj, input_split_function, coefdict)
 
-    # Get average coefs of previous runs and check if new coefs are valid
-    mean_coefs = db.get_energy_split_coefs(pj, mean=True)
-    if are_new_coefs_valid(coefdict, mean_coefs) is False:
+    # Get the coefs of previous runs and check if new coefs are valid
+    last_coefsdict = db.get_energy_split_coefs(pj, mean=True)
+    last_coefsdf = convert_coefdict_to_coefsdf(pj, input_split_function, last_coefsdict)
+    false_key = are_new_coefs_valid(coefdict, last_coefsdict)
+    if false_key is not None:
         # If coefs not valid, do not update the coefs in the db and send teams
         # message that something strange is happening
         monitoring.post_teams_alert(
-            "New splitting coefficients for pid {} deviate strongly from previously stored coefficients.".format(
-                pj["id"],
-            ),
+            f"New splitting coefficient {false_key} for pid {pj['id']} deviate strongly from previously stored coefficients."
+            + f"New coefficient: {coefdict[false_key]}, last coefficient: {last_coefsdict[false_key]}.",
             coefsdf=coefsdf,
         )
         # Use the last known coefficients for further processing
-        last_coefdict = db.get_energy_split_coefs(pj)
-        last_coefsdf = convert_coefdict_to_coefsdf(
-            pj, input_split_function, last_coefdict
-        )
         return last_coefsdf
+    else:
+        # Save Results
+        db.write_energy_splitting_coefficients(coefsdf, if_exists="append")
+        logger.info(
+            "Succesfully wrote energy split coefficients to database", pid=pj["id"]
+        )
+        return coefsdf
 
-    # Save Results
-    db.write_energy_splitting_coefficients(coefsdf, if_exists="append")
-    logger.info("Succesfully wrote energy split coefficients to database", pid=pj["id"])
-    return coefsdf
 
-
-def are_new_coefs_valid(new_coefs, mean_coefs):
+def are_new_coefs_valid(new_coefs, last_coefs):
     """Check if new coefficients are valid.
 
     Args:
-        new_coefs (dict): new coefficients for standard load profiles
-        mean_coefs (dict): average of previous coefficients for standard load profiles
+        new_coefs (dict): new coefficients for standard load profiles (i.e.
+            wind, solar, household)
+        mean_coefs (dict): average of previous coefficients for standard load
+            profiles (i.e. wind, solar, household)
 
     Returns:
         boolean: bool whether new coefs are valid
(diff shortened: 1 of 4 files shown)


openstf/model/split_energy.py
```diff
@@ -12,6 +12,8 @@
 
 import openstf.monitoring.teams as monitoring
 
+COEF_MAX_PCT_DIFF = 0.3
```
FrankKr, line 15: The variable says 'pct' (= Percent?), but you actually use a fraction and not a percentage

openstf/model/split_energy.py
```diff
@@ -42,23 +44,65 @@ def split_energy(pid):
...
+        )
+        # Use the last known coefficients for further processing
+        last_coefdict = db.get_energy_split_coefs(pj)
+        last_coefsdf = convert_coefdict_to_coefsdf(
+            pj, input_split_function, last_coefdict
+        )
+        return last_coefsdf
+
+    # Save Results
+    db.write_energy_splitting_coefficients(coefsdf, if_exists="append")
```
FrankKr, line 68: An explicit 'else: ' here is more clear (although it functionally doesn't matter)

openstf/model/split_energy.py
```diff
@@ -42,23 +44,65 @@ def split_energy(pid):
     error = components[["load", "Inschatting"]].diff(axis=1).iloc[:, 1]
     mae = error.abs().mean()
     coefdict.update({"MAE": mae})
+    coefsdf = convert_coefdict_to_coefsdf(pj, input_split_function, coefdict)
 
-    # Get average coefs of previous runs
+    # Get average coefs of previous runs and check if new coefs are valid
     mean_coefs = db.get_energy_split_coefs(pj, mean=True)
```
FrankKr, line 50: Do we still want the mean of the old coefs? Wouldn't just the most recent coefs make more sense?

openstf/model/split_energy.py
```diff
@@ -42,23 +44,65 @@ def split_energy(pid):
...
-    # Loop over keys and check if the difference with the average value is not more than 100%
-    # In case the difference is more tha 100% of the average set KPI coefs as expected to False
+    if are_new_coefs_valid(coefdict, mean_coefs) is False:
+        # If coefs not valid, do not update the coefs in the db and send teams
+        # message that something strange is happening
```
FrankKr, line 53: looks nice

openstf/model/split_energy.py
FrankKr: In addition, you wouldnt need L60-64

openstf/model/split_energy.py
```diff
@@ -42,23 +44,65 @@ def split_energy(pid):
...
+    db.write_energy_splitting_coefficients(coefsdf, if_exists="append")
+    logger.info("Succesfully wrote energy split coefficients to database", pid=pj["id"])
+    return coefsdf
+
+
+def are_new_coefs_valid(new_coefs, mean_coefs):
+    """Check if new coefficients are valid.
+
+    Args:
+        new_coefs (dict): new coefficients for standard load profiles
```
FrankKr, line 77: Perhaps make more clear that Standard Load Profiles also include wind and solar (and not just standard consumption)

openstf/model/split_energy.py
```diff
@@ -42,23 +44,65 @@ def split_energy(pid):
...
+        mean_coefs (dict): average of previous coefficients for standard load profiles
+
+    Returns:
+        boolean: bool whether new coefs are valid
+    """
+    # Loop over keys and check if the absolute difference with the average value is not
+    # more than COEF_MAX_PCT_DIFF x absolute average value.
     # If no previous coefs are stored an mean_coefs is empty and this loop wil not run
     for key in mean_coefs.keys():
```
FrankKr, line 86: What should happen if the previous coefs are empty? Can you check explicitly for this and make the desired outcome happen?

Also; what do you want to happen when the keys of old_coefs do not match the new coefs? I suppose you just want to skip that specific key, right? Can you include this to make this function more robust?

openstf/model/split_energy.py
```diff
@@ -42,23 +44,65 @@ def split_energy(pid):
...
     mean_coefs = db.get_energy_split_coefs(pj, mean=True)
-    # Loop over keys and check if the difference with the average value is not more than 100%
-    # In case the difference is more tha 100% of the average set KPI coefs as expected to False
+    if are_new_coefs_valid(coefdict, mean_coefs) is False:
+        # If coefs not valid, do not update the coefs in the db and send teams
+        # message that something strange is happening
+        monitoring.post_teams_alert(
+            "New splitting coefficients for pid {} deviate strongly from previously stored coefficients.".format(
+                pj["id"],
+            ),
```
FrankKr, line 57: Would make sense to include some information on which coefs are considered incorrect, or at least how the new coefs compare to the old onse

openstf/monitoring/teams.py
```diff
@@ -103,6 +107,28 @@ def post_teams_alert(msg, url=None):
     return post_teams(msg, url=url)
 
 
+def build_sql_query_string(df, table):
```
FrankKr, line 110: nice function!

openstf/monitoring/teams.py
```diff
@@ -103,6 +107,28 @@ def post_teams_alert(msg, url=None):
...
+
+    Returns:
+        string: sql query string of insert statement
+    """
+    sql_texts = [
+        "INSERT INTO " + table + " (" + str(", ".join(df.columns)) + ") VALUES \n"
+    ]
+    for index, row in df.iterrows():
+        if index != df.index[0]:
+            sql_texts.append(",\n")
```
FrankKr, line 125: Not needed? 1) Whitelines do not have meaning in sql, and 2) you can remove the `\n` on L121

FrankKr (changes requested):
Looking good! Nice functions with docstrings and unit tests.

Think I was able to find some points for improvement

openstf/model/split_energy.py
gfsdekoning: you are correct, i changed the name to `COEF_MAX_FRACTION_DIFF`

openstf/model/split_energy.py
gfsdekoning: 1. I kept the mean instead only the last coefs, in case the last ones were odd. I don't have a specific preference.
2. Because I am using the mean and not the last ones, I compare the coefs in dict-format with `mean_coefs` in dict-format. After that, I do not use those values further and get the last coefs.

Looking at this answer, it does make more sense to use the last ones, I thought the mean would have a preference, but that does not seem to be the case. I will compare the coefficients with the last ones.

openstf/model/split_energy.py
gfsdekoning: Added specification like in docstring of `convert_coefdict_to_coefsdf()`

openstf/model/split_energy.py
gfsdekoning: Ok

openstf/monitoring/teams.py
gfsdekoning: Added it for readability in output message in teams?

openstf/model/split_energy.py
gfsdekoning: I stop checking when one key is failing and added info of that key, new key value and old key value.

openstf/model/split_energy.py
gfsdekoning: 1. If previous coefs are empty, there is nothing to compare against and I would assume the new coefficients are always good, so they are added to the db. Why add it explicitly?
2. Good one, I added a check for `key in new_coefs`.

test/unit/model/test_split_energy.py
```diff
@@ -3,6 +3,9 @@
 # SPDX-License-Identifier: MPL-2.0
 
 import unittest
+from numpy import split
+
+from numpy.core.fromnumeric import mean
```
JanMaartenvanDoorn, line 8: split and mean seem not in use. Can you remove these import  lines?

openstf/model/split_energy.py
```diff
@@ -42,23 +44,68 @@ def split_energy(pid):
...
+    """Check if new coefficients are valid.
 
-    # Get average coefs of previous runs
-    mean_coefs = db.get_energy_split_coefs(pj, mean=True)
-    # Loop over keys and check if the difference with the average value is not more than 100%
-    # In case the difference is more tha 100% of the average set KPI coefs as expected to False
+    Args:
+        new_coefs (dict): new coefficients for standard load profiles (i.e.
+            wind, solar, household)
+        mean_coefs (dict): average of previous coefficients for standard load
```
JanMaartenvanDoorn, line 78: I think "mean_coefs" in the docstring should be changed to "last_coefs"

openstf/model/split_energy.py
gfsdekoning: 2.b if key is not present in new coefs, it should fail, because there is an old coef not in the new one

openstf/monitoring/teams.py
FrankKr: still not needed if you remove the \n on L121

gfsdekoning: now you get 
`"INSERT INTO table (key, value) VALUES 
('a', 1),
('b', 2)"`
 in your case it would be 
`"INSERT INTO table (key, value) VALUES 
('a', 1),('b', 2)"`
I would prefer option 1 for readability?

gfsdekoning: then the comma fails. You need a comma in between the values. Not at the start and not at the end.

test/unit/model/test_split_energy.py
gfsdekoning: Done

openstf/model/split_energy.py
gfsdekoning: Yes, changed.

openstf/model/split_energy.py
gfsdekoning: After a comment of JM, I changed it to checking for all keys and giving back all failing keys

FrankKr (approved):

PR #35, opened 2021-03-02
Feature pr template

FrankKr:
In addition to the issue-templates, I have suggested a pull-request-template.
Let's see if this works and add on it later?


bremme:
Nice additions! Perhaps nice/required to add the template to a specific sub-folder?: 

image

Perhaps its fine like it is and people will already interpreted it that way.  But I always like to know **why** certain changes where made. What was the overall strategy/goal to solve a certain issue.

It looks like everything after `Fixes issue: #...` is formatted as heading 1. That was probably not your intention. You could for example fix this by using something like this:

Fixes issue: `#1 name of the issues`

### Changes proposed in this PR include:

> Here you can elaborate on the chosen solution strategy, which changes did you make and which goal do they serve. Perhaps also which things are you still unsure of.

-
-
- ..

### Could you please pay extra attention to the points below when reviewing the PR:

> Here you can point out modules or complex implementation that require special attention, .e.g have a look at module `foo.py` and `bar.py`.

- 
- 
- ..

**Markdown source:**

```markdown

Fixes issue: `#1 name and number of the issue`

### Changes proposed in this PR include:

> Here you can elaborate on the chosen solution strategy, which changes did you make and which goal do they serve. Perhaps also which things are you still unsure of.

-
-
- ..

### Could you please pay extra attention to the points below when reviewing the PR:

> Here you can point out modules or complex implementation that require special attention, .e.g have a look at module `foo.py` and `bar.py`.

- 
- 
- ..
```

FrankKr:
Thanks, I included your suggestions!

PR #36, opened 2021-03-02
update teams message format

gfsdekoning:
Update teams message format to improve readability

### DIFF
--- a/openstf/model/split_energy.py
+++ b/openstf/model/split_energy.py
@@ -47,14 +47,14 @@ def split_energy(pid):
     coefsdf = convert_coefdict_to_coefsdf(pj, input_split_function, coefdict)
 
     # Get the coefs of previous runs and check if new coefs are valid
-    last_coefsdict = db.get_energy_split_coefs(pj, mean=True)
+    last_coefsdict = db.get_energy_split_coefs(pj)
     last_coefsdf = convert_coefdict_to_coefsdf(pj, input_split_function, last_coefsdict)
     invalid_coefs = determine_invalid_coefs(coefsdf, last_coefsdf)
     if not invalid_coefs.empty:
         # If coefs not valid, do not update the coefs in the db and send teams
         # message that something strange is happening
         monitoring.post_teams_alert(
-            f"New splitting coefficient(s) {list(invalid_coefs.coef_name)} for pid {pj['id']} deviate strongly from previously stored coefficients.",
+            f"New splitting coefficient(s) for pid **{pj['id']}** deviate strongly from previously stored coefficients.",
             invalid_coefs=invalid_coefs,
             coefsdf=coefsdf,
         )
--- a/openstf/model/split_energy.py
+++ b/openstf/model/split_energy.py
@@ -16,7 +16,8 @@ COEF_MAX_FRACTION_DIFF = 0.3
 
 
 def split_energy(pid):
-    """Function that caries out the energy splitting for a specific prediction job with id pid
+    """Function that caries out the energy splitting for a specific prediction job with
+    id pid.
 
     Args:
         pid (int): Prediction job id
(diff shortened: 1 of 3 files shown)


openstf/model/split_energy.py
```diff
@@ -47,14 +47,14 @@ def split_energy(pid):
...
-    last_coefsdict = db.get_energy_split_coefs(pj, mean=True)
+    last_coefsdict = db.get_energy_split_coefs(pj)
     last_coefsdf = convert_coefdict_to_coefsdf(pj, input_split_function, last_coefsdict)
     invalid_coefs = determine_invalid_coefs(coefsdf, last_coefsdf)
     if not invalid_coefs.empty:
         # If coefs not valid, do not update the coefs in the db and send teams
         # message that something strange is happening
         monitoring.post_teams_alert(
-            f"New splitting coefficient(s) {list(invalid_coefs.coef_name)} for pid {pj['id']} deviate strongly from previously stored coefficients.",
+            f"New splitting coefficient(s) for pid **{pj['id']}** deviate strongly from previously stored coefficients.",
```
bremme, line 57: This looks more then 88 character on a single line. We have disabled the build failing check on the line length, but we do use the convention. You can concatenate string over multiple lines for example like this:

```python
        monitoring.post_teams_alert(
            (
                f"New splitting coefficient(s) for pid **{pj['id']}** deviate strongly 
                "from previously stored coefficients."
            ),
            invalid_coefs=invalid_coefs,
            coefsdf=coefsdf,
        )
```

openstf/monitoring/teams.py
```diff
@@ -97,15 +97,38 @@ def post_teams_alert(msg, invalid_coefs=None, coefsdf=None, url=None):
...
             [
-                f"\ncoefficient name: {row.coef_name}, new value: {row.coef_value_new}, last value: {row.coef_value_last}"
+                f"\n* **{row.coef_name}**: {round(row.coef_value_new, 2)}, (previous: {round(row.coef_value_last, 2)})"
                 for index, row in invalid_coefs.iterrows()
             ]
         )
         query = build_sql_query_string(coefsdf, "energy_split_coefs")
-        msg = f"{msg}. {invalid_coefs_str}\nIf you would like to update the coefficients manually, use this query:\n {query}"
-
+        query_text = "If you would like to update the coefficients manually in the database, use this query:"
```
bremme, line 108: Line length? It probably a good idea to set a right margin in your editor so its easy to see when lines are too long: 
image

openstf/monitoring/teams.py
```diff
@@ -123,14 +146,23 @@ def build_sql_query_string(df, table):
...
+        df[col] = df[col].astype("str")
+
     sql_texts = [
-        "INSERT INTO " + table + " (" + str(", ".join(df.columns)) + ") VALUES \n"
+        "```INSERT INTO " + table + " (" + str(", ".join(df.columns)) + ") VALUES  \n"
     ]
     for index, row in df.iterrows():
         if index != df.index[0]:
-            sql_texts.append(",\n")
+            sql_texts.append(",  \n")  # 2 spaces and \n create a new line
```
bremme, line 163: Nice to build the query in a loop. Also very helpful to add the comment!

bremme (changes requested):
Nice improvements on the format, a couple small remarks.

openstf/model/split_energy.py
gfsdekoning: I was struggling with this indeed and didn't know what was the convention. I have replaced it.

openstf/monitoring/teams.py
gfsdekoning: I have an automatic black formatter, which take care of things, but it does not automatically split up long strings and docstrings. I have added a ruler for manual corrections.

bremme (approved):

bremme:
Look good @gfsdekoning !

PR #39, opened 2021-03-09
Bugfix predictionloop

JanMaartenvanDoorn:
Fixed a bug and some warnings.

### DIFF
--- a/openstf/model/capacity/model.py
+++ b/openstf/model/capacity/model.py
@@ -83,11 +83,15 @@ class CapacityPredictionModel:
             os.makedirs(directory, exist_ok=True)
 
         # save model
-        pickle.dump(self.model, open(directory / "model.pkl", "wb"))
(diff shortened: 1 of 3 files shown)


openstf/model/capacity/model.py
```diff
@@ -83,10 +83,14 @@ def save(self, directory):
             os.makedirs(directory, exist_ok=True)
 
         # save model
-        pickle.dump(self.model, open(directory / "model.pkl", "wb"))
+        with open(directory / "model.pkl", "wb") as fh:
```
bremme, line 86: nice to use a context manager here!

bremme:
Just a couple friendly reminders. The description does not really tell me what has been done and why. I believe @Frank has been working on a PR template, but perhaps that does not work yet?

Also I think the branch name could be more descriptive. One could argue that `predictionloop` is not really a description of the bug you're fixing?

bremme (approved):

PR #40, opened 2021-03-15
change splitting_energy teams communication from alerts to monitoring

gfsdekoning:
I have changed the message for an invalid energy splitting coefficients from alert to monitoring. Next to that, the markdown did something weird and fixed that as well.

I see that `post_teams()` has a code smell, but after discussions last time we decided to add an `if-statement` in the function, instead of creating a dedicated function for `post_teams_invalid_splitting_coef()`. Do we still want that? Or I can add a subfunction within `post_teams()` to build the invalid energy splitting coefficients message?

I solved the markdown issue by adding newlines.

### DIFF
--- a/openstf/model/figure.py
+++ b/openstf/model/figure.py
@@ -69,7 +69,7 @@ def plot_data_series(data, predict_data=None, horizon=47, names=None):
     series_names = {
         1: ("series",),
         2: ("train", "validation"),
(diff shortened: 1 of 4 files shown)


FrankKr (approved):
Good job. I think it's fine like this.
To code smell isn't that bad and in this piece of code I think it's okay.

PR #42, opened 2021-03-16
Feature ktps 490 refactor train function

gfsdekoning:
I have refactored the function `train_model_for_specific_pj()`, and renamed it to `train_model_pipeline()`. Subfunctions are pulled apart and tested.

@bremme the change you suggested in test was not working after all (i showed you a succesfull pytest of another file..)

### DIFF
--- a/openstf/model/train.py
+++ b/openstf/model/train.py
@@ -243,15 +243,19 @@ def predict_after_training_for_specific_pj(pj, model_trainer, split_input_data):
         model_trainer.confidence_interval,
     )
     train_predict = predictor.make_forecast(
-        split_input_data.train_data.iloc[:, 1:].drop("Horizon", axis=1, errors="ignore")
+        split_input_data["train_data"]
+        .iloc[:, 1:]
+        .drop("Horizon", axis=1, errors="ignore")
     )
     validation_predict = predictor.make_forecast(
-        split_input_data.validation_data.iloc[:, 1:].drop(
-            "Horizon", axis=1, errors="ignore"
-        )
+        split_input_data["validation_data"]
+        .iloc[:, 1:]
+        .drop("Horizon", axis=1, errors="ignore")
     )
     test_predict = predictor.make_forecast(
-        split_input_data.test_data.iloc[:, 1:].drop("Horizon", axis=1, errors="ignore")
+        split_input_data["test_data"]
+        .iloc[:, 1:]
+        .drop("Horizon", axis=1, errors="ignore")
     )
 
     return {
@@ -276,8 +280,8 @@ def create_evaluation_figures_for_specific_pj(
     Returns:
         tuple: tuple containing:
             plotly.graph_objects.Figure: A treemap of the features.
-            plotly.graph_objects.Figure: A line plot of of split input and predicted
-                data series.
+            dict: dict of line plots of of split input and predicted
+                data series per prediction horizon.
     """
 
     # Create figures
@@ -288,7 +292,7 @@ def create_evaluation_figures_for_specific_pj(
             list(split_predicted_data.values()),
             horizon,
         )
-        for horizon in split_input_data.train_data.Horizon.unique()
+        for horizon in split_input_data["train_data"].Horizon.unique()
     }
     return figure_features, figure_series
 
@@ -323,7 +327,9 @@ def evaluate_new_model_for_specific_pj(
     )
 
     # Combine validation + train data
-    combined = split_input_data.train_data.append(split_input_data.validation_data)
+    combined = split_input_data["train_data"].append(
+        split_input_data["validation_data"]
+    )
 
     # Evaluate model and store if better than old model
     # TODO Use predicted data series created above instead of predicting again
--- a/openstf/model/train.py
+++ b/openstf/model/train.py
@@ -75,7 +75,7 @@ def is_data_sufficient(data):
 def train_model_pipeline(pj, context, retrain_young_models=False, compare_to_old=True):
     """Main function that controls the training process of one prediction job.
 
-    Here, on a high level, the input data is collected, features are applied and the
+    On a high level: first, the input data is collected, features are applied and the
     model is trained and stored. Additionally models for energy components are also
     trained if requested. Model is saved if better than the old model unless specified
     otherwise.
@@ -88,7 +88,7 @@ def train_model_pipeline(pj, context, retrain_young_models=False, compare_to_old
         compare_to_old (bool): Compare new trained model to old trained model if True
 
     Returns:
-        Trained model (FIXME can be various datatypes at present)
+        None
     """
 
     # TODO Maybe this whole function can be replaced with a scikit learn pipeline?
@@ -99,7 +99,7 @@ def train_model_pipeline(pj, context, retrain_young_models=False, compare_to_old
 
     context.perf_meter.checkpoint("model-creation")
 
-    split_input_data = preprocessing_for_specific_pj(pj, context)
+    split_input_data = preprocess_for_specific_pj(pj, context)
 
     context.perf_meter.checkpoint("preprocessing")
 
@@ -107,7 +107,7 @@ def train_model_pipeline(pj, context, retrain_young_models=False, compare_to_old
 
     context.perf_meter.checkpoint("training")
 
-    split_predicted_data = predicting_after_training_for_specific_pj(
+    split_predicted_data = predict_after_training_for_specific_pj(
         pj,
         model_trainer,
         split_input_data,
@@ -128,15 +128,16 @@ def train_model_pipeline(pj, context, retrain_young_models=False, compare_to_old
 
 
 def create_model_for_specific_pj(pj, context, retrain_young_models=False):
-    """Create model trainer for one pj.
+    """Create model trainer and set hyperparams.
 
     Args:
-        context (openstf.task.utils.TaskContext): Task context for logging
         pj (dict): Prediction job
-        retrain_young_models (bool, optional): [description]. Defaults to False.
+        context (openstf.task.utils.TaskContext): Task context for logging
+        retrain_young_models (bool): Retrain models younger then MAX_AGE_YOUNG_MODEL
+            days if True. Defaults to False.
 
     Returns:
-        [type]: [description]
+        MLModelType: type of model trainer
     """
     # Make model trainer creator
     mc = ModelTrainerCreator(pj)
@@ -160,7 +161,19 @@ def create_model_for_specific_pj(pj, context, retrain_young_models=False):
     return model_trainer
 
 
-def preprocessing_for_specific_pj(pj, context):
+def preprocess_for_specific_pj(pj, context):
+    """Pre-process model data. Clean data and apply features
+
+    Args:
+        pj (dict): Prediction job
+        context (openstf.task.utils.TaskContext): Task context for logging
+
+    Raises:
+        ValueError: when data quality is insufficient
+
+    Returns:
+        dict: dict of cleaned data with features, split in train, validation and test
+    """
     # Specify training period
     # use hyperparam training_period_days if available
     hyperparams = {"training_period_days": TRAINING_PERIOD_DAYS, "featureset_name": "D"}
@@ -199,11 +212,29 @@ def preprocessing_for_specific_pj(pj, context):
 
 
 def train_for_specific_pj(model_trainer, split_input_data):
+    """Train model.
+
+    Args:
+        model_trainer (MLModelType): model trainer
+        split_input_data (dict): dict of cleaned data with features, split in train,
+            validation and test
+    """
     # Train model
     model_trainer.train(split_input_data.train_data, split_input_data.validation_data)
 
 
-def predicting_after_training_for_specific_pj(pj, model_trainer, split_input_data):
+def predict_after_training_for_specific_pj(pj, model_trainer, split_input_data):
+    """Predict data with new trained model for model evaluation.
+
+    Args:
+        pj (dict): Prediction job
+        model_trainer (MLModelType): model trainer
+        split_input_data (dict): dict of cleaned data with features, split in train,
+            validation and test
+
+    Returns:
+        dict: dict of predicted data, split in train, validation and test
+    """
     # Build predictor and create predictions
     predictor = PredictionModelCreator.create_prediction_model(
         pj,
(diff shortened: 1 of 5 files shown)


bremme:
@gfsdekoning I'm sure its something small, let me know if you need anymore help perhaps.

openstf/model/train.py
```diff
@@ -73,29 +72,75 @@ def is_data_sufficient(data):
...
     trained if requested. Model is saved if better than the old model unless specified
     otherwise.
 
     Args:
         pj (dict): Prediction job
         context (openstf.task.utils.TaskContext): Task context for logging
-        retrain_young_models (bool): Retrain models younger then 7 days if True
+        retrain_young_models (bool): Retrain models younger then MAX_AGE_YOUNG_MODEL
+            days if True
         compare_to_old (bool): Compare new trained model to old trained model if True
```
bremme, line 88: Maybe its more clear if you also explain the action here: Only save new model when its better than the old model. (or something similar)

openstf/model/train.py
```diff
@@ -73,29 +72,75 @@ def is_data_sufficient(data):
...
 
     Returns:
-        Trained model (FIXME can be various datatypes at present)
+        None
```
bremme, line 91: If a function return None implicitly you can remove this.

openstf/model/train.py
```diff
@@ -73,29 +72,75 @@ def is_data_sufficient(data):
...
     """
+
     # TODO Maybe this whole function can be replaced with a scikit learn pipeline?
+    model_trainer = create_model_for_specific_pj(pj, context, retrain_young_models)
+
+    if model_trainer is None:
```
bremme, line 97: This is unclear to me. Why does this happen? This seems like a unrecoverable error, should you not raise an exception here the normal flow of the function can not continue? Now it all passes silently.

openstf/model/train.py
```diff
@@ -73,29 +72,75 @@ def is_data_sufficient(data):
...
+            days if True
         compare_to_old (bool): Compare new trained model to old trained model if True
 
     Returns:
-        Trained model (FIXME can be various datatypes at present)
+        None
     """
+
     # TODO Maybe this whole function can be replaced with a scikit learn pipeline?
+    model_trainer = create_model_for_specific_pj(pj, context, retrain_young_models)
```
bremme, line 95: `create_model` creates a model_trainer?

openstf/model/train.py
```diff
@@ -73,29 +72,75 @@ def is_data_sufficient(data):
...
+
+    if model_trainer is None:
+        return
+
+    context.perf_meter.checkpoint("model-creation")
```
bremme, line 100: This looks nice, perform a step set a checkpoint and so on.

openstf/model/train.py
```diff
@@ -73,29 +72,75 @@ def is_data_sufficient(data):
...
+
+    split_input_data = preprocess_for_specific_pj(pj, context)
```
bremme, line 102: In my opinion `for_specific_pj` does not really add information.

openstf/model/train.py
```diff
@@ -73,29 +72,75 @@ def is_data_sufficient(data):
...
+    """Create model trainer and set hyperparams.
+
+    Args:
+        pj (dict): Prediction job
+        context (openstf.task.utils.TaskContext): Task context for logging
+        retrain_young_models (bool): Retrain models younger then MAX_AGE_YOUNG_MODEL
+            days if True. Defaults to False.
+
+    Returns:
+        MLModelType: type of model trainer
```
bremme, line 140: Is this right? This function created a `model_trainer` I think?

openstf/model/train.py
```diff
@@ -144,14 +204,37 @@ def train_model_for_specific_pj(
     train_data, validation_data, test_data = split_data_train_validation_test(
         clean_data_with_features
     )
+    return {
+        "train_data": train_data,
+        "validation_data": validation_data,
+        "test_data": test_data,
+    }
```
bremme, line 211: I think I like the idea of using data structures for these kind of things. I'm wondering if there is not already something available out of the machine learning tools that we already use? If not, I'm not such a fan of dictionairies, I would prefer a more rigid data container like a `dataclass` or `namedtuple` the main advantage of these containers is that they are documented (you just look up their definition and know which attributes do exists) and you can make then non extendible.

openstf/model/train.py
```diff
@@ -144,14 +204,37 @@ def train_model_for_specific_pj(
...
 
-    context.perf_meter.checkpoint("preprocessing")
 
+def train_for_specific_pj(model_trainer, split_input_data):
+    """Train model.
+
+    Args:
+        model_trainer (MLModelType): model trainer
```
bremme, line 218: This is not the right datatype I think.

openstf/model/train.py
```diff
@@ -144,14 +204,37 @@ def train_model_for_specific_pj(
...
+        split_input_data (dict): dict of cleaned data with features, split in train,
+            validation and test
+    """
     # Train model
-    model_trainer.train(train_data, validation_data)
+    model_trainer.train(split_input_data.train_data, split_input_data.validation_data)
```
bremme, line 223: You can't index a dictionary like this right? This should fail a unit test?

openstf/model/train.py
```diff
@@ -144,14 +204,37 @@ def train_model_for_specific_pj(
...
 
-    context.perf_meter.checkpoint("training")
 
+def predict_after_training_for_specific_pj(pj, model_trainer, split_input_data):
+    """Predict data with new trained model for model evaluation.
```
bremme, line 227: Why would you add for model evaluation here. This function does not need to know such things I think?

openstf/model/train.py
```diff
@@ -144,14 +204,37 @@ def train_model_for_specific_pj(
...
+
+    Args:
+        pj (dict): Prediction job
+        model_trainer (MLModelType): model trainer
```
bremme, line 231: same

openstf/model/train.py
```diff
@@ -160,30 +243,93 @@ def train_model_for_specific_pj(
...
     )
     test_predict = predictor.make_forecast(
-        test_data.iloc[:, 1:].drop("Horizon", axis=1, errors="ignore")
+        split_input_data["test_data"]
+        .iloc[:, 1:]
+        .drop("Horizon", axis=1, errors="ignore")
     )
 
-    context.perf_meter.checkpoint("predicting")
+    return {
```
bremme, line 261: see my other comment: nice to have this structures, perhaps choose a more suitable container.

openstf/model/train.py
```diff
@@ -160,30 +243,93 @@ def train_model_for_specific_pj(
...
+        "train_predict": train_predict,
+        "validation_predict": validation_predict,
+        "test_predict": test_predict,
+    }
+
+
+def create_evaluation_figures_for_specific_pj(
```
bremme, line 268: Just wondering if this should actually be part of the pipeline. Do you always want to run such a function?

openstf/model/train.py
```diff
@@ -160,30 +243,93 @@ def train_model_for_specific_pj(
...
+    model_trainer, split_input_data, split_predicted_data
+):
+    """Create figures of feature importance and data series of input and predicted data.
+
+    Args:
+        model_trainer (MLModelType): model trainer
```
bremme, line 274: same

openstf/model/train.py
```diff
@@ -160,30 +243,93 @@ def train_model_for_specific_pj(
...
+                data series per prediction horizon.
+    """
 
     # Create figures
     figure_features = plot_feature_importance(model_trainer.feature_importance)
     figure_series = {
         f"Predictor{horizon}": plot_data_series(
-            [train_data, validation_data, test_data],
-            [train_predict, validation_predict, test_predict],
+            list(split_input_data.values()),
```
bremme, line 291: Perhaps you can add the argument names here to make it more clear. If this data transformation is really required you could add a `to_list` method to your dataclass if you go that route.

openstf/model/train.py
```diff
@@ -160,30 +243,93 @@ def train_model_for_specific_pj(
...
+    Returns:
+        tuple: tuple containing:
+            plotly.graph_objects.Figure: A treemap of the features.
+            dict: dict of line plots of of split input and predicted
+                data series per prediction horizon.
+    """
 
     # Create figures
     figure_features = plot_feature_importance(model_trainer.feature_importance)
     figure_series = {
```
bremme, line 289: Perhaps this is all a bit too much for readability. You're doing dictionary comprehension, key interpolation, calling a function with nested calls as arguments while looping over a list which is constructed though some chained function call as well. I would suggest if you want to use the dict comprehension that you construct the variables first:

```python
    horizons = split_input_data["train_data"].Horizon.unique()
    input_data_list = list(split_input_data.values())
    predicted_data_list = list(split_predicted_data.values())

    figure_series = {
        f"Predictor{horizon}": plot_data_series(
            data=input_data_list, predicted_data=predicted_data_list, horizon=horizon
        ) for horizon in horizons
    }
```

openstf/model/train.py
```diff
@@ -160,30 +243,93 @@ def train_model_for_specific_pj(
...
+def evaluate_new_model_for_specific_pj(
+    pj,
+    context,
+    model_trainer,
+    split_input_data,
+    split_predicted_data,
+    compare_to_old=True,
+):
+    """Evaluate quality of new model with old model, save model, figures and send Teams
+    message.
```
bremme, line 309: Should this function really do that much? Would it not be more SRP and flexible to use seperate function for these tasks?

openstf/model/train.py
```diff
@@ -193,7 +339,7 @@ def train_model_for_specific_pj(
         model_trainer.store_model()
 
         # Save figures to disk
-        save_loc = Path(config.paths.trained_models) / str(pj["id"])
+        save_loc = Path(context.config.paths.trained_models) / str(pj["id"])
```
bremme, line 342: It would be way more flexible if this was an argument to this function. Now you can not use it from a notebook without modifying the context first.

openstf/tasks/utils/taskcontext.py
```diff
@@ -37,7 +37,7 @@ def __init__(
             post_teams_on_exception (bool, optional): If set to True the context
                 manager will automatically post a message to teams when an
                 exception is encountered. Defaults to True.
-            oon_exception (callable, optional): Callback, will be called
+            on_exception (callable, optional): Callback, will be called
```
bremme, line 40: 👍

test/unit/model/test_train.py
```diff
@@ -4,16 +4,25 @@
...
 
 import numpy as np
 import pandas as pd
 
 from openstf.model.general import MLModelType, split_data_train_validation_test
 from openstf.model.serializer.creator import ModelSerializerCreator
 from openstf.model.serializer.xgboost.xgboost import XGBModelSerializer
-from openstf.model.train import is_data_sufficient
+from openstf.model.train import (
+    is_data_sufficient,
```
bremme, line 18: you could just import `train` as well here -> `from openstf.model import train` this would give you a shorter import and it would be very clear which functions belong to the `train` module since you would have to call then with the module name -> `train.create_model`, `train.preprocess` etc.

test/unit/model/test_train.py
```diff
@@ -41,6 +50,29 @@
 }
```
bremme, line 52: `params` seem unused?

bremme (approved):
Looks good in general, added a couple remarks.

openstf/model/train.py
gfsdekoning: done

openstf/model/train.py
gfsdekoning: logging has already happened in `create_model_trainer()`. I added a comment here to clarify situation

openstf/model/train.py
gfsdekoning: chagned name to  `create_model_trainer()`

openstf/model/train.py
gfsdekoning: 👍

openstf/model/train.py
gfsdekoning: Agree, changed all names to
`create_model_trainer, preprocess_for_model_training, train_model, predict_after_training, evaluate_new_model`

openstf/model/train.py
gfsdekoning: True, I copied return from ModelTrainerCreator(), but it should be from ModelTrainerCreator.create_model_trainer(), so I changed it.

openstf/model/train.py
gfsdekoning: It did and it has been changed. odd that it is still here, I pushed that change long time ago.

openstf/model/train.py
gfsdekoning: I was not sure if the function name was clear enough.

openstf/model/train.py
gfsdekoning: I wanted to save lines, why is that bad?

openstf/model/train.py
gfsdekoning: I did not look into this too much, but agree the complexity in one line is increased. I followed your suggestion.

test/unit/model/test_train.py
gfsdekoning: seems to be, I deleteed it.

test/unit/model/test_train.py
gfsdekoning: thanks for the tip

openstf/model/train.py
gfsdekoning: added it as argument. If None is given, this is the `path_to_save`

openstf/model/train.py
gfsdekoning: Agree, I was also not happy about the number of actions in one function. I split it up

openstf/model/train.py
gfsdekoning: removed

openstf/model/train.py
bremme: Saving lines by itself can be a bad idea. I think readability and reducing complexity are more important then saving lines. You can write down very hard to read and complex code when you minimizing lines is your goal.

But its of course also al balancing act. In this case its a bit too much I think, too many things happen in one statement and also the list concatenating happens for every iteration in the dict comprehension.

bremme:
JM merged this branch by accident, not sure..but I guess you can reopen it once you pushed changed again?

PR #48, opened 2021-03-22
Feature units in ato reports

JanMaartenvanDoorn:
Added units to the values in the ATO reports


FrankKr (approved):
looks good. Side question; should we move ato_report.py out of short-term-forecasting and into our alliander-specific repo?

PR #51, opened 2021-03-22
increase minimum training_duration_days

FrankKr:
change hyperparam lookback options. too short is not good


JanMaartenvanDoorn (approved):
Looks good!

PR #56, opened 2021-03-29
KTPS-1343-hotfix-fallback-forecast

JanMaartenvanDoorn:
Fixed bug in fallback forecast


FrankKr (approved):
This is a hotfix and as such, merges directly to main instead of develop.
In future, please use 'hotfix' in the branchname

PR #59, opened 2021-04-07
Bugfix ktps 1359 fix logging exception

JanMaartenvanDoorn:
Fixed a bug that caused an exception in the logger when an old model has different features than the new model. Now the logger is called correctly. Also fixed this problem in other lines that had the same issue but did not cause problems (yet).


FrankKr (approved):
Looks good!

PR #63, opened 2021-04-15
Change installation instruction

FrankKr:
to use package at pypi.

Does this make sense?

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ SPDX-License-Identifier: MPL-2.0
 
 # Short Term Forcasting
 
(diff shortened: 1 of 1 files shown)


TonyXiang8787 (approved):

bremme:
Looks fine! Publishing a Python package at PyPi is a very common thing to do, right now it sounds a little bit like its something special: " The repository builds a package available at: https://pypi.org/project/openstf/" You could either remove it completely or rephrase is, for example: "The Python package is published on PyPi and can be installed using pip." 

But this is all not a really big deal I guess.

bremme (approved):

FrankKr:
> Looks fine! Publishing a Python package at PyPi is a very common thing to do, right now it sounds a little bit like its something special: " The repository builds a package available at: https://pypi.org/project/openstf/" You could either remove it completely or rephrase is, for example: "The Python package is published on PyPi and can be installed using pip."
> 
> But this is all not a really big deal I guess.

I'll rephrase it to make it less big of a deal;)

PR #66, opened 2021-04-20
Feature KTPS-127 reorganize modules and packages

bremme:
For this story I have reorganized the project in such a way:

* More inline with our conceptual architecture diagram
* Clear responsibilities for each module and package
* Clear separation between the orchestration layer and functional implementation layer
  * tasks are the entry point
  * machine learning tasks are implemented in a (machine learning) pipeline
  * Tasks may depend on monitoring, alerting and database
  * pipelines may depend on the database
  * Everything below the pipeline depend on none of the above

At present we have the following subpackages within `openstf`

* data
* feature_engineering
* metrics (new)
  * Everything responsible for measuring results, making figures, reports etc.
* model
  * Machine learning model implementations
* model_selection (new)
  * Similar to sklearn, at present splitting of datasets
* monitoring
* pipeline (new)
  * Implementation of the various machine learning pipelines
* postprocessing 
* preprocessing
* tasks
  * For the implementation of the tasks is moved (back) to the tasks itself, except for the ML tasks they have a pipeline with implementation
* validation (new)
  * validation code and checks, no modification of data in general

To arrive at this new structure many modules and functions have moved. For a complete overview have a look the "Files changed". In the old structure the responsibility of the `model` subpackage was not that clear for example, below an overview of where some of the modules where moved to:

* model.figure -> metrics.figure
* model.general -> model_selection.model_selection, tasks.create_solar_forecast,
* model.hyper_parameters -> pipeline.optimize_hyperparameters
* model.metrics -> metrics.metrics
* model.predict -> pipeline.create_forecast
* model.split_energy ->tasks.split_forecast
* model.train -> pipeline.train_model
* model.validation_robot -> validation.validation

**Questions**

* Is the division into the suggested subpackages what we want?
* Are all the responsibilities clear now?
* Is there some specific modules or symbol that should be moved to another (sub)package?
* Anything else?

**NOTE**
* I know the PR is broken, this will get fixed. But I think it's still useful to start the discussion.
  * (waiting for some fixes in the feature-KTPS-1329-refactor-apply-features branch)

**Attachments**
image


FrankKr:
Good idea, I've converted the PR to a draft, so notifications aren't send out for every fixing commit.
Please un-draft when you're ready;)

FrankKr (approved):
Fixed some code smells, looks good!

openstf/feature_engineering/apply_features.py
```diff
@@ -149,185 +55,50 @@ def apply_features(data, minute_list=None, h_ahead=24):
...
+    lag_feature_functions = generate_lag_feature_functions(feature_set_list, horizon)
 
-    # TODO it seems not all feature use the same convention
-    # better to choose one and stick to it
-    timedriven_functions = {
+    # Get timedrivenfeature functions
+    timedriven_feature_functions = {
         "IsWeekendDay": lambda x: (x.index.weekday // 5) == 1,
         "IsWeekDay": lambda x: x.index.weekday < 5,
-        "IsMonday": lambda x: x.index.weekday == 0,
```
bremme, line 162: @JanMaartenvanDoorn Why where these removed here?

openstf/feature_engineering/apply_features.py
```diff
@@ -149,185 +55,50 @@ def apply_features(data, minute_list=None, h_ahead=24):
                             np.random.uniform(0.7,1.7, 200)))
 
     """
-    if minute_list is None:
-        minute_list = []
 
-    lag_functions = generate_lag_functions(data, minute_list, h_ahead)
+    # Get lag feature functions
+    lag_feature_functions = generate_lag_feature_functions(feature_set_list, horizon)
```
bremme, line 60: @JanMaartenvanDoorn Why is it `generate_lag_feature_functions`  and `create_holiday_feature_functions`? If what there functions do is similar its probably more clear to use the same verb.

openstf/feature_engineering/apply_features.py
```diff
@@ -149,185 +55,50 @@ def apply_features(data, minute_list=None, h_ahead=24):
...
 
-    # TODO it seems not all feature use the same convention
-    # better to choose one and stick to it
-    timedriven_functions = {
+    # Get timedrivenfeature functions
+    timedriven_feature_functions = {
```
bremme, line 63: @JanMaartenvanDoorn maybe nice to gave a `generate`/`create` function for this as well like the other two. It seems like a similar thing you do.

openstf/feature_engineering/apply_features.py
```diff
@@ -149,185 +55,50 @@ def apply_features(data, minute_list=None, h_ahead=24):
...
-        new = {"Is" + char + str(int(thour)) + "Hour": func}
-        timedriven_functions.update(new)
-
-    df = data.copy()
-    # holiday_function are imported at the beginning of the file and includes all dutch
-    # school/work holidays
-    holiday_functions = create_holiday_functions()
+    # Only select features that occur in feature_set_list,
+    # if feature_set_list is none nothing is removed
+    if feature_set_list is not None:
```
bremme, line 76: Perhaps its simpler to remove this code and just skip the keys not in `feature_set_list` at line 95.

openstf/feature_engineering/apply_features.py
```diff
@@ -149,185 +55,50 @@ def apply_features(data, minute_list=None, h_ahead=24):
...
-
-    result = pd.DataFrame()
-    cols = []
-    for h_ahead in h_aheads:
-        res = apply_features(data.copy(), h_ahead=h_ahead)
-        res["Horizon"] = h_ahead
-        if len(res.columns) > len(cols):
-            cols = res.columns
-        result = result.append(res, sort=False)  # appending unsorts columns
+            data[name] = data.iloc[:, [0]].apply(featfunc)
```
bremme, line 95: @JanMaartenvanDoorn see my previous comment. If you remove line 76 top 86 you could just check the name here and continue. Als0 I would suggest to remove one level of complexity by removing the outer look.

```python
# CURRENT CODE
    # Only select features that occur in feature_set_list,
    # if feature_set_list is none nothing is removed
    if feature_set_list is not None:
        timedriven_feature_functions = {
            key: timedriven_feature_functions[key]
            for key in feature_set_list
            if key in timedriven_feature_functions
        }
        holiday_feature_functions = {
            key: holiday_feature_functions[key]
            for key in feature_set_list
            if key in holiday_feature_functions
        }

    # Add the features to the dataframe using previously defined feature functions
    for function_group in [
        lag_feature_functions,
        timedriven_feature_functions,
        holiday_feature_functions,
    ]:
        for name, featfunc in function_group.items():
            data[name] = data.iloc[:, [0]].apply(featfunc)
```

```python
# SUGGESTED IMPLEMENTATION
# Add the features to the dataframe using previously defined feature functions
for key, featfunc in feature_functions.items():
    # Don't generate feature is not in feature_set_list
    if key not in feature_set_list:
        continue
    data[key] = data.iloc[:, [0]].apply(featfunc)
```

openstf/feature_engineering/apply_features.py
```diff
@@ -149,185 +55,50 @@ def apply_features(data, minute_list=None, h_ahead=24):
...
-            "rated_power": 1,
-            "slope_center": 8.07,
-            "steepness": 0.664,
-        }
-    else:
-        required_properties = ["rated_power", "steepness", "slope_center"]
-        for prop in required_properties:
-            if prop not in turbine_data.keys():
-                raise KeyError(f"Required property '{prop}' not set in turbine data")
+    # Add additional winf features
```
bremme, line 97: @JanMaartenvanDoorn  winf == wind

openstf/feature_engineering/feature_applicator.py
```diff
@@ -0,0 +1,128 @@
...
+        Returns:
+            pd.DataFrame: Input DataFrame with an extra column for every added feature.
+
+        """
+
+        # Set default horizons if none are provided
+        if self.horizons is None:
+            self.horizons = [0.25, 24]
+
+        # Pre define output vairables
```
bremme, line 63: @JanMaartenvanDoorn vairables = variables

openstf/feature_engineering/feature_applicator.py
```diff
@@ -0,0 +1,128 @@
...
+        result = pd.DataFrame()
+        cols = []
+
+        # Loop over horizons and add corresponding features
+        for horizon in self.horizons:
+            res = apply_features(df.copy(), horizon=horizon)
+            res["Horizon"] = horizon
+            if len(res.columns) > len(cols):
```
bremme, line 71: @JanMaartenvanDoorn  this part is unclear to me? Why is keeping track of this required?

openstf/feature_engineering/feature_applicator.py
```diff
@@ -0,0 +1,128 @@
...
+                cols = res.columns
+            result = result.append(res, sort=False)  # appending unsorts columns
```
bremme, line 73: @JanMaartenvanDoorn is this only about columns, the rows are unsorted as well I guess?

openstf/feature_engineering/feature_applicator.py
```diff
@@ -0,0 +1,128 @@
...
+    add_missing_feature_columns,
+    remove_extra_feature_columns,
+)
+
+LATENCY_CONFIG = {"APX": 24}  # A specific latency is part of a specific feature.
+
+
+class AbstractFeatureApplicator(ABC):
+    def __init__(self, horizons: list, feature_set_list: list = None) -> None:
+        """
```
bremme, line 20: @JanMaartenvanDoorn Is seems that you use both `feature_set_list` and `feature_set_names` for the same thing? It would be more clear to use the same name for the same thing I think.

bremme, line 20: Do we need the cases there both arguments are optional? The code would be simpler without all the optional arguments, but perhaps its required?

openstf/feature_engineering/holiday_features.py
```diff
@@ -3,19 +3,24 @@
...
+HOLIDAY_CSV_PATH: str = (
+    PROJECT_ROOT / "openstf" / "data" / "dutch_holidays_2020-2022.csv"
+)
 
 
-def create_holiday_functions(
-    country="NL", years=None, path_to_school_holidays_csv=HOLIDAY_CSV_PATH
-):
+def create_holiday_feature_functions(
+    country: str = "NL",
```
bremme, line 20: @JanMaartenvanDoorn Does it makes sense to use a default argument here? Why not require to be explicit?

openstf/feature_engineering/lag_features.py
```diff
@@ -0,0 +1,152 @@
...
+
+        def func(x, shift=day):
+            return x.shift(freq="1d", periods=1 * shift)
+
+        new = {"T-" + str(int(day)) + "d": func}
+        lag_functions.update(new)
+    return lag_functions
+
+
+def generate_trivial_lag_features(horizon):
```
bremme, line 58: @JanMaartenvanDoorn I guess it makes sense to put this function right above/below `generate_non_trivial_lag_times`

openstf/feature_engineering/lag_features.py
```diff
@@ -0,0 +1,152 @@
...
+    """Function that generates relevant lag times for lag feature function creation.
```
bremme, line 59: I think you should never start a docstring with: "Function / This function". Just explain **what it does** not **what it is**.

openstf/feature_engineering/lag_features.py
```diff
@@ -0,0 +1,152 @@
...
+    This function is mostly used during training of models and not during predicting
```
bremme, line 60: I m not sure if you should explain how it could be used here. This low level function should perhaps not have knowledge about how its used.

openstf/feature_engineering/lag_features.py
```diff
@@ -0,0 +1,152 @@
...
+
+    return trivial_lag_times_minutes, lag_time_days_list
+
+
+def extract_lag_features(lag_features):
+    """Creates a list of lag minutes and a list of lag days that were used during
+    the training of the input model
+
+    Args:
+        lag_features (list[str]): All requested lag features
```
bremme, line 88: @JanMaartenvanDoorn I would not use `lag_features` as the argument here. You don't use the function in that way and it does not matter. In other places you call this thing `feature_set_names` or `feature_set_list`. Now we have three names for the same thing.

bremme: Thinking about it why not use `feature_names` ? That's what is is I guess?

openstf/feature_engineering/lag_features.py
```diff
@@ -0,0 +1,152 @@
...
+        if number_of_minutes is not None:
+            minutes_list.append(int(number_of_minutes[1]))
+        elif number_of_days is not None:
+            days_list.append(int(number_of_days[1]))
+
+    return minutes_list, days_list
+
+
+def generate_non_trivial_lag_times(data, height_treshold=0.1):
+    """This script calculates an autocorrelation curve of the load trace. This curve is
```
bremme, line 115: @JanMaartenvanDoorn "This script"? Just explain what it does, for example:

```
"""Generate non trivial lag features.

  Generate non trivial lag features by calculating a autocorrelation curve of the load trace. This curve is 
  subsequently used to add additional lag times as features.
"""

File: openstf/feature_engineering/lag_features.py [hunk 17]
```diff
@@ -0,0 +1,152 @@
...
+        height_treshold (float): minimal autocorrelation value to be recognized as a peak.
+
+    Returns:
+        list of ints with aditional non-trivial minute lags
+
+
+    """
+
+    def autocorr(x, lags):
+        """Function to make an autocorrelation curve"""
```
bremme, line 130: @JanMaartenvanDoorn explain what it does not what it is.

openstf/feature_engineering/weather_features.py
```diff
@@ -0,0 +1,332 @@
...
+M: float = 7.6
+TN: float = 240.7
+# Set some constants
+TORR: float = 133.322368  # 1 torr = 133 Pa
+# 1.168 is the mass of 1 m^3 of air on sea level with standard pressure.
+D: float = 1.168
+
+
+def calc_saturation_pressure(temperature: float or np.ndarray) -> float or np.ndarray:
+    """Function that calculates the water vapour pressure from the temperature
```
bremme, line 22: @JanMaartenvanDoorn -> Calculate the water saturation pressure. Or something similar

openstf/feature_engineering/weather_features.py
```diff
@@ -0,0 +1,332 @@
...
+
+    return TN / ((M / np.log10(vapour_pressure / A)) - 1)
+
+
+def calc_air_density(
+    temperature: float or np.ndarray,
+    pressure: float or np.ndarray,
+    rh: float or np.ndarray,
+) -> float or np.ndarray:
+    """Calculates the dewpoint
```
bremme, line 63: @JanMaartenvanDoorn calculates the dewpoint??

openstf/feature_engineering/weather_features.py
```diff
@@ -0,0 +1,332 @@
...
+        * ((pressure - 0.3783 * vapour_pressure) / 760 / TORR)
+    )
+
+    return air_density
+
+
+def add_humidity_features(
+    data: pd.DataFrame, feature_set_list: list = None
+) -> pd.DataFrame:
+    """Adds humidity features to the input dataframe.
```
bremme, line 93: @JanMaartenvanDoorn new line after title?

bremme:
@JanMaartenvanDoorn since we are coming your changes into this PR I added my comment here.

Look good, nice improvements to the feature engineering package.

A couple more general remarks:

* Double check you naming, it seems you use various names for `feature_names`
	* `feature_set_list`, `feature_set_names`, `lag_features` etc.
* I would suggest to use the same top level interface for adding features
	* Now lag features, timedrive features are added different then the weather features
* O and check the docstring, I stopped commenting, but I found several with some mistakes. In my opinion a docstring should focus on **what this function/method/class/module does** and on **what it is**. Also especially the lower level function should not have knowledge on how they should be used perhaps.

openstf/feature_engineering/apply_features.py
JanMaartenvanDoorn: These features are not included in the feature lists

openstf/feature_engineering/apply_features.py
JanMaartenvanDoorn: Yes I agree with consistency, however in this case the feature functions are simplydeclared and not generated using specific locic. In this case I think making a special 'generate' function would overcomplicate things.

openstf/feature_engineering/apply_features.py
JanMaartenvanDoorn: Good one, implemented your suggestion

openstf/feature_engineering/feature_applicator.py
JanMaartenvanDoorn: For larger horizons some features are not generated at all, this would result in a dataframe with fewer columns than for other horizons. This construction makes sure we return all columns in the end.

openstf/feature_engineering/feature_applicator.py
JanMaartenvanDoorn: Not sure but I think so

openstf/feature_engineering/feature_applicator.py
JanMaartenvanDoorn: well spotted, I thought I found them all. changed everything to feature_set_list

openstf/feature_engineering/holiday_features.py
JanMaartenvanDoorn: Yes, in case of a notebook I think we want to have minimal configuration and we just want to apply the features.

openstf/feature_engineering/feature_applicator.py
bremme: I'm sure there is a good reason to do so. My point is that I should not need the explanation you just gave to understand the code. The code should be self explanatory by using function/variable naming, docstring, comments etc.

openstf/feature_engineering/feature_applicator.py
bremme: Either remove the comment or make sure that what it says is actually true I think.

openstf/feature_engineering/holiday_features.py
bremme: We are not the only people using this library. And I don't think its really too much effort to write down which country you want in a notebook. In my opinion we should not use default/optional arguments in such cases.

Issue #71, opened 2021-04-30
[Feature] Make functionality to generate the model path


JanMaartenvanDoorn:
**WHAT:** 
Models are strored on disk/on persistent storage. For this we need a path that points to the proper location. This location is specific for a prediction job and datetime when the model was created. Determining the correct location currently is a model-specific method. In the desired situation, determining the model location is a model-independent function, which can be imported from utils.py. Additionally, when assessing if new model should be trained, the age of the latest model is checked. Include this functionality.

**HOW:**
-	Make function that returns the directory for the latest available model for a given pj (to load the model)
-	Make function that returns the model age (in days) for a given pj
-	Make function that returns the path for a newly created model.
-	Look in de model serializer in the old code base for inspiration
o https://github.com/alliander-opensource/short-term-forecasting/blob/develop/openstf/model/serializer/serializer.py
-	Implement new functions in the train_sklearn and predict_sklearn pipelines
- Think about the best place to implement these functions, maybe it is better to put these functions outside the pipeline? Make a proposal.

**TEST:**
-	Old model can be loaded during training and predicting
-	Age of the model can be checked in the train pipeline

PR #80, opened 2021-05-10
Feature #71 make functionality to generate model path

JanMaartenvanDoorn:


openstf/model/reporter.py
```diff
@@ -1,50 +1,58 @@
+# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
+#
+# SPDX-License-Identifier: MPL-2.0
 import os
```
bremme, line 4: Can you explain why the `reporter` should be part of the `model` package? Didn't we decide on the `metrics` package for reporting?

openstf/model/serializer.py
```diff
@@ -0,0 +1,204 @@
...
+from pathlib import Path
+
+import joblib
+import structlog
+from sklearn.base import RegressorMixin
+
+from ktpbase.config.config import ConfigManager
+
+
+MODEL_FILENAME = "model.sav"
```
bremme, line 15: Is this a common convention? Its maybe not that important, but a convention would be good I guess. And its not something you can easily change after this is merged. Once the models are stored using a certain extension you need extra code to handle a migration.

openstf/model/serializer.py
```diff
@@ -0,0 +1,204 @@
...
+        # Compose save path
+        save_path = self._build_save_folder_path()
+
+        # Create save path if nescesarry
+        save_path.mkdir(parents=True, exist_ok=True)
+
+        # Save model
+        try:
+            joblib.dump(model, save_path / MODEL_FILENAME)
+        except Exception as e:
```
bremme, line 64: Should this exception caught here on at the calling code? What type of things can typically go wrong?

openstf/model/serializer.py
```diff
@@ -0,0 +1,204 @@
...
+        Returns: Number of days since training of the model
+
+        """
+
+        # If we cannot determine the age return infinite days
+        if model_location is None:
+            return float("inf")
+
+        # trained_models/<pid>/<YYYYMMDDHHMMSS>/
+        datetime_string = model_location.name
```
bremme, line 196: Not sure if I get this. the `model_location` argument is the path to the most recent model file, that would be `trained_models/<pid>/<YYYYMMDDHHMMSS>/model.sav`. I don't see how that can be a datetime string.

openstf/model/serializer.py
```diff
@@ -0,0 +1,204 @@
...
+        save_folder = pid_model_folder / datetime_str
+
+        return save_folder
+
+    @staticmethod
+    def _determine_model_age_from_path(model_location: Path) -> float:
+        """Determines how many days ago a model is trained base on the folder name.
+
+        Args:
+            model_location: pathlib.Path: Path to the most recent model file
```
bremme, line 185: Is it really important that this is the most recent model file??

openstf/model/serializer.py
```diff
@@ -0,0 +1,204 @@
...
+            return float("inf")
+
+        # trained_models/<pid>/<YYYYMMDDHHMMSS>/
+        datetime_string = model_location.name
+
+        # Convert string to datetime object
+        model_datetime = datetime.strptime(datetime_string, FOLDER_DATETIME_FORMAT)
+
+        # Get time difference between now and training in days
+        model_age_days = (datetime.utcnow() - model_datetime).days
```
bremme, line 202: If we just store the datetime with the model, we would not need to do all of this actually.

openstf/model/serializer.py
JanMaartenvanDoorn: sklearn uses .joblib in their examples, I suggest we use that one.

bremme: Sound good!

openstf/model/serializer.py
JanMaartenvanDoorn: yes you are right, probably nothing goes wrong and if so then it is better to catch the exception higher up.

openstf/model/serializer.py
```diff
@@ -0,0 +1,204 @@
...
+        try:
+            loaded_model = joblib.load(location_most_recent_model / MODEL_FILENAME)
+        except Exception as e:
+            self.logger.error(
+                "Could not load most recent model!", pid=self.pj["id"], exception=e
+            )
+            raise FileNotFoundError
+
+        # exctract model age
+        model_age_in_days = self._determine_model_age_from_path(
```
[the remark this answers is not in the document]
JanMaartenvanDoorn: I think this is not possible as this is currently the method of building up a model artchive and keeping track of the model age. One could add the train data as an attribute to the model but in that case we still need a method of storing older model. I think for persistent storage serialization adding the training date to the foldename is a good way to do this.

openstf/model/serializer.py
```diff
@@ -0,0 +1,204 @@
...
+            location_most_recent_model
+        )
+
+        # Add model age to model object
+        loaded_model.age = model_age_in_days
+
+        return loaded_model
+
+    def _build_pid_model_folder_path(self, custom_folder: Path = None) -> Path:
```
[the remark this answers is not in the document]
JanMaartenvanDoorn: Good point, will do

openstf/model/serializer.py
JanMaartenvanDoorn: no not realy, removed this comment

openstf/model/serializer.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # If we cannot determine the age return infinite days
+        if model_location is None:
+            return float("inf")
+
+        # trained_models/<pid>/<YYYYMMDDHHMMSS>/
+        datetime_string = model_location.name
+
+        # Convert string to datetime object
+        model_datetime = datetime.strptime(datetime_string, FOLDER_DATETIME_FORMAT)
```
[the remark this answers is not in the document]
JanMaartenvanDoorn: added an try catch

openstf/model/reporter.py
JanMaartenvanDoorn: moved the file

bremme: If the comment is resolved, its probably a good idea to resolve it. Or you think I should do that?

openstf/model/serializer.py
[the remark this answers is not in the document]
bremme: Maybe I don't understand you properly. But why could you not add a created_data to the model before serialization? We do the same for features name and the confidence interval right? If you would do that you can remove the ~30 lines of code to parse the path back into a datetime.

```python
model.created = datetime.now(pytz.utc)

joblib.dump(model, "model.joblib")
```

PR #82, opened 2021-05-12
Fixed unit tests

JanMaartenvanDoorn:
Hello, here a pull-request that should fix all unit test issues. I removed the old unit tests that were testing code which is removed and adapted unit tests that were testing modified code. There is still one unit test failing: the test "test_something" which I think was left there intentionally. Monday I will have a look if I can expand coverage on the training pipeline a little bit.


openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -1,166 +0,0 @@
-# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
```
touronaug, line 1: Why was this file deleted?

JanMaartenvanDoorn: This hyperparameter optimization still uses the old models so I decided to delete it on this branch. Optimizing the new sklearn compatible models probably has to be implemented differently. I see there is currently no issue for this but I will make one.

Issue #70, opened 2021-04-30
[Feature] Implement fallback forecast in new pipeline


JanMaartenvanDoorn:
**WHAT:** When data completeness is low we do not want to make a regular forecast, but instead use a ‘fallback forecast’. The fallback forecast should return the daily profile of the day with the most extreme load of the past two weeks. C
Currently only a regular forecast is available in the new sklearn pipeline. Here we also implement a fallback forecast.

**HOW:**
- Look for de fallback forecast code in the old codebase 
https://github.com/alliander-opensource/short-term-forecasting/blob/2a9bf1e01f6b7554b4ba91f8fee878c1ae03b6a8/openstf/pipeline/create_forecast.py#L70
-	Implement a fallback forecast utility function that returns a forecast dataframe in case of low completeness.
-	Use above function in the predict pipeline
-	Make sure the “quality” column in de forecast dataframe has the proper value
-  Think about a logical place, should this be part of the prediction pipeline or maybe a separate pipeline?

**TEST:**
-	In case of low completeness a fallback forecast is created and written to the database in the sklearn predict pipeline
-	Information about the quality of the forecast is properly stored in de database

FrankKr:
I've started working on the issue!

FrankKr:
Issue fixed in PR. Closing issue

PR #84, opened 2021-05-20
Feature #70 ktps 1418 fallback forecast

FrankKr:
This PR resolves issue #70 
This PR includes functionality to deliver a fallbackforecast when inputdata for a regular forecast is insufficient.
Currently one type of fallback_strategy is implemented, but in the future this can be expanded.
We had some discussions on whether to integrate this function in the predict-pipeline, or as a separate pipeline. This implementation is in the predict-pipeline and seems the most logical. I included the forecast module under 'model'. If you have a suggestion for another location, please let me know.

A unit test is included for the fallback-function.
A unit test for the predict_pipeline is included but commented out. This requires a unit test for the happy flow of the predict_pipeline to function, which is not available. 

This PR is based on the pipeline-without-db-calls PR, as some issues were already fixed there which were not yet fixed in the develop or minimal-sklearn-implementation branches


openstf/model/fallback.py
```diff
@@ -0,0 +1,68 @@
...
+
+import pandas as pd
+import structlog
+
+
+def generate_fallback(
+    forecast_input: pd.DataFrame,
+    load: pd.DataFrame,
+    fallback_strategy: str = "extreme_day",
+):
```
JanMaartenvanDoorn, line 15: Nice that you are using type hints here, you should howere also add a type hint for the output. Suggestion:
def generate_fallback(
    forecast_input: pd.DataFrame,
    load: pd.DataFrame,
    fallback_strategy: str = "extreme_day",
) -> pd.DataFrame :

openstf/pipeline/predict_sklearn.py
```diff
@@ -49,14 +52,23 @@ def predict_pipeline(pj, input_data):
...
-    # Predict
-    forecast_input_data = data_with_features[forecast_start:forecast_end]
+        fallback_strategy = "extreme_day"  # this can later be expanded
+        logger.warning(
+            "Using fallback forecast",
+            forecast_type="fallback",
+            pid=pj["id"],
+            fallback_strategy=fallback_strategy,
+        )
+        model_forecast = generate_fallback(forecast_input_data, input_data.iloc[:, [0]])
```
JanMaartenvanDoorn, line 67: Instead of using inputdata.iloc[:, [0]] can we use inputdata['load']? This is maybe more readable. Moreover I think the 'load' column should always be available in this case. What do you think?

test/unit/pipeline/test_predict_sklearn.py
```diff
@@ -83,6 +83,19 @@ def test_pre_process_input_data(
         for mock_func in [nonzero_flatliner_mock, replace_invalid_data_mock]:
             self.assertEqual(mock_func.call_count, 1)
 
+    # TODO include the test below to see if the pipeline uses the fallback forecast correctly
+    # This test requires a functioning 'happy_flow_pipeline' test which is not yet developed
+    # @patch("openstf.validation.validation.is_data_sufficient")
+    # def test_incomplete_input(self, is_data_sufficient_mock):
+    #     """Test if a fallback forecast is used when input is incomplete"""
+    #     input_data = TestData.load('input_data.pickle')
```
JanMaartenvanDoorn, line 91: Do we want this commented out code to be included in the code? Maybe it's better make an issue for the pipeline unittests and add this code as a suggestion to that issue instead of here.

JanMaartenvanDoorn (approved):
Few minor things to think about. Overall nice pull request and good we have a fallback forecast in the new pipeline!

openstf/model/fallback.py
FrankKr: Good suggestion, I'll add it

openstf/pipeline/predict_sklearn.py
FrankKr: I went back and forth on this, but I agree with you. I'll change it and include this requirement in the docstring

test/unit/pipeline/test_predict_sklearn.py
FrankKr: Good point. I think the code would be a logical place for this unittest-in-progress, but commented code isn't very nice. I've uncommented the code and added a @unittest.skip("reason for skipping").

openstf/model/fallback.py
```diff
@@ -0,0 +1,68 @@
...
+
+import pandas as pd
+import structlog
+
+
+def generate_fallback(
+    forecast_input: pd.DataFrame,
+    load: pd.DataFrame,
+    fallback_strategy: str = "extreme_day",
+) -> pd.DataFrame:
```
bremme, line 15: Maybe good to think about consistency for the naming here as well.
 
 We used to have a `create_forecast` task which ran the `create_forecast.create_forecast_pipeline()` which used the `prediction_model.make_forecast()` method to make a `forecast`.
 
 Now we are moving to some other/better? vocabulary, but we seem to also introduce some inconsistencies. We still have the `create_forecast` task, but we now use `predict_sklearn.predict_pipeline` to run the pipeline which calls a sklearn model `model.predict()` to make a `forecast`
 
 I do like generate here, but I think you should also update the other vocabulary to restore consistency.
 
 Starting at the bottom if we have `model.predict()` to make a regular forecast perhaps it makes sense to go for a similar interface and or vocabulary here as well. So `fallback_model.predict()` or `model.predict_fallback()` etc.

openstf/model/fallback.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Args:
+        forecast_input (pandas.DataFrame): dataframe desired for the forecast
+        load (pandas.DataFrame): index=datetime, columns=['load']
+        fallback_strategy (str): strategy to determine fallback. options:
+            - extreme_day: use daily profile of most extreme day
+    Returns:
+        pandas.DataFrame: Fallback forecast DataFrame with columns:
+            'forecast', 'quality'
+    """
```
bremme, line 29: perhaps nice to add the exceptions that this function can raise to the docs as well.

openstf/model/fallback.py
```diff
@@ -0,0 +1,68 @@
...
+    # slice load dataframe, all rows for the day with the highest load
+    highest_daily_loadprofile = load.loc[
+        (load.index >= from_datetime) & (load.index < till_datetime)
+    ]
+
+    highest_daily_loadprofile.loc[:, "time"] = highest_daily_loadprofile.index.time
+
+    forecast = pd.DataFrame(index=forecast_input)
+    forecast["time"] = forecast.index.time
+    forecast = (
```
bremme, line 58: I could not really tell what exactly happens here, but that's probably my lack of pandas knowledge. If its easier to make it more clear (maybe just by a comment) that could improve readability perhaps.

openstf/model/fallback.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Currently only fallback_strategy=extreme day is implemented which return historic profile of most extreme day.
+
+    Args:
+        forecast_input (pandas.DataFrame): dataframe desired for the forecast
+        load (pandas.DataFrame): index=datetime, columns=['load']
+        fallback_strategy (str): strategy to determine fallback. options:
+            - extreme_day: use daily profile of most extreme day
+    Returns:
+        pandas.DataFrame: Fallback forecast DataFrame with columns:
```
bremme, line 27: This is an ongoing thing, its now easy to tell what a forecast actually is. Food for thoughts, some kind of schema or model would be really helpful, but out of scope here I guess.

openstf/model/fallback.py
FrankKr: after discussion, let's leave it as is for this pr

openstf/model/fallback.py
FrankKr: thanks for suggesting, I've included them

openstf/model/fallback.py
FrankKr: I've added some comments in the function to make it (hopefully) a little more clear.

PR #91, opened 2021-05-26
Feature 85 ktps 1423 unit tests pipelines

### LINKED ISSUE #85: [Feature] Add pipeline tests
**Is your feature request related to a problem? Please describe.**
Currently no unit tests are available for the train_model or create_forecast pipelines in the sklearn-branch of the project. These should definitely be included before merging back to develop. Is another name and location than 'unit' more logical for this type of test (e.g. implementation test), also since these tests potentially take a long time.

**Describe the solution you'd like**
Include unit tests which, for a given input dataframe:
- train a model 
- generate a forecast (using a fixed, previously stored model)

For input data you can probably use: https://github.com/alliander-opensource/short-term-forecasting/blob/features-KTPS-1368-implement-minimal-sklearn/test/unit/data/input_data.pickle
Have a look at an already available unit test for an example on how to use this input data using the BaseTestCase class.

This feature should use the branch: features-KTPS-1368-implement-minimal-sklearn as a starting point

FrankKr:
This PR closes #85 
Add unit tests of train and predict pipelines.
Mainly happy flows are tested, and some unhappy paths.

During the setup of the unit tests, the following issues were found and addressed:
- remove save_path parameter from generate_report
- remove param deep from df.copy, not supported in required pandas version
- fix fallback input
- a handful of other minor changes.

The actual tests themselves are at the bottom of 'files changed'.
Note that the test coverage of this PR is relatively low. This is due to adding tests of existing code not counting towards the coverage percentage.


openstf/pipeline/train_model_sklearn.py
```diff
@@ -135,7 +137,9 @@ def train_model_pipeline(
         train_y,
         eval_set=eval_set,
         early_stopping_rounds=DEFAULT_EARLY_STOPPING_ROUNDS,
+        verbose=0,
```
bremme, line 140: Just wondering what does this do? Would it not be more clear to use `False`? Is the default not `False` anyway?

test/unit/model/test_fallback.py
```diff
@@ -20,7 +20,7 @@ def test_fallback_happy(self):
         forc_section = TestData.load("fallback_index.csv")
 
         fallback_forecast = generate_fallback(
-            forecast_input=forc_section.index,
+            forecast_input=forc_section,
```
bremme, line 23: General comment about unit test naming. What do you think about using the conventions `test_<unit-under-test>_[<test-variant>]`. There are already quite a couple test which apply this convention. So in this case I would suggest to name the tests:

* `test_generate_fallback_happy_flow`
* `test_generate_fallback_empty_load`
* `test_generate_fallback_not_implemented_fallback_strategy`

Or something similar.

test/unit/pipeline/test_predict_sklearn.py
```diff
@@ -83,20 +86,35 @@ def test_pre_process_input_data(
...
-    # TODO include the test below to see if the pipeline uses the fallback forecast correctly
-    # This test requires a functioning 'happy_flow_pipeline' test which is not yet developed
-    @unittest.skip(
-        "No functioning happy_flow_pipeline test available. Add this one later"
+    @patch(
+        "openstf.pipeline.predict_sklearn.MODEL_LOCATION", Path("./test/trained_models")
     )
     @patch("openstf.validation.validation.is_data_sufficient")
     def test_incomplete_input(self, is_data_sufficient_mock):
         """Test if a fallback forecast is used when input is incomplete"""
```
bremme, line 94: Same as my previous comment. It can be hulpful to easily determine which test is for which unit by using consistent naming. So in this case `test_predict_pipeline_incomplete_inputdata` could be more suitable.

test/unit/pipeline/test_predict_sklearn.py
```diff
@@ -83,20 +86,35 @@ def test_pre_process_input_data(
...
+    def test_happy_flow_pipeline(self):
+        """Test the happy flow of the predict pipeline, using a previously trained model"""
+        self.pj = TestData.get_prediction_job(pid=307)
+        self.forecast_input = forecast_input
+        self.pj["feature_names"] = self.forecast_input.columns[1:]
+
+        forecast = predict_sklearn.predict_pipeline(
+            pj=self.pj, input_data=self.forecast_input
+        )
+        assert len(forecast) == 2878
```
bremme, line 114: I would suggest to use either `self.assert<AssertMethod>` or plain `assert`. Both are fine, but switching between the two might be confusing.

test/unit/pipeline/test_train_model_sklearn.py
```diff
@@ -40,6 +40,11 @@ def setUp(self) -> None:
             index=pd.date_range(datetime_start, datetime_end, freq="15T")
         )
 
+        self.train_input = TestData.load("reference_sets/307-train-data.csv")
+        self.pj["feature_names"] = self.train_input.columns[
```
bremme, line 44: If features_names are part of our prediction job model this feels bit hacky. In general it would be really nice if we can have a clear data model for our prediction job.

test/unit/pipeline/test_train_model_sklearn.py
```diff
@@ -56,6 +61,25 @@ def test_split_data_train_validation_test_period(self):
         self.assertEqual(len(validation_data), 1296)
         self.assertEqual(len(test_data), 1)
 
+    def test_train_pipeline_happy(self):
```
bremme, line 64: See previous comments, maybe I don't need to repeat this. But it would probably make more sense to name this test `test_train_model_pipeline_happy_flow`.

test/unit/pipeline/test_train_model_sklearn.py
```diff
@@ -56,6 +61,25 @@ def test_split_data_train_validation_test_period(self):
...
+        # Run the this should return a model and report
+        with self.assertLogs() as captured:
+            model, report = train_model_pipeline(
+                pj=self.pj,
+                input_data=self.train_input,
+                # check_old_model_age=False,
+                # compare_to_old=False,
+            )
+        # Check if a new model was stored based on logging
+        assert captured.records[-1].getMessage() == "Fitted a new model, not yet stored"
```
bremme, line 79: I think there should be a better way to check if a certain path through the code was taken. Comparing logger strings is a bit hacky and brittle.

test/unit/pipeline/test_train_model_sklearn.py
```diff
@@ -56,6 +61,25 @@ def test_split_data_train_validation_test_period(self):
...
+        assert str(type(model)) == "<class 'xgboost.sklearn.XGBRegressor'>"
```
bremme, line 80: This seems very brittle. It would be better to compare the type (not the string) to some kind of base estimator I think.

bremme:
Added a couple comments, hope this helps.

openstf/pipeline/train_model_sklearn.py
FrankKr: verbosity is [0-3] and specifies how much info is printed during training. Anything other than 0 is too much for practical use;)

test/unit/pipeline/test_train_model_sklearn.py
FrankKr: I did some searching, but this was the commonly proposed solution, at least without modifying the output of the original function

FrankKr:
> Added a couple comments, hope this helps.

Thanks, I'll incorporate them in the code!

test/unit/pipeline/test_train_model_sklearn.py
bremme: Not sure where you found this solution, but In my opinion its very bad practise to do such a thing. Assuming you want to check with this unit test if the model was actually fitted I would propose the following solution. sklearn has a `check_is_fitted` function which will raise a `NotFittedError` if the model is not fitted.

```python
    def test_train_model_pipeline_happy_flow(self):
        """Test happy flow of the train model pipeline
            
            NOTE this does not explain WHY this is the case?
            The input data should not contain features (e.g. T-7d),
            but it can/should include predictors (e.g. weather data)

        """
        model, report = train_model_pipeline(pj=self.pj, input_data=self.train_input)

        # check if the model was fitted (raises NotFittedError when not fitted)
        self.assertIsNone(sklearn.utils.validation.check_is_fitted(model))

        # check if model is sklearn compatible
        self.assertTrue(isinstance(model, sklearn.base.BaseEstimator))

        # check if report is a Report
        self.assertTrue(isinstance(report, Report))
```

FrankKr: Great, I'll use that.
Looking back at the code, I first made this unit test before your latest commits to the pipeline. Then, the pipeline did not return a model of report but stored them to disk, and had more paths that could be executed. For the current pipeline, your proposed tests makes a lot more sense, thanks.

test/unit/pipeline/test_train_model_sklearn.py
FrankKr: Good idea, I copied your suggestion below, and now it's included;)

test/unit/pipeline/test_predict_sklearn.py
FrankKr: Okay, I'll use self.assert

test/unit/pipeline/test_predict_sklearn.py
FrankKr: I'll change it, good suggestion!

test/unit/model/test_fallback.py
FrankKr: Good idea, I added it

test/unit/pipeline/test_train_model_sklearn.py
FrankKr: Thanks, it was changed to your suggestion

test/unit/pipeline/test_train_model_sklearn.py
```diff
@@ -56,6 +61,25 @@ def test_split_data_train_validation_test_period(self):
...
+        assert str(report.__class__) == "<class 'openstf.metrics.reporter.Report'>"
```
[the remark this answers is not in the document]
FrankKr: It was changed to your suggestion, thanks for pointing out the isinstance() function

PR #95, opened 2021-05-31
Bugfix #86 ktps 1424 calc kpi constant load

### LINKED ISSUE #86: [BUG] Calc_kpi for constant load
**Describe the bug**
If the historic load for a prediction_job is constant, the calc_kpi task returns an exception, while this is not desirable.
Would be more logical that a warning is raised (since this input is suspicious), but return valid kpi's nonetheless. E.g. rMAE = np.inf.
Also, please include a nice unit test for this to the current test cases for this task: https://github.com/alliander-opensource/short-term-forecasting/blob/eea5e3c2a669b31f08399c48f44b40963b4bfe36/test/unit/tasks/test_calculate_kpi.py
Good candidate for test-driven-development, and for first contributions.

**To Reproduce**
1. Run the https://github.com/alliander-opensource/short-term-forecasting/blob/develop/openstf/tasks/calculate_kpi.py for an input with constant load
2. See that an exception is raised

**Expected behavior**
- Raise a warning, since the input is suspiscious
- Calculate and return valid kpi's

**Screenshots**
image

**Additional context**
Lines with current exception: 
https://github.com/alliander-opensource/short-term-forecasting/blob/eea5e3c2a669b31f08399c48f44b40963b4bfe36/openstf/tasks/calculate_kpi.py#L167-L168

FrankKr:
fixes #86 

First replicated bug by adding unit test, subsequently fixed the code to make the test pass.
As described in the issue, if the `load` is constant, force the `range_` to be nan. 
This will result in nan's for the relative metrics, as desired.

### DIFF
--- a/openstf/metrics/metrics.py
+++ b/openstf/metrics/metrics.py
@@ -48,7 +48,11 @@ def r_mae(realised, forecast):
     The range is based on the load range of the previous two weeks"""
 
     # Determine load range on entire dataset
-    range_ = realised.max() - realised.min() if (realised.max() - realised.min()) != 0 else np.nan
+    range_ = (
+        realised.max() - realised.min()
+        if (realised.max() - realised.min()) != 0
+        else np.nan
+    )
 
     return mae(realised, forecast) / range_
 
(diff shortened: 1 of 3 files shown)


openstf/metrics/metrics.py
```diff
@@ -48,7 +48,11 @@ def r_mae(realised, forecast):
     The range is based on the load range of the previous two weeks"""
 
     # Determine load range on entire dataset
-    range_ = realised.max() - realised.min()
+    range_ = (
+        realised.max() - realised.min()
+        if (realised.max() - realised.min()) != 0
+        else np.nan
+    )
```
JanMaartenvanDoorn, line 55: netjes

JanMaartenvanDoorn (approved):
Ziet er goed uit!

Issue #88, opened 2021-05-26
[Feature] Implement basecase forecast model for sklearn pipeline


JanMaartenvanDoorn:
WHAT: The old model has a capability to make a base case forecast, Here we make a separate sklearn compatible model to make this basecase. 

HOW:
- Have a look at  the old basecase implementation
- Make a new estimator with a predict method that yields a base base forecast
- Have a look at the fallback forecast for timeshifting
- Think about best way how to implement this, perhaps use the forecast type as an argument to the pipeline.
- Use the forecast pipeline with the new model in the create_basecase_forecast task
- Try to reuse as much code as possible while keeping the ability to run the bascase forecast proces separately once every day.

TEST:
- basecases are generated with the new sklearn compatible model

PR #96, opened 2021-06-04
Feature #88 implement basecase forecast

JanMaartenvanDoorn:
Implemented the basecase model into sklearn. Also moved some functionality from the create_forecast_pipeline to the validation, preprocessing and postprocessing modules


FrankKr:
I think the tests are failing? Sure this PR is ready for review?

JanMaartenvanDoorn:
> I think the tests are failing? Sure this PR is ready for review?

Thought I fixed all issues but I didnt. Now everything should be fixed.

openstf/model/basecase.py
```diff
@@ -0,0 +1,91 @@
...
+import pandas as pd
+
+
+from sklearn.base import BaseEstimator, RegressorMixin
+
+MINIMAL_RESOLUTION: int = 15  # Used for validating the forecast input
+
+
+class BaseCaseModel(BaseEstimator, RegressorMixin):
+    def predict(self, forecsst_input_data: pd.DataFrame) -> pd.DataFrame:
```
FrankKr, line 16: `forecsst_input` <- typo

openstf/model/basecase.py
```diff
@@ -0,0 +1,91 @@
...
+        """Predict using the basecase method
```
FrankKr, line 17: Being a little picky; 'Predict using the basecase method' can be more clear -> Something to indicate that this class matches the sklearn class

openstf/model/basecase.py
```diff
@@ -0,0 +1,91 @@
...
+            overwrite_delay_hours (float): times before this in the future are not
+                forecasted
+
+        Raises:
+            ValueError: In case T-7d or T-14d is not present a ValueError is raised.
+            Also in the case the start of the forecast is before the horizon of the regular forecast,
+             a value error is raised
+        Returns:
+            pd.DataFrame: Basecase forecast
+
```
FrankKr, line 45: Good docstring, very clear!
Minor remark: under 'raises', it would be more clear if it was like this:


```
Raises:
    ValueError: if columns T-7d or T-14d is not present
    ValueError: If the start of the forecast is before the horizon of the regular forecast
```

openstf/model/basecase.py
```diff
@@ -0,0 +1,91 @@
...
+            raise ValueError(
+                f"Basecase forecast requested for horizon of regular forecast! Please check input! Requested start {requested_start}, allowed start {allowed_start}"
+            )
+
+        # Check if required features are provided
+        if not all(
+            item in forecast_input_data.columns.to_list() for item in ["T-14d", "T-7d"]
+        ):
+            raise ValueError(
+                "Could not make basecase, features T-7d and T-14d are required! Tip: Generate these features with a FeatureApplicator object."
```
FrankKr, line 66: I don't think you need both features, one is sufficient, right?

openstf/model/basecase.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+        basecase_forecast = basecase_forecast[
+            np.invert(basecase_forecast.index.duplicated())
+        ]
+
+        # Also make a basecase forecast for the forecast_other component. This will make a
+        # simple basecase components forecast available and ensures that the sum of
+        # the components (other, wind and solar) is equal to the normal basecase forecast
+        # This is important for sending GLMD messages correctly to TenneT!
+        basecase_forecast["forecast_other"] = basecase_forecast["forecast"]
```
FrankKr, line 89: Should this be part of this function? 
This function is currently used as model.predict. However, our regular model implementation does not return a dataframe with 'forecast_other' in the columns. Can 'forecast_other' be included more similarly to the regular forecast?

openstf/model/confidence_interval_applicator.py
```diff
@@ -2,18 +2,45 @@
...
+import pandas as pd
 from scipy import stats
 
 
+class ConfidenceIntervalApplicatorBaseCase:
+    def __init__(self, model):
+        self.confidence_interval = model.confidence_interval
+
+    def add_confidence_interval(self, basecase_forecast: pd.DataFrame) -> pd.DataFrame:
+        """Adds confidence interval to basebase forecast
```
FrankKr, line 14: Could you add some docstring to explain how ConfidenceIntervalApplicatorBaseCase differs from ConfidenceIntervalApplicator? Would help the next reader:)

test/unit/pipeline/test_create_basecase_pipeline.py
```diff
@@ -0,0 +1,56 @@
...
+        # Test length of the output
+        self.assertEqual(len(base_case_forecast), 1153)
+
+        # Test available columns
+        self.assertEqual(
+            [
+                "forecast",
+                "forecast_other",
+                "T-7d",
+                "T-14d",
```
FrankKr, line 47: Do we want the basecaseforecast to have these columns? 
I would expect that these feature-columns would NOT be returned in the basecase dataframe.

test/unit/pipeline/test_create_basecase_pipeline.py
```diff
@@ -0,0 +1,56 @@
...
+        # Shift example data to match current time interval as code expects data
+        # available relative to the current time.
+        utc_now = (
+            pd.Series(datetime.utcnow().replace(tzinfo=pytz.utc))
+            .min()
+            .round("15T")
+            .to_pydatetime()
+        )
+        most_recent_date = forecast_input.index.max().ceil("15T").to_pydatetime()
+        delta = utc_now - most_recent_date + timedelta(3)
```
FrankKr, line 32: This works. Alternatively (see it as a suggestion for next yime) you could mock the 'datetime.utcnow()' function, so this returns a pre-defined datetime. Like that, you are certain the input of the function is always the same when the test is executed

FrankKr (changes requested):
Looks good. 
I have commented some points where a little more documentation would be helpful.
And I placed 2 in-depth comments/remarks.

Additionally, sonar cloud has found 5 code smells: https://sonarcloud.io/project/issues?id=alliander-opensource_short-term-forecasting&pullRequest=96&resolved=false&types=CODE_SMELL
Perhaps good idea to fix those?

openstf/model/basecase.py
JanMaartenvanDoorn: In the old implementation both were used. I tried to stick as much as possible to the old methodology. We can however decide to now to drop the T-7d requierement.

openstf/model/basecase.py
JanMaartenvanDoorn: Good point, I will make a postproccessing step for this.

test/unit/pipeline/test_create_basecase_pipeline.py
JanMaartenvanDoorn: you are right I will remove them.

FrankKr (approved):

PR #98, opened 2021-06-07
Hotfix ktps 1584 completeness check

FrankKr:
fix two aspects:
- calculate completeness correctly
- do not generate T-30min feature for horizon=47hours.

JM and I discussed changes over the pfone


FrankKr:
Note that the duplication is fixed in the sklearn branch.

JanMaartenvanDoorn (approved):
Look good!

Issue #92, opened 2021-05-27
Modularize train_sklearn


FrankKr:
**Description**
In the sklearn-branch, the train_model_pipeline is halfway modularized. Recently, some functionality has been moved from the pipeline and is now included as TODO commented-out-code. Please add this functionality back into the pipeline, but make each step easiliy cofigurable, e.g. using boolean inputs.

Make a new layer with all persistence storage dependencies accoring to the picture below.

![image](https://user-images.githubusercontent.com/52956303/120295994-f931c100-c2c7-11eb-9c10-a7c4c89ac14e.png)

**Expected behavior**
There is a pipeline where all possible steps (check previous model age, compare to previous model, store the model, ...) are configurable in the input.
The complete functionality is available in a pipeline.
Current "train_model_pipeline" is renamed to "train_model_pipeline_core"
Do not forget to expand the currently available unit tests

**Additional context**
Start from this branch: https://github.com/alliander-opensource/short-term-forecasting/blob/features-KTPS-1368-implement-minimal-sklearn/openstf/pipeline/train_model_sklearn.py
Re-implement this code: https://github.com/alliander-opensource/short-term-forecasting/blob/84ceced4518418c1f8268f2a32117f16cc5a6892/openstf/pipeline/train_model_sklearn.py#L32-L67

PR #101, opened 2021-06-08
Feature #92 modularize sklearn

JanMaartenvanDoorn:
Modularized the code according to the picture in the issue #92 . Expanded unittest coverage and merged changed from develop and main into this branch. Hopefully this will make it easier to merge everything back into develop.

Most of the changes are in these files(other changes result from merging with main and develop):

- openstf/pipeline/create_component_forecast.py
- openstf/pipeline/create_forecast_sklearn.py
- openstf/pipeline/train_model_sklearn.py
- openstf/tasks/create_basecase_forecast.py
- openstf/tasks/create_components_forecast.py
- openstf/tasks/create_forecast.py
- openstf/tasks/train_model.py

Tests:
- test/unit/pipeline/test_create_component_forecast_pipeline.py
- test/unit/tasks/test_create_basecase_forecast.py
- test/unit/tasks/test_create_forecast.py
- test/unit/tasks/test_create_component_forecast.py
- test/unit/tasks/test_train_model.py


openstf/metrics/metrics.py
```diff
@@ -48,7 +48,11 @@ def r_mae(realised, forecast):
     The range is based on the load range of the previous two weeks"""
 
     # Determine load range on entire dataset
-    range_ = realised.max() - realised.min()
+    range_ = (
```
bremme, line 51: SonarCloud complains that we have some duplicated code here. It will probably also improve readability to  move the determination of the range and setting it to np.nan under some conditions to a separate function.

openstf/feature_engineering/feature_applicator.py
```diff
@@ -66,6 +64,9 @@ def add_features(
             pd.DataFrame: Input DataFrame with an extra column for every added feature.
         """
 
+        if latency_config is None:
```
bremme, line 67: Just by reading the docstring its not totally clear when exactly and which features will not be included. I assume by default only the timedriven features based on the historic load are not included in some cases. Perhaps we can make this even more clear (or perhaps I just have to read it better).

openstf/feature_engineering/lag_features.py
```diff
@@ -86,10 +90,14 @@ def extract_lag_features(feature_names: List[str]) -> Tuple[list, list]:
         elif number_of_days is not None:
             days_list.append(int(number_of_days[1]))
 
+    # Discard lag times that are not available for the specified horizon
+    minutes_list = list(set([i for i in minutes_list if i > horizon * 60]))
```
bremme, line 94: just checking, Is the sign the right direction here? Should it not be the other way around -> `if i <= horizon * 60`. Also why is `set` required, can we have duplicates here and why? Perhaps explain when applicable.

openstf/model_selection/model_selection.py
```diff
@@ -5,17 +5,18 @@
...
-    back_test=False,
-    period_sampling=True,
-    period_timedelta=timedelta(days=2),
-):
+    data: pd.DataFrame,
+    test_fraction: float = 0.0,
+    validation_fraction: float = 0.15,
+    back_test: bool = False,
+    period_sampling: bool = True,
+    period_timedelta: timedelta = timedelta(days=2),
```
bremme, line 18: It might be unsafe to use a timedelta instance here (similar to not use [] or {} as default arguments). Perhpas better to just use a plain datatype in days/hours here. Especially since we are not using any of the timedelta features -> `int(period_timedelta / delta)`. I'm not even sure what happens here? Which unit will we get the timedelta? Seconds, days etc?

openstf/pipeline/create_component_forecast.py
```diff
@@ -0,0 +1,51 @@
...
+        input_data (pd.DataFrame): Input data fore the components forecast, this usually is a regular forecast
+        weather_data (pd.DataFrame): Weather data with 'radiation' and 'windspeed_100m' columns
+        split_coefs (dict): coefieicnts for the splitting that are determined earlier
+
+    Returns:
+        pd.DataFrame with component forecasts
+
+    """
+    logger = structlog.get_logger(__name__)
+    logger.info("Make components prediction", prediction_id=pj["id"])
```
bremme, line 25: This one we don't use that often I think. I guess its better to use `pid` or `prediction_job_id` (even better would be to always use the same).

openstf/pipeline/create_component_forecast.py
```diff
@@ -0,0 +1,51 @@
...
+
+import openstf.postprocessing.postprocessing as postprocessing
+
+
+def create_components_forecast_pipeline(pj, input_data, weather_data, split_coefs):
+    """Pipeline for creating a component forecast
+
+    Args:
+        pj (dict): Prediction job
+        input_data (pd.DataFrame): Input data fore the components forecast, this usually is a regular forecast
```
bremme, line 16: Would it not be more clear to call it forecast in this case. The docstring does not explain what columns are expected here. I guess you need a very specific forecast dataframe to create the components forecast right?

openstf/pipeline/create_component_forecast.py
```diff
@@ -0,0 +1,51 @@
...
+    except Exception as e:
+        # In case something goes wrong we fall back on aan empty dataframe
+        logger.warning(
+            f"Could not make component forecasts: {e}, falling back on series of zeros!",
+            exc_info=e,
+        )
+        forecasts = pd.DataFrame()
+    else:
+
+        forecasts = forecasts.drop(
```
bremme, line 41: How come that column exist if it should not? Would it not make more sense that the `split_forecast_in_components` function does this?

openstf/pipeline/create_component_forecast.py
```diff
@@ -0,0 +1,51 @@
...
+    """Pipeline for creating a component forecast
+
+    Args:
+        pj (dict): Prediction job
+        input_data (pd.DataFrame): Input data fore the components forecast, this usually is a regular forecast
+        weather_data (pd.DataFrame): Weather data with 'radiation' and 'windspeed_100m' columns
+        split_coefs (dict): coefieicnts for the splitting that are determined earlier
+
+    Returns:
+        pd.DataFrame with component forecasts
```
bremme, line 21: How does the return look like, which columns will be added to the components forecast for example?

bremme: (Future, all these explanation would be unnecessary once we have defined our data models of course)

openstf/pipeline/create_forecast_sklearn.py
```diff
@@ -1,52 +1,55 @@
...
 
-# TODO add loading of model to task
-# # Load most recent model for the given pid
-# model = PersistentStorageSerializer(
-#     trained_models_folder=MODEL_LOCATION
-# ).load_model(pid=pj["id"])
+def create_forecast_pipeline(
+    pj: dict, input_data: pd.DataFrame, trained_models_folder: Union[str, Path]
+) -> pd.DataFrame:
+    """Midle level pipeline that takes care of all persistent storage dependencies
```
bremme, line 28: The title of the docstring should be a bit shorter I think. More detailed explanation goes in the body I guess. It seems also a bit inconsistent with the other pipeline docstrings. Why is this a middle level pipeline and some others not, what does middle  mean in this context? Perhaps something like this makes more sense:?

```python
    """Create forecast pipeline.
        
        This is the top-level pipeline which included loading the most recent model 
        for the given prediction job.
    """
```

openstf/pipeline/create_forecast_sklearn.py
```diff
@@ -1,52 +1,55 @@
...
 
-# def create_forecast_pipeline(pj, input_data, trained_models_folder=None):
-#     logger = structlog.get_logger(__name__)
-#     config = ConfigManager.get_instance()
+    Args:
+        pj (dict): Prediction job
+        input_data (pd.DataFrame): Raw training input data
```
bremme, line 32: Is Raw training input data a logical name here?

openstf/pipeline/create_forecast_sklearn.py
```diff
@@ -1,52 +1,55 @@
...
 
-#     # TODO write forecast to db ???
+    return create_forecast_pipeline_core(pj, input_data, model)
 
 
-def create_forecast_pipeline_core(pj, input_data, model):
+def create_forecast_pipeline_core(
+    pj: dict, input_data: pd.DataFrame, model: RegressorMixin
+) -> pd.DataFrame:
     """Computes the forecasts and confidence intervals given a prediction job and input data.
```
bremme, line 51: Same the `summary` line and the `description` are mixed here: 
image

openstf/pipeline/train_model_sklearn.py
```diff
@@ -29,59 +26,69 @@
...
+        logger.warning(
+            f"Old model is younger than {MAXIMUM_MODEL_AGE} days, skip training"
+        )
+        return
+
+    # Train model with core pipeline
+    try:
+        model, report = train_model_pipeline_core(pj, input_data, old_model)
+    except RuntimeError as e:
+        logger.error(f"Exception occured during training: {e}")
```
bremme, line 75: probably better to log the exception with the `exc_info` key (if I'm correct)

openstf/pipeline/train_model_sklearn.py
```diff
@@ -29,59 +26,69 @@
...
 
 
 def train_model_pipeline_core(
     pj: dict,
     input_data: pd.DataFrame,
     old_model: RegressorMixin = None,
-    horizons: List[float] = DEFAULT_TRAIN_HORIZONS,
-) -> Tuple[RegressorMixin, Report]:
-    """Run the train model pipeline.
+    horizons: List[float] = None,
```
bremme, line 89: nice, I like this construction better, more flexible.

openstf/pipeline/train_model_sklearn.py
```diff
@@ -29,59 +26,69 @@
...
+) -> tuple[RegressorMixin, Report]:
+    """Run the train model core pipeline. This pipeline has no database or persisitent storage dependencies.
```
bremme, line 91: Same as the other docstrings. You probably know. But if in doubt there are good resources:

[Python Google docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)
[PEP-257](https://www.python.org/dev/peps/pep-0257)

openstf/tasks/calculate_kpi.py
```diff
@@ -165,7 +165,9 @@ def calc_kpi_for_specific_pid(pid, start_time=None, end_time=None):
 
     # Raise exception in case of constant load
     if combined.load.nunique() == 1:
-        raise ValueError("The load is constant!")
+        structlog.get_logger().warning(
```
bremme, line 168: I would suggest to always give the logger a name. I guess `structlog.get_logger(__name__)` should work well here.

openstf/tasks/create_basecase_forecast.py
```diff
@@ -18,16 +18,50 @@
...
+    # Retrieve input data
+    input_data = context.database.get_model_input(
+        pid=pj["id"],
+        location=[pj["lat"], pj["lon"]],
+        datetime_start=datetime_start,
+        datetime_end=datetime_end,
+    )
+
+    # Make basecase forecast using the corresponding pipeline
+    basecase_forecast = basecase_pipeline(pj, input_data)
```
bremme, line 52: This seems inconsistent with the other pipeline names -> `create_forecast_pipeline`, `train_model_pipeline` etc. They all use a verb as the first element.

openstf/tasks/create_components_forecast.py
```diff
@@ -29,21 +29,78 @@
...
 from openstf.tasks.utils.utils import check_status_change, update_status_change
 from openstf.tasks.utils.predictionjobloop import PredictionJobLoop
 from openstf.tasks.utils.taskcontext import TaskContext
 
+T_BEHIND_DAYS = 0
+T_AHEAD_DAYS = 3
+
 
-def create_components_forecast_pj(pj, context):
+def create_components_forecast_task(pj, context):
```
bremme, line 46: This seems inconsistent with the task above. `create_basecase_forecast` Perhaps it makes sense to postfix the task function with `_task`. But I would suggest to apply that to every task or not.

openstf/tasks/create_components_forecast.py
```diff
@@ -29,21 +29,78 @@
...
+    """Top level task that creates a components forecast.
+    On this task level all database and context manager dependencies are resolved.
+
+    Args:
+        pj (dict): Prediction job
+        context (TaskContext): Contect object that holds a config manager and a database connection
+    """
+    logger = structlog.get_logger(__name__)
     if pj["train_components"] == 0:
```
bremme, line 55: I guess this name is a little misleading as well. `train_components` also means don't create a components forecast. But I guess it makes somewhat sense.

bremme: On the other hand we do not even train components or models for components right? Now I m confused :)

openstf/tasks/create_components_forecast.py
```diff
@@ -29,21 +29,78 @@
...
 
+    # Define datetime range for input data
+    datetime_start = datetime.utcnow() - timedelta(days=T_BEHIND_DAYS)
+    datetime_end = datetime.utcnow() + timedelta(days=T_AHEAD_DAYS)
+
+    logger.info(
+        "Get predicted load", datetime_start=datetime_start, datetime_end=datetime_end
+    )
+    # Get most recent load forecast
+    input_data = context.database.get_predicted_load(
```
bremme, line 69: hmm. this seems confusing as well. We get something which we saved as a **forecast** by calling **get_predicted_load** and call it **input_data** and commented it with **load forecast**. Perhaps its possible to be a bit more consistent.

openstf/tasks/create_components_forecast.py
```diff
@@ -29,21 +29,78 @@
...
+    )
+    # Check if forecast is not empty
+    if len(input_data) == 0:
+        logger.warning(f'No forecast found. Skipping pid {pj["id"]}')
+        return
+
+    logger.info("retrieving weather data")
+    # TODO make ktpbase function to retrieve inputdata for component forecast in one call,
+    #  this will make this function much shorter
+    # Get required weather data
```
bremme, line 80: How could I know that the "required" weather data is only "radiation", "windspeed_100m". Seems really arbitrary.

openstf/tasks/create_components_forecast.py
```diff
@@ -29,21 +29,78 @@
...
+    weather_data = context.database.get_weather_data(
+        [pj["lat"], pj["lon"]],
+        ["radiation", "windspeed_100m"],
+        datetime_start=datetime_start,
+        datetime_end=datetime_end,
+        source="optimum",
```
bremme, line 86: Is this not the default? Do you get less optimum data if you don't specifically ask for it?

openstf/tasks/create_components_forecast.py
```diff
@@ -29,21 +29,78 @@
...
+        return
+
     # Make forecast for the demand, wind and pv components
-    make_components_prediction(pj)
+    forecasts = create_components_forecast_pipeline(
+        pj, input_data, weather_data, split_coefs
+    )
+
+    # save forecast to database #######################################################
+    context.database.write_forecast(forecasts)
```
bremme, line 102: Whats the difference to `write_forecast` and `write_forecast_to_db`?

openstf/pipeline/create_component_forecast.py
JanMaartenvanDoorn: Stdev is one of the columns in the dataframe that is returned when loading the forecast inline 35

openstf/tasks/create_components_forecast.py
JanMaartenvanDoorn: These variables are used when determing the splitting coeficients, and should therefore be reused when making the component forcasts. I added a comment to make this more clear

openstf/tasks/create_components_forecast.py
JanMaartenvanDoorn: yes you are right I removed the optimum

openstf/tasks/create_components_forecast.py
JanMaartenvanDoorn: write_forecast_to_db is appparently renamed to write_forecast, i will change this

Issue #73, opened 2021-04-30
[Feature] Make a quantile model implementation that is compatible with the sklearn pipeline


JanMaartenvanDoorn:
**WHAT:** 
Currently only xgb models have been properly tested with the new pipeline. Here we implement a quantile version similar to the old pipeline.

**HOW:**
-	Decide/ check whether a multicolumn output for the predict method is compatible with sklearn.
o	If not compatible make sure the confidence interval applicator makes the proper quantile predictions
-	Make a new model based on the sklearn regressor
o	https://scikit-learn.org/stable/developers/develop.html 
-	Implement a fit method
-	Implement a predict method
-	Put the quantile model in the openstf/model folder
-	Test is new object is sklearn compatible
o	This can be easily done be running sklearn.utils.estimator_checks.check_estimator() on the new estimator

**TEST:**
-	Quantile model and forecasts are available in new pipeline.
- New Estimator object is sklearn compatible

FrankKr:
Is this issue completed?

PR #102, opened 2021-06-14
Feature #73 make quantile model

JanMaartenvanDoorn:
Added a quantile model that is compatible with the sklearn interface. The sklearn interface itself does not support multicolumn output. To make it compatible I added an optional quantile argument to the predict method with 0.5 as a default quantile. The quantile forecasts are added with the confidence interval applicator, nice thing about this solution is that most of the code could be reused and only minimal additions are nescesarry to make the pipeline work with quantile regression

All files in the pull request are relevant. However the most interesting changes are in the newly created openstf/model/xgb_quantile.py and in the openstf/model/confidence_interval_applicator.py


openstf/model/confidence_interval_applicator.py
```diff
@@ -7,19 +7,29 @@
...
+        self.forecast_input_data = forecast_input_data
 
-    def add_confidence_interval(
-        self, forecast: pd.DataFrame, quantiles: list
-    ) -> pd.DataFrame:
+    def add_confidence_interval(self, forecast: pd.DataFrame, pj: dict) -> pd.DataFrame:
         temp_forecast = self._add_standard_deviation_to_forecast(forecast)
-        return self._add_quantiles_to_forecast(temp_forecast, quantiles)
+
+        if pj["model"] == MLModelType.XGB_QUANTILE:
```
FrankKr, line 25: Zou het logisch zijn als het 'type-confidence-interval' een attribuut van het `model` zou zijn? Zodat je bij de definitie van een model opgeeft dat het bijv `model.confidence_type=quantile` oid. 
In plaats van hier een lijstje van modeltypes te definiëren.

openstf/model/confidence_interval_applicator.py
```diff
@@ -7,19 +7,29 @@
...
+
+from openstf.enums import MLModelType
 
 MINIMAL_RESOLUTION: int = 15  # Minimal time resolution in minutes
 
 
 class ConfidenceIntervalApplicator:
-    def __init__(self, model):
-        self.confidence_interval = model.confidence_interval
+    def __init__(self, model: RegressorMixin, forecast_input_data: pd.DataFrame):
```
bremme, line 18: Is er eigenlijk niet een nog hoge base class van sklearn? Iets van een `Estimator` oid. Voor nu denk wel prima verder.

openstf/model/confidence_interval_applicator.py
```diff
@@ -7,19 +7,29 @@
...
+        self.model = model
+        self.forecast_input_data = forecast_input_data
 
-    def add_confidence_interval(
-        self, forecast: pd.DataFrame, quantiles: list
-    ) -> pd.DataFrame:
+    def add_confidence_interval(self, forecast: pd.DataFrame, pj: dict) -> pd.DataFrame:
         temp_forecast = self._add_standard_deviation_to_forecast(forecast)
```
FrankKr, line 23: Een docstring over wat deze functie doet en wat de output is (welke kolommen bijv) zou mega chill zijn:)

openstf/model/confidence_interval_applicator.py
```diff
@@ -137,3 +147,24 @@ def _add_quantiles_to_forecast(
...
+    ) -> pd.DataFrame:
+        """Add quantiles to forecast.
+        Use trained quantile regression model to calculate the quantiles.
+        Args:
+            forecast (pd.DataFrame): Forecast
+            quantiles (list): List with desired quantiles
+        Returns:
+            (pd.DataFrame): Forecast DataFrame with quantile (e.g. 'quantile_PXX')
+                columns added.
+        """
```
FrankKr, line 162: Duidelijke docstring, fijn!

openstf/model/confidence_interval_applicator.py
```diff
@@ -7,19 +7,29 @@
...
-    ) -> pd.DataFrame:
+    def add_confidence_interval(self, forecast: pd.DataFrame, pj: dict) -> pd.DataFrame:
         temp_forecast = self._add_standard_deviation_to_forecast(forecast)
-        return self._add_quantiles_to_forecast(temp_forecast, quantiles)
+
+        if pj["model"] == MLModelType.XGB_QUANTILE:
+            return self._add_quantiles_to_forecast_quantile_regression(
+                temp_forecast, pj["quantiles"]
+            )
+        else:
```
bremme, line 29: Its a bit personal preference. But the `else` case is not required here since you have already returned from the `if`.

openstf/model/confidence_interval_applicator.py
```diff
@@ -137,3 +147,24 @@ def _add_quantiles_to_forecast(
...
         return forecast
+
+    def _add_quantiles_to_forecast_quantile_regression(
+        self, forecast: pd.DataFrame, quantiles: list
+    ) -> pd.DataFrame:
+        """Add quantiles to forecast.
+        Use trained quantile regression model to calculate the quantiles.
+        Args:
+            forecast (pd.DataFrame): Forecast
+            quantiles (list): List with desired quantiles
```
FrankKr, line 158: voor het 'quantiles' argument zou het helpen om een voorbeeldje te zien. is dit bijv 
`[1,10,90,99]` of `[0.01, 0.1, 0.9, 0.99]` of [P_01,P10,...]` ...

openstf/model/confidence_interval_applicator.py
JanMaartenvanDoorn: The Base estimator also exists: https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator. The type of the regressor is a combination of the regressor mixin and the base eatimator

openstf/model/confidence_interval_applicator.py
```diff
@@ -41,7 +51,7 @@ def _add_standard_deviation_to_forecast(
                 "forecast", "stdev"
         """
 
-        standard_deviation = self.confidence_interval
+        standard_deviation = self.model.confidence_interval
```
bremme, line 54: To be this is confusion. I would suggest we choose one name for one thing. Either we call is `standard_deviation` or we call it `confidence_interval`. Or do we really have a good reason to use a different name for the same thing?

openstf/model/confidence_interval_applicator.py
```diff
@@ -137,3 +147,24 @@ def _add_quantiles_to_forecast(
             )
 
         return forecast
+
+    def _add_quantiles_to_forecast_quantile_regression(
```
bremme, line 151: Good to add some docstring. A couple small suggestions, add new lines, indent and specify the quantiles data type

```python
        """Add quantiles to forecast.

            Use the standard deviation to calculate the quantiles.

        Args:
            forecast (pd.DataFrame): Forecast (should contain a 'forecast' + 'stdev' column)
            quantiles (List[float]): List with desired quantiles

        Returns:
            (pd.DataFrame): Forecast DataFrame with quantile (e.g. 'quantile_PXX')
                columns added.
        """
```

openstf/model/model_creator.py
```diff
@@ -6,21 +6,31 @@
...
         # check if model type is valid
-        if model_type not in [k.value for k in ModelCreator.MODEL_TRAINER_CONSTRUCTORS]:
-            raise KeyError(f"Unknown model type: {model_type}")
+        if pj["model"] not in [
+            k.value for k in ModelCreator.MODEL_TRAINER_CONSTRUCTORS
+        ]:
+            raise KeyError(f"Unknown model type: {pj['model']}")
 
-        return ModelCreator.MODEL_TRAINER_CONSTRUCTORS[MLModelType(model_type)]()
+        if MLModelType(pj["model"]) == MLModelType.XGB_QUANTILE:
```
FrankKr, line 31: Kunnen we nog een list verzinnen om hier niet een uitzondering voor dit specifieke model type te hoeven maken.?
Bijv dat de model_trainer_constructors ook **kwargs als input toestaan, waar ze niet per se iets mee hoeven te doen. 
En dat de return dan zoiets wordt (niet getest):

```
return ModelCreator.MODEL_TRAINER_CONSTRUCTORS[MLModelType(pj["model"])](
                quantiles=tuple(pj.get("quantiles"), None)
                )
```

openstf/model/xgb_quantile.py
```diff
@@ -0,0 +1,102 @@
...
+from xgboost import XGBRegressor
+import numpy as np
+
+import openstf.metrics.metrics as metrics
+
+DEFAULT_QUANTILES: Tuple[float, ...] = (0.9, 0.5, 0.1)
+
+
+class XgbQuantile(BaseEstimator, RegressorMixin):
+    def __init__(self, quantiles: Tuple[float, ...] = DEFAULT_QUANTILES):
```
FrankKr, line 18: Een beetje docstring met uitleg over wat het quantile model is, of een linkje naar een internetpagina oid zou chill zijn!

openstf/model/confidence_interval_applicator.py
```diff
@@ -137,3 +147,24 @@ def _add_quantiles_to_forecast(
...
+            quantiles (list): List with desired quantiles
+        Returns:
+            (pd.DataFrame): Forecast DataFrame with quantile (e.g. 'quantile_PXX')
+                columns added.
+        """
+
+        for quantile in quantiles:
+            quantile_key = f"quantile_P{quantile * 100:02.0f}"
+            forecast[quantile_key] = self.model.predict(
+                self.forecast_input_data.sort_index(axis=1), quantile=quantile
```
bremme, line 167: Why do we need to call `sort_index(axis=1)` here? Why is the input data not already in the right format? If it makes sense that the input data is not in the right format I would suggest to do this formatting in a separate step.

openstf/model/model_creator.py
```diff
@@ -6,21 +6,31 @@
...
 
 class ModelCreator:
+
     """Factory object for creating model trainer objects"""
 
     # Set object mapping
     MODEL_TRAINER_CONSTRUCTORS = {
         MLModelType.XGB: XGBRegressor,
         MLModelType.LGB: LGBMRegressor,
+        MLModelType.XGB_QUANTILE: XgbQuantile,
```
bremme, line 20: Would it not make more sense to call the the constructor `XGBQuantileRegressor` ?

FrankKr (commented):
Ziet er goed uit.
Heb nog wat suggesties ter overweging in de comments gezet

openstf/model/model_creator.py
```diff
@@ -6,21 +6,31 @@
...
-            raise KeyError(f"Unknown model type: {model_type}")
+        if pj["model"] not in [
+            k.value for k in ModelCreator.MODEL_TRAINER_CONSTRUCTORS
+        ]:
+            raise KeyError(f"Unknown model type: {pj['model']}")
 
-        return ModelCreator.MODEL_TRAINER_CONSTRUCTORS[MLModelType(model_type)]()
+        if MLModelType(pj["model"]) == MLModelType.XGB_QUANTILE:
+            return ModelCreator.MODEL_TRAINER_CONSTRUCTORS[MLModelType(pj["model"])](
+                tuple(pj["quantiles"])
```
bremme, line 33: Why do we need a tuple here? If you really need a tuple internally I would suggest to do the conversion in the constructor.

openstf/model/model_creator.py
```diff
@@ -6,21 +6,31 @@
...
+            )
+        else:
```
bremme, line 35: This is not really required since you already return on the `if`. This pattern is often called `return early` and is a general best practice and improves readability imho (see for example: http://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html)

openstf/model/model_creator.py
```diff
@@ -6,21 +6,31 @@
...
     # Set object mapping
     MODEL_TRAINER_CONSTRUCTORS = {
         MLModelType.XGB: XGBRegressor,
         MLModelType.LGB: LGBMRegressor,
+        MLModelType.XGB_QUANTILE: XgbQuantile,
     }
 
     @staticmethod
-    def create_model(model_type: MLModelType) -> RegressorMixin:
+    def create_model(pj: dict) -> RegressorMixin:
```
bremme, line 24: I'm not totally sure I like this interface better. The old interface was a bit more clean and the `prediction_job` is still a very loosely defined data structure (perhaps we can/should improve this). I guess an alternative solution would be:

```python
def create_model(model_type: MLModelType, *args, **kwargs):
  return ModelCreator.MODEL_TRAINER_CONSTRUCTORS[MLModelType(model_type)](**args, **kwargs)
```
We could in this case be more specific with the type hinting, I found this stackoverflow answer which could be helpful: https://stackoverflow.com/questions/37031928/type-annotations-for-args-and-kwargs

openstf/model/xgb_quantile.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # Check if quantile 0.5 is pressent this is required
+        if 0.5 not in quantiles:
+            raise ValueError(
+                "Cannot train quantile model as 0.5 is not in requested quantiles!"
+            )
+        else:
```
bremme, line 25: This `else` is not required here.

openstf/model/xgb_quantile.py
```diff
@@ -0,0 +1,102 @@
...
+from functools import partial
+
+from sklearn.base import RegressorMixin, BaseEstimator
+from sklearn.utils.validation import check_X_y, check_array, check_is_fitted
+from xgboost import XGBRegressor
+import numpy as np
+
+import openstf.metrics.metrics as metrics
+
+DEFAULT_QUANTILES: Tuple[float, ...] = (0.9, 0.5, 0.1)
```
bremme, line 14: Why are the `...` required here? (probably just my knowledge of these dots)

openstf/model/xgb_quantile.py
```diff
@@ -0,0 +1,102 @@
+# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
+#
+# SPDX-License-Identifier: MPL-2.0
+from typing import Tuple
+from functools import partial
+
+from sklearn.base import RegressorMixin, BaseEstimator
+from sklearn.utils.validation import check_X_y, check_array, check_is_fitted
+from xgboost import XGBRegressor
```
bremme, line 9: Should we not use the sklearn `XGBRegressor` here?

openstf/model/xgb_quantile.py
```diff
@@ -0,0 +1,102 @@
...
+import numpy as np
+
+import openstf.metrics.metrics as metrics
+
+DEFAULT_QUANTILES: Tuple[float, ...] = (0.9, 0.5, 0.1)
+
+
+class XgbQuantile(BaseEstimator, RegressorMixin):
+    def __init__(self, quantiles: Tuple[float, ...] = DEFAULT_QUANTILES):
+
```
bremme, line 19: I think we should initialize all attributes in the constructor for more clarity. So in this case `estimators_` and `is_fitted_` or are those `RegressorMixin` attributes?

openstf/model/xgb_quantile.py
```diff
@@ -0,0 +1,102 @@
...
+            specific_quantile_model = XGBRegressor()
+
+            # Set hyperparameters
+            specific_quantile_model.set_params(params=params_quantile)
+
+            # Train model for this specific quantile
+            quantile_models[quantile] = specific_quantile_model.fit(x, y)
+
+        # Update state of the estimator
+        self.estimators_ = quantile_models
```
bremme, line 71: Are these `RegressorMixing` attributes? Why is the `_` postfix used? To prevent a clash?

openstf/model/xgb_quantile.py
```diff
@@ -0,0 +1,102 @@
...
+        self.is_fitted_ = True
```
bremme, line 72: If this is not some base class attribute it should be initialized to `False` in the constructor.

openstf/model/xgb_quantile.py
```diff
@@ -0,0 +1,102 @@
...
+
+        return self
+
+    def predict(self, x: np.array, quantile: float = 0.5) -> np.array:
+        """Makes a prediction for a desired quantile
+
+        Args:
+            x np.array: Feature matrix
```
bremme, line 80: Try to stay consistent with the docstrings, the datatype should be in brackets.

openstf/model/xgb_quantile.py
```diff
@@ -0,0 +1,102 @@
...
+        Raises:
+            ValueError in case no model is trained for the requested quantile
+
+        """
+        # Check if model is trained for this quantile
+        if quantile not in self.quantiles:
+            raise ValueError("No model trained for requested quantile!")
+
+        # Check/validate input
+        check_array(x)
```
bremme, line 96: Dot these checks raise additional exceptions? If so it should be in the docstring ideally.

test/unit/model/prediction/xgboost/test_xgb_quantile.py
```diff
@@ -0,0 +1,66 @@
...
+        # More info: https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html
+        check_estimator(XgbQuantile(tuple(self.quantiles)))
+
+    def test_quantile_loading(self):
+        model = XgbQuantile(tuple(self.quantiles))
+        self.assertEqual(model.quantiles, tuple(self.quantiles))
+
+    def test_value_error_raised(self):
+        # Check if Value Error is raised when 0.5 is not in the requested quantiles list
+        with pytest.raises(ValueError):
```
bremme, line 42: I think we should not fix two strategies. I believe we chose the unittest framework for our unit test. In that case you should use:
```python
with self.assertRaises(ValueError):
  XgbQuantile((0.2, 0.3, 0.6, 0.7))
```

test/unit/model/prediction/xgboost/test_xgb_quantile.py
```diff
@@ -0,0 +1,66 @@
...
+            XgbQuantile((0.2, 0.3, 0.6, 0.7))
+
+    def test_value_error_raised_no_model_trained_for_quantile(self):
+        # Test if value error is raised when model is not available
+        with pytest.raises(ValueError):
```
bremme, line 47: see above.

test/unit/model/prediction/xgboost/test_xgb_quantile.py
```diff
@@ -0,0 +1,66 @@
...
+    def test_quantile_loading(self):
+        model = XgbQuantile(tuple(self.quantiles))
+        self.assertEqual(model.quantiles, tuple(self.quantiles))
+
+    def test_value_error_raised(self):
+        # Check if Value Error is raised when 0.5 is not in the requested quantiles list
+        with pytest.raises(ValueError):
+            XgbQuantile((0.2, 0.3, 0.6, 0.7))
+
+    def test_value_error_raised_no_model_trained_for_quantile(self):
```
bremme, line 45: I believe the convention is `test_<function_name>_<test_description>` so here it should be `test_predict_raises_valueerror_no_model_trained_for_quantile` or something similar. You can always suggest a different convention, but I think we should aim for consistency.

test/unit/model/prediction/xgboost/test_xgb_quantile.py
```diff
@@ -0,0 +1,66 @@
...
+    def setUp(self) -> None:
+        self.quantiles = [0.9, 0.5, 0.6, 0.1]
+
+    def test_sklearn_compliant(self):
+        # Use sklearn build in check, this will raise an exception if some check fails
+        # During these tests the fit and predict methods are elaborately tested
+        # More info: https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html
+        check_estimator(XgbQuantile(tuple(self.quantiles)))
+
+    def test_quantile_loading(self):
```
bremme, line 36: I'm not sure this test really adds anything actually.

test/unit/model/prediction/xgboost/test_xgb_quantile.py
```diff
@@ -0,0 +1,66 @@
...
+        with pytest.raises(ValueError):
+            XgbQuantile((0.2, 0.3, 0.6, 0.7))
+
+    def test_value_error_raised_no_model_trained_for_quantile(self):
+        # Test if value error is raised when model is not available
+        with pytest.raises(ValueError):
+            model = XgbQuantile((0.2, 0.3, 0.5, 0.6, 0.7))
+            model.predict("test_data", quantile=0.8)
+
+    def test_add_quantiles_to_forecast(self):
```
bremme, line 51: Does this unit test belong here? It seems a test for the ConfidenceIntervalApplicator.

test/unit/model/prediction/xgboost/test_xgb_quantile.py
```diff
@@ -0,0 +1,66 @@
+# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
```
bremme, line 1: This test seems to be in the wrong location? It should be `test/unit/model/test_xgb_quantile.py` given the current layout.

test/unit/model/prediction/xgboost/test_xgboost.py
```diff
@@ -15,13 +15,13 @@ class MockModel:
 
 
 class TestXGBPredictionModel(BaseTestCase):
-    def test_add_quantiles_to_forecast(self):
+    def test_add_quantiles_to_forecast_default(self):
```
bremme, line 18: This looks like a ConfidenceIntervalApplicator test ?

test/unit/model/test_model_creator.py
```diff
@@ -0,0 +1,37 @@
...
+    def setUp(self) -> None:
+        self.pj = {"model": "xgb", "quantiles": [0.5, 0.2, 0.5]}
+
+    def test_happy_flow(self):
+        # Test happy flow
+        model = ModelCreator.create_model(self.pj)
+
+        self.assertIsInstance(model, XGBRegressor)
+
+    def test_quantile_model(self):
```
bremme, line 24: convention -> `test_<function_name>_<test_description>`

test/unit/model/test_model_creator.py
```diff
@@ -0,0 +1,37 @@
...
+        # Test if quantile model is properly returned
+        self.pj["model"] = "xgb_quantile"
+        # Create relevant model
+        model = ModelCreator.create_model(self.pj)
+
+        self.assertIsInstance(model, XgbQuantile)
+        self.assertEqual(model.quantiles, tuple(self.pj["quantiles"]))
```
bremme, line 31: This is not related to the model creator? Its not responsible for this.

test/unit/model/test_model_creator.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_unknown_model(self):
```
bremme, line 33: to me this is confusion. The `model` prooperty of a prediction job is actually the model type. I think we should use that name to prevent confusion.

test/unit/model/test_model_creator.py
```diff
@@ -0,0 +1,37 @@
...
+        # Test if keyerror is raised when model type is unknown
+        self.pj["model"] = "Unknown"
+        with pytest.raises(KeyError):
```
bremme, line 36: Use unittest framework here as well.

bremme:
I like the general implementation! Nice way to implement the quantile model and stay sklearn compatible. Try to stick to our conventions and be consistent. Perhaps it should not take so much time to do the reviews and point out all the little things. If our conventions are not clear we should probably make it more clear to prevent spending time on it over and over.

openstf/model/confidence_interval_applicator.py
JanMaartenvanDoorn: I agree this is a bit confusing. We have i fact two ways to make a confidence interval: quantiles and the stdev. I renamed everything that only makes/applies the stdev to standard_deviation(including the example above). The confidence_interval_applicator is more broader as it also adds the quantiles, I keeped that name.

openstf/model/model_creator.py
JanMaartenvanDoorn: good idea, this just works

openstf/model/model_creator.py
JanMaartenvanDoorn: This is because of sklearn requirements, they want inputs in the __init__ to be imutable and not be muted in the __init__ itself.

openstf/model/xgb_quantile.py
JanMaartenvanDoorn: These are requirements by sklearn. If you add these parameters to the __init__ the compatibility check will not pass. The "_" postfix is the sklearn way of defining private attributes.

openstf/model/xgb_quantile.py
JanMaartenvanDoorn: The "_" postfix is the sklearn way of defining private attributes.

openstf/model/xgb_quantile.py
JanMaartenvanDoorn: This is a base class attribute

openstf/model/xgb_quantile.py
JanMaartenvanDoorn: Yes but I dont know which ones in which cases

test/unit/model/prediction/xgboost/test_xgb_quantile.py
JanMaartenvanDoorn: I had some problems with the quantiles loading correctly, I used this to check if the issue was fixed (which it is).

test/unit/model/test_model_creator.py
JanMaartenvanDoorn: Here I also had an issue with the quantiles are being properly passed into the Regressor, this test checks if this works correctly

test/unit/model/test_model_creator.py
JanMaartenvanDoorn: yes, but this also requires renaming the field in the mysql database

openstf/model/model_creator.py
JanMaartenvanDoorn: I disagree here, with the above construction you would need to pass multiple entries of the prediciton jobs as arguments (model type and quantiles). I think it is better to work on a better defenition of the prediction jobs to make the interface more clear.

openstf/model/confidence_interval_applicator.py
JanMaartenvanDoorn: I disgree here, I think the prediction job should contain the configuration on what we want to happen for a specific prediction/ model. If we also add an attribute to the model we are moving away from the centralized config in the prediction job. I agree that keeping a list here on which confidence interval should be made for which model_type here is not elegant. Maybe it is better to keep the list somewhere else and generate a "confidence_interval_type" entry in the prediction job based on that list.

openstf/model/xgb_quantile.py
JanMaartenvanDoorn: https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_array.html sklearn documentation is also not clear on which Exceptions are raised

openstf/model/confidence_interval_applicator.py
bremme: @FrankKr does the model_type (which we call "model") not imply the confidence_type as well, e.g. a `xgb_quantile` model always has a confidence_type of quantile. If so, I think its better to just keep using the model_type to do the selection.

@JanMaartenvanDoorn if `pj["model"]` already an enum? If not you should either convert it or compare to `MLModelType.XGB_QUANTILE.value`

openstf/model/model_creator.py
bremme: That would solve a lot of long lasting issues :) and sound like a good solution as well. Probably a better solution in general. But right now we went from a good defined interface to a vaguely defined interface.

bremme: It does look like the current commit is not valid: 
image
Not all constructors should take the quantiles argument right. And I still think the caller of the XGBQuantileRegressor should not be responsible to do this tuple conversion. We should either always use a tuple (but there are many more cases where we could use a tuple instead of a list) or just trust the developer by using plain lists or let the constructor handle this if we really think its required.

test/unit/model/test_model_creator.py
bremme: We don't have to do everything at once. We can do the conversion first in the database interface and later in the database if we want.

openstf/model/confidence_interval_applicator.py
FrankKr: @JanMaartenvanDoorn I don't think I completely understand your comment. The configuration of the type of confidence interval is not defined in the prediction job, but based on the modeltype, right?

Suppose in the future we would include for example and LGB_QUANTILE model. 
I think it would be elegant if then only in one place in the project code has to be added; in the code implementing that model type. Currently, it looks like also in other places this new model type has to be added to an if-check.
Perhaps we could also find somewhere in between that we check if the ModelType ends in _QUANTILE and agree that the name of quantile-type models should always end in quantile?

bremme: @FrankKr if we have multiple variations of quantile models (in the future) which can use the same implementation to calculate the quantiles it would indeed be nice to make the selection on a shared attribute. There are different strategies to accomplish such a thing. I would not be too bothered to add a line or two for a new model_type (I will not happen that often I guess). But we could also use an  additional property on the prediction_job, .e.g model_class | model_group | model_type_class | model_type_group -> "quantile" or something similar. To treat a group of model_types in a certain way.

bremme: Looking at the `ConfidenceIntervalApplicator` again. The `model` (RegressorMixin) is already an attribute. We could also make the selection based on a base class perhaps. But then we have to wrap a common base around the sklearn implementations I guess..

PR #104, opened 2021-06-17
Feature KTPS-1435 sklearn optimize hyperparameters pipeline

bremme:
For this story I have implemented a sklearn compatible hyperparameter optimization pipeline. I have chosen for the following architecture: 

* Like the `create_forecast` and `train_model` pipelines there is no database connectivity in the pipeline
* All database connectivty is moved to the `task`
* I added a `ObjectiveCreator` which uses our familiar factory pattern to create an objective (function) for a given `model_type`
  * I've chosen to use the "callable classes" implementation as suggested by the optuna documentation https://optuna.readthedocs.io/en/stable/faq.html#objective-func-additional-args
  * This seems a nice and flexible solution to pass additional argument to the objective function
* I add a callback to the study to do some (optional) debug logging
* I also made sure that the internal optuna logger is passed on to our structlog logger
* In the objective implementation I used the sklearn interface to fit and predict
* I added the `XGBoostPruningCallback` for additional pruning optimization

Let me know what you think and where I can still improve this PR.

> NOTE I have not implemented feature set and or training period optimization yet since it was not described in the issue #104 

# Questions

* Should the optimize hyperparameters pipeline also do validation like the train pipeline:
image


openstf/model/objective.py
```diff
@@ -0,0 +1,105 @@
...
+from openstf.metrics.metrics import mae
+
+EARLY_STOPPING_ROUNDS: int = 10
+TEST_FRACTION: float = 0.1
+VALIDATION_FRACTION: float = 0.1
+# See https://xgboost.readthedocs.io/en/latest/parameter.html for all possibilities
+EVAL_METRIC: str = "mae"
+
+# https://optuna.readthedocs.io/en/stable/faq.html#objective-func-additional-args
+class XGBRegressorObjective:
```
JanMaartenvanDoorn, line 18: Nice to have these links in the documentation

openstf/model/objective.py
```diff
@@ -0,0 +1,105 @@
...
+            # initialize the objective function
+            objective = XGBRegressorObjective(input_data, test_fraction)
+            # use the objective function
+            study.optimize(objective)
+    """
+
+    MODEL_TYPE = MLModelType.XGB
+
+    def __init__(
+        self,
```
JanMaartenvanDoorn, line 35: Here you could add type hints as well

openstf/model/objective_creator.py
```diff
@@ -0,0 +1,19 @@
...
+from openstf.enums import MLModelType
+from openstf.model.objective import XGBRegressorObjective
+
+
+class ObjectiveCreator:
+    OBJECTIVES = {
+        MLModelType.XGB: XGBRegressorObjective
+    }
+    @staticmethod
+    def create_objective(model_type: Union[MLModelType, str]):
```
JanMaartenvanDoorn, line 16: Nice to have type ints for the input, for the output you could define is as well

openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -0,0 +1,91 @@
...
+# See https://optuna.readthedocs.io/en/stable/reference/generated/optuna.study.Study.html#optuna.study.Study.optimize
+N_TRIALS: int = 8   # The number of trials.
+TIMEOUT: int = 200  # Stop study after the given number of second(s).
+
+def optimize_hyperparameters_pipeline(
+    pj: dict,
+    input_data: pd.DataFrame,
+    n_trials: int = N_TRIALS,
+    timeout: int = TIMEOUT
+):
```
JanMaartenvanDoorn, line 25: output type hints could be added here

openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -0,0 +1,91 @@
...
+    optimized_error = study.best_value
+
+    logger.info(
+        f"Finished hyperparameter optimization, error objective {optimized_error} "
+        f"and params {optimized_hyperparams}"
+    )
+
+    return optimized_hyperparams
+
+def _log_study_progress(study, trial):
```
JanMaartenvanDoorn, line 80: Nice to make this compatible, this probably will also help with imporving monitoring

openstf/tasks/optimize_hyperparameters.py
```diff
@@ -13,22 +13,68 @@
...
 from openstf.tasks.utils.predictionjobloop import PredictionJobLoop
 from openstf.tasks.utils.taskcontext import TaskContext
+from openstf.monitoring import teams
+
+MAX_AGE_HYPER_PARAMS_DAYS = 31
+
+
+def optimize_hyperparameters_task(pj: dict, context: TaskContext) -> None:
+
+    db = context.database
```
JanMaartenvanDoorn, line 31: A matter of taste I quess but defining db her is not strictly nescesarry. I personally would use context.database directly instead of db

JanMaartenvanDoorn (commented):
Overall I very much like the implementation, it is clean and simple yet more robust than the old implementation. One point to consider: The objective class is now in name quite specific for xgboost while there are only a few lines of code that are realy specific for xgboost. Do you have any ideas on how we micht make a more general implementation so we have to add minimal extra code while implementing another model? Furthermore I found some minor points where typehints could be improved.

openstf/model/objective.py
```diff
@@ -0,0 +1,105 @@
...
+        input_data,
+        test_fraction=TEST_FRACTION,
+        validation_fraction=VALIDATION_FRACTION,
+        eval_metric=EVAL_METRIC,
+        verbose=False
+    ):
```
JanMaartenvanDoorn, line 41: output typehints

openstf/model/objective.py
```diff
@@ -0,0 +1,105 @@
...
+        self.input_data = input_data
+        self.test_fraction = test_fraction
+        self.validation_fraction = validation_fraction
+        self.eval_metric = eval_metric
+        self.verbose = verbose
+
+    def __call__(self, trial):
```
JanMaartenvanDoorn, line 48: input and output type hints

openstf/pipeline/optimize_hyperparameters.py
JanMaartenvanDoorn: input type hints

test/unit/pipeline/test_optimize_hyperparameters.py
```diff
@@ -0,0 +1,34 @@
+
+import unittest
```
JanMaartenvanDoorn, line 2: Also unittests need license info headers

openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -0,0 +1,91 @@
...
+            f"Input data is insufficient for {pj['name']} "
+            f"after validation and cleaning"
+        )
+
+    input_data_with_features = TrainFeatureApplicator(horizons=[0.25, 24]).add_features(
+        input_data
+    )
+
+    # Create relevant model
+    Objective = ObjectiveCreator.create_objective(model_type=pj["model"])
```
JanMaartenvanDoorn, line 56: Objective should be objective (this also generates a code smell)

openstf/tasks/optimize_hyperparameters.py
```diff
@@ -13,22 +13,68 @@
...
+            "Skip hyperparameter optimization",
+            pid=pj["id"],
+            last_optimized_days=last_optimized_days,
+            max_age=MAX_AGE_HYPER_PARAMS_DAYS
+        )
+        return
 
+    # Get input data
+    current_hyperparams = db.get_hyper_params(pj)
+    # FIXME this conversion should be done in the database
```
JanMaartenvanDoorn, line 48: This FIXME generates a code smell, better to use TODO

openstf/model/objective.py
bremme: thx. I like to add links if its not obvious and or if I think I need to look it up later perhaps.

openstf/model/objective.py
bremme: I'm not totally sure how useful that is in this case. It really obvious that a constructor always return an instance of the class. Moreover, I'm not even sure its possible? My editor complaints since the class is not yet defined:
image

openstf/model/objective_creator.py
bremme: thx, done!

openstf/pipeline/optimize_hyperparameters.py
bremme: done

openstf/pipeline/optimize_hyperparameters.py
bremme: done

openstf/tasks/optimize_hyperparameters.py
bremme: Its not necessary indeed. I've implemented your suggestion!

test/unit/pipeline/test_optimize_hyperparameters.py
bremme: thx. fixed!

openstf/model/objective.py
bremme: fixed.

openstf/pipeline/optimize_hyperparameters.py
bremme: hmm..okay. Since Objective is a class and not an instance of a class to me it makes sense to use a capital here. But I can also call it `objective_constructor` or `objective_class` to prevent the smell.

openstf/tasks/optimize_hyperparameters.py
bremme: But its a valid point imo. By having a smell we have a clear reminder that we need to fix this. On the other hand we don't have a clear system to use these terms `FIXME`, `TODO`, `NOTE` etc. So perhaps nicer to not have the smell and hopefully we will fix it eventually.

openstf/pipeline/optimize_hyperparameters.py
bremme: I used `objective` and added a comment that this is a class and not an instance to prevent confusion.

openstf/model/objective.py
bremme: hmm..I though I already replied to this comment? I'm not sure if this is required and or even possible. Its really obvious that the output of a `RegressorObjective` constructor is a `RegressorObjective`. Furthermore, it looks like you can't even do this since you are referring to a class which is not yet fully defined.

bremme:
@JanMaartenvanDoorn I've splitter the objective function into two classes a generic `RegressorObjective` and a more specific `XGBRegressorObjective`. We can always improve this once we add more objectives of course. But for now that's perhaps out of scope.

Issue #89, opened 2021-05-26
[Feature] Integrate monitoring in new pipeline


JanMaartenvanDoorn:
WHAT: In the old pipeline a report is generated about the number of succesfull jobs. This functionality is mostly implemented in the prediction job loop. However it relies on a few functions inside the pipeline. Add this funcitonality to the new pipeline. This story also includes the performance monitor.

HOW:
- Make sure implementation is modular
-  Make sure the pipeline could also be run in a notebook without the monitoring option
- Probably the prediction loop itself does not need to be modified
-  Sugestion: use callback function in the pipelines

TEST:
- (Teams) monitoring reports are available with the new pipeline
- After running a prediction job loop (train_model_task, create_forecast, optimize_hyperparameters) a report is send with information on which pj's failed and how many were succesfull.

FrankKr:
I've started working on this issue

PR #105, opened 2021-06-18
Feature #89 ktps 1426 sklearn pipeline monitoring

FrankKr:
Fixes https://github.com/alliander-opensource/short-term-forecasting/issues/89

Concluded that the 'context-specific' steps that used to be in the train_model_core aren't required.
I have left them out, and added a few minor checkpoints on a higher level in the pipeline.

Most work of this story was to include unit tests to test the pipeline:
1) store a new trained model (useful for when the interface of models changes)
2) run the create_forecast_pipeline with only mocking ktpbase-stuff and actually using a previously stored model. Also, here I added a test to check the MAE of the forecast versus the actual load. Perhaps it would make sense to move this test to 'test/regression' or something like that, but for now, this seems Okay. Running the test is very fast, so that is not a reason to move it.

Other minor improvements:
- change Contribution description of workflow from gitflow to github flow
- added some logging on where stuff is stored
- added try/except if old model cannot be compared to new model
- remove ktpbase.logging dependency in taskcontext
- stored a trained model with new interface (e.g. including model.standard_deviation) in unit test folder


CONTRIBUTING.md
```diff
@@ -35,7 +35,7 @@ Furthermore the following conventions apply:
 
 ## Git branching
 
-This project uses the [Gitflow Workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) and branching model. The `master` branch always contains the latest release, after a release is made new feature branches are branched of `develop`. When a feature is finished it is merged back into `develop`. At the end of a sprint `develop` is merged back into `master` or (optional) into a `release` branch first before it is merged into `master`.
+This project uses the [GitHub flow Workflow](https://guides.github.com/introduction/flow/) and branching model. The `main` branch always contains the latest release. New feature branches are branched from `main`. When a feature is finished it is merged back into `main` via a [Pull Request](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#:~:text=Pull%20requests%20let%20you%20tell,merged%20into%20the%20base%20branch.).
```
JanMaartenvanDoorn, line 38: Good to update this.

openstf/tasks/train_model.py
```diff
@@ -52,6 +52,8 @@ def train_model_task(pj: dict, context: TaskContext) -> None:
         pj["hyper_params"]["featureset_name"]
     )
 
+    context.checkpoint("Meta-data")
```
JanMaartenvanDoorn, line 55: Maybe: "Retrieved meta-data" or "Retrieved prediction job"

openstf/tasks/train_model.py
```diff
@@ -66,6 +68,8 @@ def train_model_task(pj: dict, context: TaskContext) -> None:
         datetime_end=datetime_end,
     )
 
+    context.checkpoint("TimeSeries input")
```
JanMaartenvanDoorn, line 71: Maybe "Retrieved timeseries input" is better here as indicates more accurately where we are in the procces.

openstf/pipeline/train_model.py
```diff
@@ -159,25 +159,26 @@ def train_model_pipeline_core(
...
-    if old_model is not None:
+    if old_model:
         combined = train_data.append(validation_data)
         x_data, y_data = combined.iloc[:, 1:], combined.iloc[:, 0]
 
         # Score method always returns R^2
         score_new_model = model.score(x_data, y_data)
-        score_old_model = old_model.score(x_data, y_data)
+        try:
+            score_old_model = old_model.score(x_data, y_data)
```
JanMaartenvanDoorn, line 169: Just being curious: Why is the try, catch  required here?

JanMaartenvanDoorn (approved):
Nice additions! Some minor comments regarding the phrasing of the check points.

openstf/tasks/train_model.py
FrankKr: good suggestion, I'll add it

openstf/tasks/train_model.py
FrankKr: Good suggestion, I'll add it

openstf/pipeline/train_model.py
FrankKr: if the old model has different feature names compared to the new model, I'll add it as a comment

Issue #90, opened 2021-05-26
[Feature] Make tracy compatible with new pipeline


JanMaartenvanDoorn:
WHAT: To manually train a model or optimize hyper parameters we have tracy. Tracy still calls the old pipelines. Change this to the new pipelines.

HOW:
- have a look at the tracy code (task: run_tracy)
- change the "train_model_for_specific_pid" and the "optimize_hyper_parameters_for_specifc_pid" function such that they use the new pipeline
- Check if the "allowed functions" in tracy are still up to date

TEST:
- tracy retrains models with new pipeline
- tracy optimizes hyper parameters with new pipeline

FrankKr:
I think Bram started working on this
https://github.com/alliander-opensource/short-term-forecasting/tree/feature-KTPS-1428-make-tracy-compatible-with-sklearn-pipelines

FrankKr:
fixed by #106

PR #106, opened 2021-06-22
Feature KTPS-1428 make tracy compatible with sklearn pipelines

bremme:
This PR should solve issue #90.

In short, this PR makes the `run_tracy` task compatible with the our new sklearn compatible pipelines.

While working on this story I've also tried to simplified the code and reduced the task by around 50 lines. The general strategy is as follows:

* Tracy calls the `train_model_task` or the `optimize_hyperparameters_task`
  * It made most sense to call into the `task` instead of the pipelines to not duplicate all the task code
* Tracy always performs a task on a prediction job (the argument is always the pid)
* There are three possible outcomes of when tracy runs a job
  * job is successful 
  * job is unkown (name is not known)
  * job fails (an exception in the task/pipeline occured)
* I've also added some unittest.

Let me know what you think and if my simplification is useful or if I have removed important functionality perhaps.


test/unit/model/test_objective.py
```diff
@@ -20,7 +20,7 @@ def test_call(self, model_creator_mock, mea_mock, split_data_mock):
 
         objective = XGBRegressorObjective(input_data)
         trial_mock = MagicMock()
-        error = objective(trial_mock)
+        objective(trial_mock)
```
JanMaartenvanDoorn, line 24: Is this part of the Tracy PR or the hyperparameters PR?

JanMaartenvanDoorn (approved):
Verry nice implementation! Looks much clearer this way. Also nice unit tests.

test/unit/model/test_objective.py
bremme: No not really. This was a small/bug leftover which is already solved upstream.

PR #107, opened 2021-06-22
Feature ktps 1422 back test

JanMaartenvanDoorn:
Implemented a backtest train and predict pipeline. Also fixed a bug with sorting of columns. These new changes made the BackTestFeature applicator redundant so I removed it.


openstf/feature_engineering/feature_applicator.py
```diff
@@ -11,6 +11,7 @@
 from openstf.feature_engineering.general import (
     add_missing_feature_columns,
     remove_extra_feature_columns,
+    enforce_feature_order,
```
FrankKr, line 14: Good plan to make a separate function for this

openstf/model/objective.py
```diff
@@ -63,9 +63,9 @@ def __call__(self, trial: optuna.trial.FrozenTrial) -> float:
             back_test=True,
         )
         # Split in x, y data (x are the features, y is the load)
-        train_x, train_y = train_data.iloc[:, 1:], train_data.iloc[:, 0]
-        valid_x, valid_y = validation_data.iloc[:, 1:], validation_data.iloc[:, 0]
-        test_x, test_y = test_data.iloc[:, 1:], test_data.iloc[:, 0]
+        train_x, train_y = train_data.iloc[:, 1:-1], train_data.iloc[:, 0]
+        valid_x, valid_y = validation_data.iloc[:, 1:-1], validation_data.iloc[:, 0]
+        test_x, test_y = test_data.iloc[:, 1:-1], test_data.iloc[:, 0]
```
FrankKr, line 68: Since order of columns is very important here, perhaps it's a good idea to add a check and make sure that the last columns is 'Horizon' and should be omitted

openstf/pipeline/create_forecast.py
```diff
@@ -72,13 +72,18 @@ def create_forecast_pipeline_core(
...
-    ).add_features(validated_data)
+    data_with_features = (
+        OperationalPredictFeatureApplicator(
+            # TODO use saved feature_names (should be saved while training the model)
+            horizons=[0.25],
+            feature_names=model._Booster.feature_names,
+        )
+        .add_features(validated_data)
+        .drop(columns="load")
+    )
```
FrankKr, line 83: Shouldnt 'data_with_features' also have the target-column (e.g. load)?

If not, perhaps another name for the variable would make more sense ('e.g. features or forecast_input or something else)

openstf/pipeline/train_model.py
```diff
@@ -135,15 +135,18 @@ def train_model_pipeline_core(
...
     model = ModelCreator.create_model(pj)
 
     # split x and y data
-    train_x, train_y = train_data.iloc[:, 1:], train_data.iloc[:, 0]
-    validation_x, validation_y = validation_data.iloc[:, 1:], validation_data.iloc[:, 0]
+    train_x, train_y = train_data.iloc[:, 1:-1], train_data.iloc[:, 0]
+    validation_x, validation_y = (
+        validation_data.iloc[:, 1:-1],
+        validation_data.iloc[:, 0],
+    )
```
FrankKr, line 149: Here also, perhaps wise to add a check on column names to make sure the last column should be excluded

openstf/pipeline/train_predict_backtest.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+    # Add features
+    data_with_features = TrainFeatureApplicator(
+        horizons=training_horizons, feature_names=pj["feature_names"]
+    ).add_features(validated_data)
+
+    # Split data
+    train_data, validation_data, test_data = split_data_train_validation_test(
+        data_with_features.sort_index(axis=0), test_fraction=0.15, back_test=True
```
FrankKr, line 70: 'backtest' is one word

openstf/pipeline/train_predict_backtest.py
```diff
@@ -0,0 +1,135 @@
...
+    serializer = PersistentStorageSerializer(trained_models_folder)
+    serializer.save_model(model, pid=pj["id"])
+
+    # Save reports/figures
+    report.save_figures(save_path=save_figures_folder)
+
+    # Load most recent model for the given pid
+    model = PersistentStorageSerializer(
+        trained_models_folder=trained_models_folder
+    ).load_model(pid=pj["id"])
```
FrankKr, line 114: Do you think it's needed to save a model and reports when performing a backtest?
Possibly we can do without storing a model and subsequently loading it back in when making the backtest.
Perhaps storing the model and reports is nice feature, but then I would not use the default locations, but for example to require this to be specified as an argument in the function.
What do you think?

openstf/pipeline/train_predict_backtest.py
```diff
@@ -0,0 +1,135 @@
...
+    forecast = add_prediction_job_properties_to_forecast(
+        pj,
+        forecast,
+    )
+
+    # Add column with realised load and horizon information
+    forecast["realised"] = test_data.iloc[:, 0]
+    forecast["Horizon"] = test_data.iloc[:, -1]
+
+    return forecast
```
FrankKr, line 135: Should we also include some method for evaluation of the backtest. E.g. a report on the performance of the model?
Can be very lightweight.

I think you can reuse some of the plot functions from https://github.com/Alliander/icarus-ops-gui/blob/master/app/pages/predictability.py and maybe also from the already available 'Reporter'

FrankKr (commented):
Looks like a good setup.

I've added some comments throughout.

Also, Sonarcloud notices a high % of duplication. Can you tackle this by generating reusable functions from code blocks which are identical in some pipelines?

openstf/feature_engineering/general.py
```diff
@@ -67,6 +67,10 @@ def remove_extra_feature_columns(
 
     extra_features = [f for f in list(input_data) if f not in features]
 
+    # Do not see "load" as an extra feature as it is no feature
+    if "load" in extra_features:
```
bremme, line 71: I'm not sure if this should be the responsibility of this function. Did this actually happen by the way? This looks a bit like solving a bug/error somewhere else silently. In my opinion both is not a really good idea. But perhaps I'm missing something?

openstf/feature_engineering/general.py
```diff
@@ -76,3 +80,37 @@ def remove_extra_feature_columns(
         )
 
     return input_data.drop(extra_features, axis=1)
+
+
+def enforce_feature_order(df):
```
bremme, line 85: I'm thinking about if `enforce_feature_order` is the right name for what this function does. Perhaps `sort_features` or `sort_feature_columns` is more descriptive of what it does? I guess it does enforce the order, but it does that by sorting.

openstf/feature_engineering/general.py
```diff
@@ -76,3 +80,37 @@ def remove_extra_feature_columns(
...
+    """Enforces correct order of features.
+
+    Alphabetically orders the feature columns. The load column remains the first column
+        and the Horizons column remains the last column. Everything in between is alphabetically sorted.
```
bremme, line 89: This line seems a little long (minor issue ofc).

openstf/feature_engineering/general.py
```diff
@@ -76,3 +80,37 @@ def remove_extra_feature_columns(
...
+
+    # Sort columns
+    columns = list(np.sort(df.columns.to_list()))
+
+    # Remove first column and add to the start
+    columns.remove(first_col_name)
+    col_order = [first_col_name] + columns
+
+    # If "Horzion" column is available add to the end
+    if (
```
bremme, line 109: suggestion, use a single line: `if "Horizon" in columns:`

openstf/feature_engineering/general.py
bremme, line 85: More in general. Why do we need this function, could we not just make sure that our input_data is always ordered properly instead of fixing it afterwards? Which part of the code determines the original order and why is that different from what we need later on?

openstf/feature_engineering/general.py
```diff
@@ -76,3 +80,37 @@ def remove_extra_feature_columns(
...
+
+    Returns:
+        pd.DataFrame with features with all columns in the correct order
+    """
+
+    # Extract first column name
+    first_col_name = df.columns.to_list()[0]  # Most of the time this is load
+
+    # Sort columns
+    columns = list(np.sort(df.columns.to_list()))
```
bremme, line 102: I would suggest to use the same convention within a function. So you either use:

`first_col_name` and `col_names` OR `first_column` and `columns` OR (most descriptive) `first_column_name` and `column_names`.

Now it's a bit mixed up.

openstf/feature_engineering/general.py
```diff
@@ -76,3 +80,37 @@ def remove_extra_feature_columns(
...
+
+    # Remove first column and add to the start
+    columns.remove(first_col_name)
+    col_order = [first_col_name] + columns
```
bremme, line 106: See above, if you choose `col` as an abbreviation for `columns` this is fine. But if you go for the more descriptive naming this should be `column_order`. 

I guess `sorted_columns` would make more sense here btw.

openstf/feature_engineering/general.py
```diff
@@ -76,3 +80,37 @@ def remove_extra_feature_columns(
         )
 
     return input_data.drop(extra_features, axis=1)
+
+
+def enforce_feature_order(df):
+    """Enforces correct order of features.
+
+    Alphabetically orders the feature columns. The load column remains the first column
```
bremme, line 88: This is not completely inline with what the function does. I also think the general term used in Python and pandas is sort/sorting instead of orders (not even sure that is the right English term, google says orders is bestellingen (Dutch))

The missing information is:
* that you assume the first column contains the load
* And that the Horizon column (if present) is moved to the back in any case (also if it was somewhere else)

openstf/metrics/reporter.py
```diff
@@ -74,7 +74,7 @@ def _make_data_series_figures(self, model: RegressorMixin) -> None:
 
         # Make model predictions
         for data_set in self.input_data_list:
-            model_forecast = model.predict(data_set.iloc[:, 1:])
+            model_forecast = model.predict(data_set.iloc[:, 1:-1])
```
bremme, line 77: This need some explanation? Why is the last column ignored?

openstf/model/objective.py
bremme: I agree, this looks like black magic to me. We could probably think about this a bit more to have a better understanding on how and when the structure of our data changes and or is required in a different form. But to keep it simple for now we should (like Frank said) probably check if the column is there and remove it and or set the slice depending on the outcome. (assuming both are valid use cases; with and without a Horizons column)

openstf/pipeline/train_model.py
```diff
@@ -135,15 +135,18 @@ def train_model_pipeline_core(
 
     # Split data
     train_data, validation_data, test_data = split_data_train_validation_test(
-        data_with_features.sort_index(axis=1)
+        data_with_features.sort_index(axis=0)
```
bremme, line 138: We seem to still quite a few of these `sort_index` calls. This raises similar question as with the `load` and `horizon` columns. Why do we need to do this, when is this required, why is the data not in the right format in the first place, is it really the callers responsibility. It seems a bit random and messy right now?

openstf/pipeline/train_predict_backtest.py
```diff
@@ -0,0 +1,135 @@
...
+def train_model_and_forecast_back_test(
+    pj: dict,
+    input_data: pd.DataFrame,
+    trained_models_folder: Union[str, Path],
+    save_figures_folder: Union[str, Path],
+    training_horizons: List[float] = None,
+) -> pd.DataFrame:
+    """Pipeline for a back test.
+
+        DO NOT USE THIS PIPELINE FOR OPERATIONAL FORECASTS
```
bremme, line 35: Prehaps this is such a general term `backtest`? But maye it would still be useful to explain what you mean by a `backtest` here.

openstf/pipeline/train_predict_backtest.py
```diff
@@ -0,0 +1,135 @@
...
+from openstf.model.serializer import PersistentStorageSerializer
+from openstf.model_selection.model_selection import split_data_train_validation_test
+from openstf.validation import validation
+from openstf.feature_engineering.feature_applicator import TrainFeatureApplicator
+
+DEFAULT_TRAIN_HORIZONS: List[float] = [0.25, 24.0]
+DEFAULT_EARLY_STOPPING_ROUNDS: int = 10
+
+
+def train_model_and_forecast_back_test(
```
bremme, line 26: I have not looked at the implementation yet. But if this pipeline is a combination of the `create_forecast` and `train_model` pipeline for the backtest usecase I would suggest to call it `train_model_and_create_forecast_backtest` (one wordt backtest is also consistent with the module name)

openstf/pipeline/train_predict_backtest.py
```diff
@@ -0,0 +1,135 @@
...
+    training_horizons: List[float] = None,
+) -> pd.DataFrame:
+    """Pipeline for a back test.
+
+        DO NOT USE THIS PIPELINE FOR OPERATIONAL FORECASTS
+
+    Args:
+        pj (dict): Prediction job.
+        input_data (pd.DataFrame): Input data
+        trained_models_folder:
```
bremme, line 40: Duplicated line.

openstf/pipeline/train_predict_backtest.py
```diff
@@ -0,0 +1,135 @@
...
+)
+from openstf.model.standard_deviation_generator import StandardDeviationGenerator
+from openstf.model.model_creator import ModelCreator
+from openstf.metrics.reporter import Reporter
+from openstf.model.serializer import PersistentStorageSerializer
+from openstf.model_selection.model_selection import split_data_train_validation_test
+from openstf.validation import validation
+from openstf.feature_engineering.feature_applicator import TrainFeatureApplicator
+
+DEFAULT_TRAIN_HORIZONS: List[float] = [0.25, 24.0]
```
bremme, line 22: If we use this in multiple modules would it make sense to have constant/config in a single file?

openstf/pipeline/train_predict_backtest.py
```diff
@@ -0,0 +1,135 @@
+# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
```
bremme, line 1: Perhaps `predict` is also fine. But I think we should aim for consistency, up until now we called this processing `create_forecast`. I would suggest either use `train_model_create_forecast_backtest.py` or perhaps just `backtest.py` if we want to use a shorter name.

openstf/pipeline/train_predict_backtest.py
```diff
@@ -0,0 +1,135 @@
...
+        trained_models_folder (Path): Path where trained models are stored
+        save_figures_folder (Path): path were reports (mostly figures) about the training procces are stored
+        training_horizons (list): Horizons to train on in hours.
+            These horizons are also used to make predictions (one for every horizon)
+
+    Returns:
+        forecast (pandas.DataFrame)
+
+    """
+    if training_horizons is None:
```
bremme, line 50: This part (line 50 to 102) looks identical to the train_model_core() function line 117:201. Could we not just use that? Or did I miss something perhaps?

openstf/pipeline/train_predict_backtest.py
bremme: I was thinking a bit of the same. It seems a bit inflexible to always require to store stuff to disk by default. Could/should we not go for a bit more flexibility and less redundancy/duplication by having a structure like this:

* backtest pipeline
  * `train_model_pipeline()` (reused/calls `train_model.train_model_pipeline_core`)
  * `create_forecast_pipeline()` (reused/calls `create_forecast.create_forecast_pipeline_core`)
  * `train_forecast_pipeline()` (calls the above pipelines)

test/unit/feature_engineering/test_feature_applicator.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def test_operational_feature_applicator_correct_order(self):
+        # Test for expected column order of the output
+        # Also check "Horizons" is not in the output
+        input_data = TestData.load("input_data.pickle")
+        data_with_features = OperationalPredictFeatureApplicator(
+            horizons=[0.25]
+        ).add_features(input_data[["load"]])
+        self.assertEqual(data_with_features.columns.to_list()[0], "load")
+        self.assertTrue("Horizon" not in data_with_features.columns.to_list())
```
bremme, line 31: I probably have to look into this myself. But how come that when you ask for `horizons=[0.25, 24.0]` you get a "Horizion" column but when you ask for `horizons=[0.25]` you don't?

test/unit/feature_engineering/test_feature_applicator.py
```diff
@@ -0,0 +1,31 @@
...
+from unittest import TestCase
+
+from openstf.feature_engineering.feature_applicator import (
+    TrainFeatureApplicator,
+    OperationalPredictFeatureApplicator,
+)
+from test.utils import TestData
+
+
+class TestFeatureApplicator(TestCase):
```
bremme, line 13: Nice unit tests!

test/unit/feature_engineering/test_general.py
```diff
@@ -0,0 +1,27 @@
...
+class TestGeneral(TestCase):
+    def test_enforce_feature_order_with_horizon_columns(self):
+
+        df = pd.DataFrame(
+            np.arange(12).reshape(3, 4), columns=["load", "Horizon", "A", "E"]
+        )
+
+        result = enforce_feature_order(df)
+
+        self.assertEqual(result.columns.to_list(), ["load", "A", "E", "Horizon"])
```
bremme, line 21: Minor suggestion. if you use a variable for the columns you could test against a sorted version. This makes it even more clear that you expect the columns to be sorted:
```python
columns = ["load", "Horizon", "A", "E"]
self.assertEqual(result.columns.to_list(), sorted(columns))
```

test/unit/pipeline/test_train_predict_backtest.py
```diff
@@ -0,0 +1,42 @@
...
+
+        forecast = train_model_and_forecast_back_test(
+            pj=self.pj,
+            input_data=self.train_input,
+            trained_models_folder="TEST",
+            save_figures_folder="OTHER_TEST",
+            training_horizons=[0.25, 24.0],
+        )
+
+        self.assertTrue("forecast" in forecast.columns.to_list())
```
bremme, line 28: I don't think you need the `to_list()` calls -> `"forecast" in forecast.columns` should be enough for example.

test/unit/pipeline/test_train_predict_backtest.py
```diff
@@ -0,0 +1,42 @@
...
+        self.assertTrue("realised" in forecast.columns.to_list())
+        self.assertTrue("Horizon" in forecast.columns.to_list())
+        self.assertEqual(list(forecast.Horizon.unique()), [24.0, 0.25])
```
bremme, line 31: minor remark. I personally like to compare to constants. if you change the training_horizons you have to update two things.

test/unit/pipeline/test_train_predict_backtest.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_train_model_pipeline_core_exception_incomplete_data(self):
```
bremme, line 33: now you are actually also unit testing `is_data_sufficient` this should already be done. I guess right here you should just mock patch it:

```python
    @patch("openstf.pipeline.train_predict_backtest.validation")
    def test_train_model_pipeline_core_exception_incomplete_data(self, validation_mock):
        """Test if Valueerror is raised when too little data is passed into the pieline."""
        validation_mock.is_data_sufficient.return_value = False
```

bremme, line 33: You really need to mock everything which in not part of the unit under test. Right now it even saves files to my local machine! That should certainly not happen. If you really want to test saving files (which is not part of this unit test) you should always use a tmpdir which is deleted after use.

bremme (changes requested):
I've added some comment to the code. My major concerns are:

1. Do we not like a bit more flexibility in de backtest?
2. Can we not reused some of the core pipelines (perhaps with small modifications)
3. Can we not come up with a more robust strategy for the `load` and `Horizon` columns (perhaps a different story).
4. Nice that you added some unittest! But right now the unit test are testing too many other units and create files on my local machine. In Principe everything not part of the unit under test should be mocked, especially disk access.

test/unit/pipeline/test_train_predict_backtest.py
bremme: Maybe not required. But most things look really good on these tests, just a couple important things not so much.

openstf/feature_engineering/general.py
bremme: I noticed that we use both "Horizon" and "horizon". I think we should change all column names `lower_snake_case` when possible, right now we mixing many conventions.

openstf/pipeline/create_forecast.py
JanMaartenvanDoorn: yes good point, I now moved the dropping of the load column to the next step were "forecast_input_data" is created.

openstf/pipeline/train_predict_backtest.py
JanMaartenvanDoorn: True, I removed the saving of the model and the reporting in the pipeline. I made the pipeline also return the model such that the user can save it him/herself.

openstf/pipeline/train_predict_backtest.py
JanMaartenvanDoorn: I have added the evaluation of the backtest to the notebook using the openstf.metrics metrics

openstf/feature_engineering/general.py
JanMaartenvanDoorn: The function basically removes all columns that are not in features, "load" is not a feature but it is essential for what we want to predict. I could slice (.iloc[:,1:]) the input of this function such that load is not passed at all. Is that (combined with the order check earlier) a better solution?

openstf/feature_engineering/general.py
JanMaartenvanDoorn: Yes but the first column ("load") and last column ('Horizon') are excluded from the sorting and therefore the end result is not a sorted dataframe. I think therefore the enforce order is more descriptive: we want to enforce this order: ["load"] -- [alphabetically sorted features] -- ['horizon']

openstf/feature_engineering/general.py
JanMaartenvanDoorn: made this more consistent

openstf/feature_engineering/general.py
JanMaartenvanDoorn: Added this information

openstf/metrics/reporter.py
JanMaartenvanDoorn: This column contains the horizon information and is not expected by the model as input, I have added a comment to clarify

openstf/feature_engineering/general.py
bremme: I think I did not understand the function properly. I really had to read it a couple times to see what was happening. Perhaps we can improve on this to make it more clear. I'm thinking what makes it confusion? I couple things come to mind:

* I'm not sure "extra" is the right English term in this context. I think "additional" is better perhaps. So we could use `remove_additional_feature_columns` for example? Or maybe any of the other terms you already use in the function: `remove_not_requested_features`, `remove_unrequested_features`
* Perhaps the above combined with a better name of the input arguments could make it more clear:
  * `remove_unrequested_features(input_data, requested_features)` is a lot more clear imo compared to the current situation.

Coming back to your answer. We run into this "issue" on multiple occasion. We have all features our load and horizon in the same datastructure and we don't know if these exist and what their location is. Ideally we would know all these things are these things would be part of a more strictly defined datastructure.

But, until that is the case I don't think the current implementation is a bad idea. You should probably also check on the "Horizon" column besides the "load" column.

JanMaartenvanDoorn: Good naming sugestions, I implemented them, good to make a separate issue on the datamodels

openstf/feature_engineering/general.py
JanMaartenvanDoorn: Good one I changed it to snake_case

openstf/model/objective.py
JanMaartenvanDoorn: Check added

openstf/pipeline/train_predict_backtest.py
JanMaartenvanDoorn: Good point

test/unit/pipeline/test_train_predict_backtest.py
JanMaartenvanDoorn: Good one, changed this

JanMaartenvanDoorn: Removed the saving and loading model as a whole, this also solves this problem

test/unit/pipeline/test_train_predict_backtest.py
JanMaartenvanDoorn: yes, removed the calls

openstf/feature_engineering/general.py
bremme: `sort_features_columns` also tells you that the function only sorts the features columns. In my opinion this implies that it does not sort non feature columns like the "horizon" and "load" column. But its not a really big deal.

openstf/pipeline/train_predict_backtest.py
JanMaartenvanDoorn: Yes, now made a train_pipeline_common function

PR #108, opened 2021-06-23
add fix and unit test

FrankKr:
As discussed by phone.
There was a bug when calculating the completeness of T-7d. This would result in a negative completeness for this column, falsely resulting in a very low completeness overall.

I added a small unit test for this and fixed this behavior in the code.
I plan to also merge this hotfix to the sklearn branch after approvement

### DIFF
--- a/test/unit/validation/test_validation_calc_completeness.py
+++ b/test/unit/validation/test_validation_calc_completeness.py
@@ -123,7 +123,7 @@ class CalcCompletenessTest(BaseTestCase):
             },
         )
         completeness = calc_completeness(df, time_delayed=True)
-        self.assertEqual(completeness, 11/12.)
+        self.assertEqual(completeness, 11 / 12.0)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


test/unit/validation/test_validation_calc_completeness.py
```diff
@@ -109,6 +109,22 @@ def test_empty_dataframe(self):
...
+            index=[0, 1, 3],
+            data={
+                "T-15min": [1, np.nan, np.nan],
+                "T-7d": [2, 3, 4],
+                "T-24d": [4, 5, 6],
+                "col1": [1, np.nan, 2],
+            },
+        )
+        completeness = calc_completeness(df, time_delayed=True)
+        self.assertEqual(completeness, 11 / 12.0)
```
JanMaartenvanDoorn, line 126: Maybe add a comment what 11/12.0 means. Sugestion: "(total_number_of_datapoints-unexpected_nans)/total_number_of_datapoints".
Also good to make both numbers either a float of an int

JanMaartenvanDoorn (approved):
Looks good! Minor comment about adding an extra bit of clarification.

test/unit/validation/test_validation_calc_completeness.py
FrankKr: Good idea

PR #110, opened 2021-06-30
Features ktps 1368 implement minimal sklearn

JanMaartenvanDoorn:


bremme:
I noticed that there where a couple code smells. I fixed most of them except the FIXME ones. You could make stories/issues and remove them if you want.

I also bumped the version to 2.0.0 is that what you intended as well?

bremme (approved):

PR #112, opened 2021-07-07
Hotfix fix python 37 38 compatibility

bremme:
This PR will fix Python 3.7 and Python 3.8 incompatibilites with `list[float], tuple[int]` etc. For now we are just going to use the `from typing import List` type hints.


JanMaartenvanDoorn (approved):
Looks good!

PR #113, opened 2021-07-08
Bugfix create model wrong signature objective function

bremme:
There was a small bug where some parts of the code expected a different interface to the `create_model()` method. This PR fixes this by making a couple small adjustments.


openstf/model/model_creator.py
```diff
@@ -33,12 +34,11 @@ def create_model(pj: dict) -> RegressorMixin:
             RegressorMixin: model
         """
         # This will raise a ValueError when an invalid model_type str is used
-        model_type = MLModelType(pj["model"])
+        model_type = MLModelType(model_type)
 
-        kwargs = {}
+        model_kwargs = {}
+        # only pass relevant argument to model constructor to prevent warnings
+        if "quantiles" in kwargs and model_type is MLModelType.XGB_QUANTILE:
```
JanMaartenvanDoorn, line 41: Maybe add a RELEVANT_KWARGS config constant at the start of the file where we can add more of these kwargs in the future

JanMaartenvanDoorn (approved):
Looks good, one remark to consider

openstf/model/model_creator.py
bremme: Okay, I can do that

PR #114, opened 2021-07-13
Bugfix always sort index after applying horizons

bremme:
Changelog

* always sort the index after adding feature for multiple horizons
* use the forecast tzinfo format when add the tAhead column
* Use pytz.utc for the create_basecase_forecast pipeline
* Add documentation on the expected/required keys in prediction jobs for multiple functions/methods
* Add option to set model_type when calling main() of
    * create_forecast task
    * train_model task


JanMaartenvanDoorn (approved):
Looks good, nice changes!

PR #115, opened 2021-07-14
Bugfix check model age before getting inputs (KTPS-1479)

FrankKr:
fixes KTPS-1479

Move the check of the age of the most recent model to the point in the pipeline *before* loading all the input data.
To do this, I made a function 'get_model_age' in pipeline/train_model, which is called from task/train_model.
Please have a look at `openstf/tasks/train_model.py` for this implementation

This way, it does mean that the task/train_model uses a function which depends on the model_serializer.
However, since this is covered in a seperate function, if in the future we move to a different way of loading/storing models, this can be changed in a single location.

I also included two unit tests to check behavior in case there does and does not exist a recent model

### DIFF
--- a/openstf/model/serializer.py
+++ b/openstf/model/serializer.py
@@ -154,12 +154,15 @@ class PersistentStorageSerializer(AbstractSerializer):
 
         return loaded_model
 
-    def determine_model_age_from_pid(self, pid) -> float:
+    def determine_model_age_from_pid(self, pid: int) -> float:
         """Determine model age in days of most recent model for a given pid.
         If no previous model is found, float(Inf) is returned
 
+        Args:
+            pid: int
+
         Returns:
-            object: """
+            float: model age in days"""
         model_path = self.find_most_recent_model_path(pid)
         model_age_days = self._determine_model_age_from_path(model_path)
         return model_age_days
--- a/openstf/model/serializer.py
+++ b/openstf/model/serializer.py
@@ -113,8 +113,13 @@ class PersistentStorageSerializer(AbstractSerializer):
         elif model_id is not None:
             model_path = self.convert_model_id_into_model_path(model_id)
 
-        if model_path is None or model_path.is_file() is False:
-            msg = f"No model file found at save location: '{model_path}'"
+        if model_path is None:
+            msg = f"No (most recent) model found"
+            self.logger.error(msg)
+            raise FileNotFoundError(msg)
+
+        if model_path.is_file() is False:
+            msg = f"model_path is not a file ({model_path})"
             self.logger.error(msg)
             raise FileNotFoundError(msg)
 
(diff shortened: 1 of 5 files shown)


openstf/model/serializer.py
```diff
@@ -113,8 +118,13 @@ def load_model(
         elif model_id is not None:
             model_path = self.convert_model_id_into_model_path(model_id)
 
-        if model_path is None or model_path.is_file() is False:
-            msg = f"No model file found at save location: '{model_path}'"
+        if model_path is None:
```
JanMaartenvanDoorn, line 121: Nice to split this, makes it better readable

openstf/model/serializer.py
```diff
@@ -144,6 +154,16 @@ def load_model_from_path(self, model_path):
 
         return loaded_model
 
+    def determine_model_age_from_pid(self, pid) -> float:
```
JanMaartenvanDoorn, line 157: Great function, nicely implemented. One minor thing: could you add a typehint for the pid argument? (self, pid: int)
Also the argument is missing in the docstring.

openstf/pipeline/train_model.py
```diff
@@ -247,3 +247,17 @@ def train_pipeline_common(
         pj, validation_data
     ).generate_standard_deviation_data(model)
     return model, train_data, validation_data, test_data
+
+
+def get_model_age(trained_models_folder: str, pid: int) -> float:
```
JanMaartenvanDoorn, line 252: Do we still need this function? Can we not directly call the PersistentStorageSerializer(trained_models_folder).determine_model_age_from_pid(pid) instead?

test/unit/tasks/test_train_model.py
```diff
@@ -25,3 +26,31 @@ def test_create_train_model_task_happy_flow(self, train_model_pipeline_mock):
         self.assertEqual(
             train_model_pipeline_mock.call_args_list[0][0][0]["id"], self.pj["id"]
         )
+
+    @patch(
+        "openstf.tasks.train_model.get_model_age", return_value=MAXIMUM_MODEL_AGE - 1
+    )
+    @patch("openstf.tasks.train_model.train_model_pipeline")
```
JanMaartenvanDoorn, line 33: Good to remember making unittests!

JanMaartenvanDoorn (commented):
Looks good, a few minor suggestions to make the code more lean.

openstf/pipeline/train_model.py
FrankKr: You're right, I'll remove this superfluous function

JanMaartenvanDoorn (approved):

JanMaartenvanDoorn:
Approved, otherwise you cannot merge

PR #118, opened 2021-07-26
Bugfix ktps 1499 basecase stdev const load

FrankKr:
There was a bug where basecase_task gave exception if load was constant.
By solving that bug, other issues showed, which were also tackled in this PR.

Major:
- if basecase if stored, `algtype` should be basecase, but `type` should be pj['typ']
- basecase should not be stored for t < NOW+48hours. However, before this PR is wasn't possible to make a basecase forecast for those moments, while we do want that option for e.g. backtests. This responsibility is moved to the task-level
- improved `validate` step, so basecase can also be made when load is constant

Minor:
- there were some old and obsolete references to horizon = 'near' or 'far'. These have been updated to be in t_ahead_hours, reference is therefore removed.

Also added unit tests to test desired behavior of pipeline and task

### DIFF
--- a/openstf/pipeline/create_basecase_forecast.py
+++ b/openstf/pipeline/create_basecase_forecast.py
@@ -2,10 +2,10 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 from pathlib import Path
+from datetime import datetime, timedelta
 
 import pandas as pd
-from datetime import timedelta, tzinfo
-import pytz
+import numpy as np
 import structlog
 from openstf.enums import ForecastType
 from openstf.feature_engineering.feature_applicator import (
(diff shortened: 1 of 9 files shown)


openstf/pipeline/create_basecase_forecast.py
```diff
@@ -73,21 +56,21 @@ def create_basecase_forecast_pipeline(
         ],  # Generate features for load 7 days ago and load 14 days ago these are the same as the basecase forecast.
     ).add_features(validated_data)
 
-    # Select the basecase forecast interval
-    forecast_input_data = data_with_features[forecast_start:forecast_end]
+    # Only try to make a forecast for moments where load = nan
+    forecast_input = data_with_features.loc[data_with_features["load"].isna(), :]
```
JanMaartenvanDoorn, line 60: What happens if historical load values are also nan due to some error? Should we not enforce a forecast is always for the future?

JanMaartenvanDoorn (approved):

PR #122, opened 2021-07-27
Bugfix ktps 1497 hyperparam issue first times

FrankKr:
if no old hyperparams were available, this would lead to an exception. 
Now, it does not lead to an exception anymore.

Commits include reproducing the issue and subsequently solving it.
Moreover, no unit tests were available for the task, so I added those as well.

!!! Note, the post_teams dependency in optimize_hyperparameters_task caused issues in unit testing.
This dependency is undesirable in the first place, and also not required by KTP operationally.
I removed this dependency altogether in this PR. Please raise a point if you disagree with this.

### DIFF
--- a/openstf/tasks/optimize_hyperparameters.py
+++ b/openstf/tasks/optimize_hyperparameters.py
@@ -39,7 +39,12 @@ def optimize_hyperparameters_task(pj: dict, context: TaskContext) -> None:
 
     # Determine if we need to optimize hyperparams
     datetime_last_optimized = context.database.get_hyper_params_last_optimized(pj)
-    last_optimized_days = (datetime.utcnow() - datetime_last_optimized).days
+    try:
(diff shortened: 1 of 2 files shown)


openstf/tasks/optimize_hyperparameters.py
```diff
@@ -39,7 +38,12 @@ def optimize_hyperparameters_task(pj: dict, context: TaskContext) -> None:
 
     # Determine if we need to optimize hyperparams
     datetime_last_optimized = context.database.get_hyper_params_last_optimized(pj)
-    last_optimized_days = (datetime.utcnow() - datetime_last_optimized).days
+    try:
```
JanMaartenvanDoorn, line 41: Good solution

JanMaartenvanDoorn (approved):
Looks good, nice to add unittests

PR #124, opened 2021-07-30
Bugfix no confidence in case of flatliner

JanMaartenvanDoorn:
Fixed bug that gave error when making a forecast with partly unknown stdev info


FrankKr (approved):
Looks good!

PR #125, opened 2021-08-03
use modeltype enum

FrankKr:
small improvement; use central Enum to get available model types instead of hard coded lists everywhere


JanMaartenvanDoorn (approved):
Nice improvement!

PR #128, opened 2021-08-03
raise distinct exceptions for calc kpi

FrankKr:
Made reusable, custom exception. 

Considered also improving the teams logging message, to show all separate exceptions individually. Did not implement this, since we don't want to keep using teams as primary monitoring solution but something based on logs, e.g. Kibana. Exceptions are already individually logged.

### DIFF
--- a/openstf/exceptions.py
+++ b/openstf/exceptions.py
@@ -6,6 +6,8 @@
 These are not complete, but will be added on a case by case basis
 """
 
+from typing import Union
+
 # Define custom exception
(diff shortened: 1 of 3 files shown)


openstf/exceptions.py
```diff
@@ -0,0 +1,31 @@
+"""Define custom exception classes.
```
JanMaartenvanDoorn, line 1: I think you need to add some license info here in order to pass all the build checks

PR #129, opened 2021-08-03
add default training period days hyperparam

FrankKr:
fixes minor bug. This argument used to be a hyperparameter, but is not anymore.

image


JanMaartenvanDoorn (approved):
Looks good!

PR #131, opened 2021-08-05
Feature add draft releases action

FrankKr:
Automatically draft new release with info on updates.

See https://github.com/marketplace/actions/release-drafter for details.

Note that this can only be properly tested when the action is included in main


FrankKr:
The Release Drafter action fails, but don;t be alarmed. This is a known bug: https://github.com/release-drafter/release-drafter/issues/398 which is supposedly fixed after merging to main

PR #149, opened 2021-08-17
raise clear exception on unimplemented model

FrankKr:
Hyperparameter optimization for model_types other than xgb are not yet implemented.
I created a follow-up story to implement that functionality (https://alliander.atlassian.net/browse/KTPS-1543)

In this PR, I at least specified a clear NotImplementedException, and added a unit test for this exception.

### DIFF
--- a/openstf/model/objective_creator.py
+++ b/openstf/model/objective_creator.py
@@ -13,6 +13,14 @@ class ObjectiveCreator:
 
     @staticmethod
     def create_objective(model_type: Union[MLModelType, str]) -> RegressorObjective:
+        if model_type not in ObjectiveCreator.OBJECTIVES.keys():
+            raise NotImplementedError(
+                "This model_type is not implemented "
+                "for hyperparam optimization."
+                f"Received: {model_type}, "
+                "Should be in: "
+                f"{list(ObjectiveCreator.OBJECTIVES.keys())}"
+            )
         model_type = MLModelType(model_type)
 
         return ObjectiveCreator.OBJECTIVES[model_type]
--- a/openstf/model/objective_creator.py
+++ b/openstf/model/objective_creator.py
@@ -13,13 +13,11 @@ class ObjectiveCreator:
 
     @staticmethod
     def create_objective(model_type: Union[MLModelType, str]) -> RegressorObjective:
-        if model_type not in ObjectiveCreator.OBJECTIVES.keys():
+        valid_types = list(ObjectiveCreator.OBJECTIVES.keys())
+        if model_type not in valid_types:
             raise NotImplementedError(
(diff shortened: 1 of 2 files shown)


openstf/model/objective_creator.py
```diff
@@ -13,6 +13,14 @@ class ObjectiveCreator:
 
     @staticmethod
     def create_objective(model_type: Union[MLModelType, str]) -> RegressorObjective:
+        if model_type not in ObjectiveCreator.OBJECTIVES.keys():
+            raise NotImplementedError(
+                "This model_type is not implemented "
+                "for hyperparam optimization."
+                f"Received: {model_type}, "
+                "Should be in: "
+                f"{list(ObjectiveCreator.OBJECTIVES.keys())}"
```
bremme, line 22: Perhaps to make the message a little shorter and more readable you could use:

```python
        valid_types = list(ObjectiveCreator.OBJECTIVES.keys())
        if model_type not in valid_types:
            raise NotImplementedError(
                "This model_type is not implemented "
                "for hyperparam optimization."
                f"Received: {model_type}, "
                "Should be in: "
                f"{', '.join([t.value for t in valid_types])}"
            )
```

This gives `"xgb"`, while the current implementation gives `[<MLModelType.XGB: 'xgb'>]`.

Minor remark. This module does not need knowledge about hyperparameters optimization. It just provides an objective function. So you could be more precise with the error message on what's going on; the objective function for this model type is not implemented yet: `f"No objective function for {model_type} valid model_types are: {', '.join([t.value for t in valid_types])}"`

test/unit/model/test_objective_creator.py
```diff
@@ -7,14 +7,19 @@
...
+from openstf.model.objective import XGBRegressorObjective, RegressorObjective
 
 
 class TestObjectiveCreator(BaseTestCase):
-    def test_create_objective(self):
+    def test_create_objective_happy(self):
         Objective = ObjectiveCreator.create_objective(MLModelType.XGB)
         self.assertEqual(Objective, XGBRegressorObjective)
 
+    def test_create_objective_all_model_types(self):
```
bremme, line 18: I would suggest to call this test: `test_create_objective_not_implemented_model_type` or something along that line.

bremme:
Looks fine!

bremme (approved):

openstf/model/objective_creator.py
FrankKr: Thanks for your suggestion, I'll incorporate it

test/unit/model/test_objective_creator.py
FrankKr: good idea

PR #150, opened 2021-08-18
if multiple exceptions are raised in a task, present an overview of pids per exception in teams message

FrankKr:
added this functionality to the predictionloop + taskcontext, and added a unit test to test this.

### DIFF
--- a/openstf/tasks/utils/predictionjobloop.py
+++ b/openstf/tasks/utils/predictionjobloop.py
@@ -4,6 +4,7 @@
 
 import random
 import sys
+from collections import defaultdict
 
 
 class PredictionJobLoop:
@@ -106,6 +107,7 @@ class PredictionJobLoop:
         """
         pids_successful = []
         pids_unsuccessful = []
+        pids_unsuccessful_dict = defaultdict(list)
         last_job_exception = None
 
         num_jobs = len(self.prediction_jobs)
@@ -138,6 +140,7 @@ class PredictionJobLoop:
 
             except Exception as exception:
                 pids_unsuccessful.append(prediction_job["id"])
+                pids_unsuccessful_dict[str(exception)].append(prediction_job["id"])
                 last_job_exception = exception
 
                 self._handle_exception_during_iteration(prediction_job, exception)
@@ -168,6 +171,7 @@ class PredictionJobLoop:
                 "num_jobs": num_jobs,
                 "pids_successful": pids_successful,
                 "pids_unsuccessful": pids_unsuccessful,
+                "exceptions": pids_unsuccessful_dict,
                 "jobs_successful": jobs_successful,
                 "jobs_unsuccessful": jobs_unsuccessful,
                 "jobs_started": jobs_started,
--- a/openstf/tasks/utils/taskcontext.py
+++ b/openstf/tasks/utils/taskcontext.py
@@ -122,6 +122,17 @@ class TaskContext:
                     ]
                 }
             )
+            # Add details of exceptions and pids
+            string_pids_exceptions = "\n".join(
+                [f"{key}:{value}\n" for key, value in metrics["exceptions"].items()]
+            )
+            msg["sections"].append(
+                {
+                    "facts": [
+                        ("Exceptions: pid(s)", string_pids_exceptions),
+                    ]
+                }
+            )
 
         msg["sections"].append(
             {
--- a/test/unit/tasks/utils/test_task_context.py
+++ b/test/unit/tasks/utils/test_task_context.py
@@ -7,7 +7,10 @@ import unittest
 from test.utils import TestData
 from unittest.mock import MagicMock, Mock, patch
 
-from openstf.tasks.utils.predictionjobloop import PredictionJobLoop
+from openstf.tasks.utils.predictionjobloop import (
+    PredictionJobLoop,
+    PredictionJobException,
+)
 
 # import project modules
 from openstf.tasks.utils.taskcontext import TaskContext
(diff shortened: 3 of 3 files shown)


openstf/tasks/utils/predictionjobloop.py
```diff
@@ -106,6 +107,7 @@ def map(self, function, *args, **kwargs):
         """
         pids_successful = []
         pids_unsuccessful = []
+        pids_unsuccessful_dict = defaultdict(list)
```
bremme, line 110: Perhaps we can keep it simpler by only using one data structure for unsuccessful pid's. Now both contain the pid's. I would suggest to use `pid_unsucessful` or `pid_exceptions` as a normal dicts:

```python
pids_unsucessful = {}
```

Or perhaps I'm missing why a defaultdict is required here?

openstf/tasks/utils/predictionjobloop.py
```diff
@@ -138,6 +140,7 @@ def map(self, function, *args, **kwargs):
 
             except Exception as exception:
                 pids_unsuccessful.append(prediction_job["id"])
```
bremme, line 142: I would suggest to not strip information here and use a single datastructure:

```python
# We can always remove information later if we want, but for now keep the full exception I guess (minor)
pids_unsucessful[prediction_job["id"]] = exception
```

openstf/tasks/utils/predictionjobloop.py
```diff
@@ -168,6 +171,7 @@ def map(self, function, *args, **kwargs):
                 "num_jobs": num_jobs,
                 "pids_successful": pids_successful,
                 "pids_unsuccessful": pids_unsuccessful,
+                "exceptions": pids_unsuccessful_dict,
```
bremme, line 174: This can be remove/simplified if we use  a single data structure.

openstf/tasks/utils/taskcontext.py
```diff
@@ -122,6 +122,17 @@ def _send_teams_message(self, exc_type, exc_info, stack_info):
...
                 }
             )
+            # Add details of exceptions and pids
+            string_pids_exceptions = "\n".join(
+                [f"{key}:{value}\n" for key, value in metrics["exceptions"].items()]
+            )
+            msg["sections"].append(
+                {
+                    "facts": [
+                        ("Exceptions: pid(s)", string_pids_exceptions),
```
bremme, line 132: Perhaps easier to read to add 1-2 newlines before the <pid>:<exception string> messages.  Not sure how it looks like this, perhaps making separate facts makes more sense or an unordered list.

openstf/tasks/utils/predictionjobloop.py
FrankKr: defaultdict(list) is very practical, as it allows easy 'append to a list if a key is already present, or create the list for that key if the key is not yet present'-functionality.

openstf/tasks/utils/predictionjobloop.py
bremme: Now I see what you try to achieve grouping the exceptions. I can see some benefits to doing that. Still I would recommend to use a single data structure. Perhaps `pids_exceptions` is more clean in this case.

FrankKr: I agree that the list of pids_unsecessful is a bit superfuous af we move to a dict with {exceptions : pid-list}. I think your code results in something different from both those options, right?

openstf/tasks/utils/taskcontext.py
```diff
@@ -122,6 +122,17 @@ def _send_teams_message(self, exc_type, exc_info, stack_info):
                     ]
                 }
             )
+            # Add details of exceptions and pids
+            string_pids_exceptions = "\n".join(
+                [f"{key}:{value}\n" for key, value in metrics["exceptions"].items()]
```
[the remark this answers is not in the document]
bremme: In the updated case, you would get the pids of the ussuccesful jobs something like this I guess.

```python
            string_pids_unsuccessful = ", ".join(
                [f"{', '.join(pids)}" for pids in metrics["pid_exceptions"].values()]
            )
```

openstf/tasks/utils/predictionjobloop.py
FrankKr: yes

openstf/tasks/utils/taskcontext.py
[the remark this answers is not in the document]
FrankKr: your suggestion changes the implementation from
`exception1: pid1, pid2, pid3
exception2: pid4, pid5`
to something else, which would also be an option, but would give a less clear overview of the separate exceptions that occured.

openstf/tasks/utils/taskcontext.py
FrankKr: If all works well, adding it as a separate 'facts' portion, already results in some nice line breaks;
image

bremme:
Nice improvement. I had/have a little trouble to figure out the strategy and especially how the Teams message is gonna look. After reading the code the strategy of grouping the exception using the exception message as a key seems to make sense. So you can partly ignore my first comments. Perhaps a screenshot of the formatted message would help a lot to see how readable the message actually is.

openstf/tasks/utils/predictionjobloop.py
bremme: I think we mean the same. Probably a dict with exception messages as keys and list of pid as values is the way to go.

bremme (approved):

openstf/tasks/utils/taskcontext.py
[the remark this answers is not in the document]
bremme: You are right, it was just a quick suggestion. It should be

```shell
exception message one: pid1, pid2, pid3
exception message two: pid4, pid5, pid6
```

PR #151, opened 2021-08-18
Feature better exceptions train model pipeline

bremme:
This PR adds better exceptions to the train model pipeline functions.

**OLD exceptions**

* `ValueError` When input data is insufficient
* `RuntimeError`: When old model is better than new model
* `RuntimeError`: When column order is incorrect

**NEW expceptions**
* `InputDataInsufficientError`: when input data is insufficient.
* `InputDataWrongColumnOrderError`: when input data has a invalid column order.
* `OldModelHigherScoreError`: When old model is better than new model.

Especially the identical RuntimeError for two different reasons has been problematic when client codes calls the train model pipeline. (This PR is required to solve some issues in our openstf api code)

> NOTE: In order to use these improvements staighaway I want to make a v2.2.2 release after merging this PR.

### DIFF
--- a/openstf/exceptions.py
+++ b/openstf/exceptions.py
@@ -10,7 +10,7 @@ from typing import Union
 
 
 # Define custom exception
-class NoPredictionException(Exception):
+class NoPredictionError(Exception):
     """Custom exception if no historic predictions are available
     Attributes:
         pid -- prediction job id for which the exception occurred
@@ -27,7 +27,7 @@ class NoPredictionException(Exception):
         super().__init__(self.message)
 
 
-class NoLoadException(Exception):
+class NoLoadError(Exception):
     """Custom Exception if no historic load is available
     Attributes:
         pid -- prediction job id for which the exception occurred
--- a/openstf/exceptions.py
+++ b/openstf/exceptions.py
@@ -2,46 +2,17 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-"""Define custom exception classes.
-These are not complete, but will be added on a case by case basis
+"""openstf custom exceptions
 """
-
-from typing import Union
 
 
 # Define custom exception
-class NoPredictionError(Exception):
-    """Custom exception if no historic predictions are available
-    Attributes:
-        pid -- prediction job id for which the exception occurred
-        message -- explanation of the error
-    """
-
-    def __init__(
-        self,
-        pid: Union[int, str],
-        message: str = "No historic predictions found." "Could not calc KPI",
-    ):
-        self.pid = pid
-        self.message = message
-        super().__init__(self.message)
+class NoPredictedLoadError(Exception):
+    """No predicted load for given datatime range"""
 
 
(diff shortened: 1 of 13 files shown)


openstf/exceptions.py
```diff
@@ -41,3 +42,19 @@ def __init__(
...
+class InputDataInsufficientError(InputDataInvalidError):
+    """Insufficient input data"""
+
+
+class InputDataWrongColumnOrderError(InputDataInvalidError):
+    """Wrong column order input data"""
+
+
+class OldModelHigherScoreError(Exception):
+    """Old model has a higher score then new model"""
```
FrankKr, line 60: Suggestion, can we include a `pid` argument with this exception definition? That would make it more easy to see for which pid the exception was raised.

setup.py
```diff
@@ -28,7 +29,7 @@ def read_long_description_from_readme():
 
 setup(
     name="openstf",
-    version="2.2.1",
+    version="2.2.2",
```
FrankKr, line 32: Good idea to make a new release, but I'm wondering if including the increase in version number in the PR is the workflow we desire for this. 
However, this is very practical:)

FrankKr (approved):
Looks good! One minor suggestion

openstf/exceptions.py
bremme: Good point, I agree that you want to know for which `pid` a certain exception is raised. I'm not sure if this knowledge belongs to the exception itself. I would suggest to add it to the logging (if not already done so) or to the exception message when we raise the exception.

update, I just checked. In case of the `train_model_pipeline()` and `train_model_pipeline_core()` function we use the prediction job name in the exception: 
image
image

We could change that to the prediction job id, perhaps that makes more sense?

setup.py
bremme: What would you suggest? Or what would be an alternative workflow? Merging to master first and then incrementing the version number? The disadvantage of that strategy would be to have commits which are not releases on master. I would prefer that every commit is a new version on master. Or did you have an other alternative strategy in mind, happy to learn:)

openstf/exceptions.py
FrankKr: Putting it in the logging is also a valid option. Is remembered seeing the inclusing in the exception itself somewhere else in openstf: 
https://github.com/alliander-opensource/openstf/blob/main/openstf/exceptions.py

I guess either the pj['name']  or id would work, but perhaps nice to pick 1. Feel free to pick;)

setup.py
FrankKr: Alternative would be the workflow you describe. That's more or less what typically happens now.

bremme: I believe (not totally sure) that incrementing the version before merging is more common in both in Gitflow or Githubflow (most images indicate a version number for every commit/merge on master). After the merge a new release is ready in that case. I would suggest we do just that. But its not a very big deal if someone forgets and bumps the version on main. (You could also always just do it on the already merged feature branch and merge it again ofc).

Issue #155, opened 2021-08-31
Bug: model_type should be string


FrankKr:
Context: 

In the hyperparameter task, a 'objective' is created based on the model_type:
https://github.com/alliander-opensource/openstf/blob/841579dbded3b00e692c8161300e4570156ddb8f/openstf/pipeline/optimize_hyperparameters.py#L66
In the task, this is one of ['xgb','xgb_quantile',...], for which currently only 'xgb' is implemented.
In the unit test however, a Enum-key is supplied. This causes a NotImplementedException in operation (https://alliander.atlassian.net/browse/KTPS-1572)

Unit test to modify (`model_type='xgb'`):
https://github.com/alliander-opensource/openstf/blob/841579dbded3b00e692c8161300e4570156ddb8f/test/unit/model/test_objective_creator.py#L15

Place in the code to fix the bug:
https://github.com/alliander-opensource/openstf/blob/841579dbded3b00e692c8161300e4570156ddb8f/openstf/model/objective_creator.py#L16

bremme:
I was wondering why do we choose to change the interface instead of just fixing the bug. I thought it was quite a nice feature to use `model_type: Union[MLModeltype, str]` as input argument. Furthermore I would suggest to keep using the same interface on `create_model` and `create_objective`. My suggestion would be to just fixes the interface:

```python
class ObjectiveCreator:
    OBJECTIVES = {MLModelType.XGB: XGBRegressorObjective}

    @staticmethod
    def create_objective(model_type: Union[MLModelType, str]) -> RegressorObjective:
        """Create an objective function based on model type.

        Args:
            model_type (Union[MLModelType, str]): Model type to construct.

        Raises:
            NotImplementedError: When using an invalid model_type string.

        Returns:
            RegressorObjective: Objective function
        """
        try:
            # This will raise a ValueError when an invalid model_type str is used
            # and nothing when a MLModelType enum is used.
            model_type = MLModelType(model_type)
        except ValueError as e:
            valid_types = [t.value for t in MLModelType]
            raise NotImplementedError(
                f"No objective for '{model_type}', "
                f"valid model_type's are: {valid_types}"
            ) from e

        return ObjectiveCreator.OBJECTIVES[model_type]
```

```python
class ModelCreator:
    """Factory object for creating machine learning models"""

    # Set object mapping
    MODEL_CONSTRUCTORS = {
        MLModelType.XGB: XGBOpenstfRegressor,
[...]

PR #156, opened 2021-09-01
Bugfix #155 ktps 1572 enum hyperparam optim

EnricoASchmitz:
Small fix for the object creation and the corresponding test


FrankKr (approved):
Looks good!

PR #158, opened 2021-09-02
Feature ktps 1548 error formatting

EnricoASchmitz:
The formatting of the error handeling is changed to allow for easy listing of PID's encountering the same error.
Slightly adapted  the unit test for the changes.
@FrankKr please take a look at the exception classes, it now contains a print statement. Could be useful information but might be beter to write to a logfile?


FrankKr (commented):
Looks good! If you could remove the print() statement, that would be perfect. 
Also nice that you included typehints!

FrankKr (approved):

PR #159, opened 2021-09-03
Bugfix #155 KTPS-1572 enum hyperparam optim

bremme:
Small PR to suggest some improvements to the #155 bugfix.

* Use the same interface for `create_model` and `create_objective`
* Fix type hint `model_type: Union[MLModelType, str]` was not respected
* Simplify unit test

### DIFF
--- a/test/unit/model/test_model_creator.py
+++ b/test/unit/model/test_model_creator.py
@@ -4,46 +4,34 @@
 
 from unittest import TestCase
 
-from xgboost import XGBRegressor
-
-from openstf.model.regressors.xgb_quantile import XGBQuantileOpenstfRegressor
-from openstf.model.regressors.regressor_interface import OpenstfRegressorInterface
-from openstf.model.model_creator import ModelCreator
 from openstf.enums import MLModelType
+from openstf.model.model_creator import ModelCreator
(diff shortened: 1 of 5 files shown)


EnricoASchmitz (approved):
Looks good!

test/unit/model/test_model_creator.py
```diff
@@ -4,46 +4,34 @@
...
+        # Test happy flow (both str and enum model_type arguments)
+        valid_types = [t.value for t in MLModelType] + [t for t in MLModelType]
+        for model_type in valid_types:
+            model = ModelCreator.create_model(model_type)
+            self.assertIsInstance(model, OpenstfRegressorInterface)
 
     def test_create_model_quantile_model(self):
         # Test if quantile model is properly returned
-        model_type = "xgb_quantile"
+        model_type = MLModelType.XGB_QUANTILE
```
FrankKr, line 23: @bremme Please be advised that this change in the unittests breaks the production code (!!).
In the pipeline, the str representation is given to the function and NOT the enum.key.
Let's have a call about this when you read this.

FrankKr: I was incorrect and did not interpret the change correctly. This in fact DOES NOT break the production code.

PR #163, opened 2021-09-10
Feature ktps 1543 optimize hyperparams

EnricoASchmitz:
Optuna hyperparameter optimization with unit test for the objective

### DIFF
--- a/openstf/model/objective.py
+++ b/openstf/model/objective.py
@@ -32,6 +32,7 @@ class RegressorObjective:
         # use the objective function
         study.optimize(objective)
     """
+
     def __init__(
         self,
         input_data,
@@ -65,7 +66,7 @@ class RegressorObjective:
         # Check elapsed time and after n minutes end trial and stop the study
         study_duration = datetime.utcnow() - self.start_time
         if study_duration > timedelta(minutes=2, seconds=0):
-            logging.info('Study stopped due to time constraint')
+            logging.info("Study stopped due to time constraint")
             trial.study.stop()
 
         # Perform data preprocessing
@@ -132,6 +133,7 @@ class RegressorObjective:
     def get_loss(self):
         pass
 
+
 class XGBRegressorObjective(RegressorObjective):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -155,9 +157,10 @@ class XGBRegressorObjective(RegressorObjective):
         return {**model_params, **params}
 
     def get_pruning_callback(self, trial: optuna.trial.FrozenTrial):
-        return optuna.integration.XGBoostPruningCallback( trial,
-            observation_key=f"validation_1-{self.eval_metric}"
+        return optuna.integration.XGBoostPruningCallback(
+            trial, observation_key=f"validation_1-{self.eval_metric}"
         )
+
 
 class LGBRegressorObjective(RegressorObjective):
     def __init__(self, *args, **kwargs):
@@ -192,11 +195,9 @@ class LGBRegressorObjective(RegressorObjective):
         metric = self.eval_metric
         if metric == "mae":
             metric = "l1"
-        return optuna.integration.LightGBMPruningCallback(trial,
-                                                          metric = metric,
-                                                          valid_name = "valid_1")
-
-
+        return optuna.integration.LightGBMPruningCallback(
+            trial, metric=metric, valid_name="valid_1"
+        )
 
 
 class XGBQRegressorObjective(RegressorObjective):
@@ -220,6 +221,6 @@ class XGBQRegressorObjective(RegressorObjective):
         return {**model_params, **params}
 
     def get_pruning_callback(self, trial: optuna.trial.FrozenTrial):
-        return optuna.integration.XGBoostPruningCallback( trial,
-            observation_key=f"validation_1-{self.eval_metric}"
+        return optuna.integration.XGBoostPruningCallback(
+            trial, observation_key=f"validation_1-{self.eval_metric}"
         )
--- a/openstf/model/objective.py
+++ b/openstf/model/objective.py
@@ -7,7 +7,7 @@ import optuna
 from openstf.metrics import metrics
 from openstf.model.regressors.regressor_interface import OpenstfRegressorInterface
 from openstf.enums import MLModelType
-from datetime import datetime, timedelta
+from datetime import datetime
 from openstf.model_selection.model_selection import split_data_train_validation_test
 
 EARLY_STOPPING_ROUNDS: int = 10
@@ -15,7 +15,6 @@ TEST_FRACTION: float = 0.1
 VALIDATION_FRACTION: float = 0.1
 # See https://xgboost.readthedocs.io/en/latest/parameter.html for all possibilities
 EVAL_METRIC: str = "mae"
-MAX_DURATION_MINUTES = 2
 
 # https://optuna.readthedocs.io/en/stable/faq.html#objective-func-additional-args
 class RegressorObjective:
@@ -37,7 +36,6 @@ class RegressorObjective:
         self,
         input_data,
         model: OpenstfRegressorInterface,
-        max_duration_minutes=MAX_DURATION_MINUTES,
         test_fraction=TEST_FRACTION,
         validation_fraction=VALIDATION_FRACTION,
         eval_metric=EVAL_METRIC,
@@ -51,7 +49,6 @@ class RegressorObjective:
         self.eval_metric = eval_metric
         self.eval_metric_function = metrics.get_eval_metric_function(eval_metric)
         self.verbose = verbose
-        self.max_duration_minutes = max_duration_minutes
         # Should be set on a derived classes
         self.model_type = None
 
--- a/openstf/model/objective.py
+++ b/openstf/model/objective.py
@@ -37,15 +37,15 @@ class RegressorObjective:
     Union[int, float]:
         """ Pick hyperparameter for tuple  """
         tup, log = value
-        startValue, endValue = tup
+        start_value, end_value = tup
 
         error = "input values for the tuple can be float/float, float/int, int/float, int/int"
 
-        if isinstance(startValue, float) or isinstance(endValue, float):
-            return trial.suggest_float(key, float(startValue), float(endValue), log=log)
+        if isinstance(start_value, float) or isinstance(end_value, float):
+            return trial.suggest_float(key, float(start_value), float(end_value), log=log)
 
-        elif isinstance(startValue, int) and isinstance(endValue, int):
-            return trial.suggest_int(key, startValue, endValue)
+        elif isinstance(start_value, int) and isinstance(end_value, int):
+            return trial.suggest_int(key, start_value, end_value)
 
         else:
             raise TypeError(error)
@@ -113,7 +113,7 @@ class RegressorObjective:
         Returns:
             float: Mean absolute error for this trial.
         """
-        # Check elapsed time
+        # Check elapsed time and after n minutes end trial and stop the study
         elapsed = datetime.utcnow() - self.start_time
         if elapsed > timedelta(minutes=2, seconds=0):
             trial.study.stop()
--- a/openstf/model/objective.py
+++ b/openstf/model/objective.py
@@ -63,12 +63,6 @@ class RegressorObjective:
         Returns:
             float: Mean absolute error for this trial.
         """
-        # Check elapsed time and after n minutes end trial and stop the study
-        study_duration = datetime.utcnow() - self.start_time
-        if study_duration > timedelta(minutes=2, seconds=0):
-            logging.info("Study stopped due to time constraint")
-            trial.study.stop()
-
         # Perform data preprocessing
         train_data, validation_data, test_data = split_data_train_validation_test(
             self.input_data,
@@ -117,6 +111,13 @@ class RegressorObjective:
         return self.eval_metric_function(test_y, forecast_y)
 
     def get_default_params(self, trial: optuna.trial.FrozenTrial) -> dict:
+        """get default parameters function.
+
+                Args: trial
+
+                Returns:
+                    dict: {parameter: hyperparameter_value}
+        """
         # Default parameters
         params = {
             "learning_rate": trial.suggest_float("learning_rate", 0.01, 0.2),
@@ -130,9 +131,23 @@ class RegressorObjective:
         }
         return params
 
-    def get_loss(self):
-        pass
+    def get_params(self, trial: optuna.trial.FrozenTrial) -> dict:
+        """get parameters for objective without model specific get_params function.
 
+                        Args: trial
+
+                        Returns:
+                            dict: {parameter: hyperparameter_value}
+        """
+        default_params = super().get_default_params(trial)
+
+        # Compare the list to the default parameter space
+        model_parameters = self.model.get_params()
+        keys = [x for x in model_parameters.keys() if x in default_params.keys()]
+        # create a dictionary with the matching parameters
+        params = {parameter: default_params[parameter] for parameter in keys}
+
+        return params
 
 class XGBRegressorObjective(RegressorObjective):
     def __init__(self, *args, **kwargs):
@@ -141,6 +156,14 @@ class XGBRegressorObjective(RegressorObjective):
 
     # extend the parameters with the model specific ones per implementation
     def get_params(self, trial: optuna.trial.FrozenTrial) -> dict:
+        """get parameters for XGB Regressor Objective
+        with objective specific parameters.
+
+                                Args: trial
+
+                                Returns:
+                                    dict: {parameter: hyperparameter_value}
+        """
         default_params = super().get_default_params(trial)
 
         # Compare the list to the default parameter space
@@ -168,6 +191,14 @@ class LGBRegressorObjective(RegressorObjective):
         self.model_type = MLModelType.LGB
 
     def get_params(self, trial: optuna.trial.FrozenTrial) -> dict:
+        """get parameters for LGB Regressor Objective
+        with objective specific parameters.
+
+                                Args: trial
+
+                                Returns:
+                                    dict: {parameter: hyperparameter_value}
+        """
         default_params = super().get_default_params(trial)
 
         # Compare the list to the default parameter space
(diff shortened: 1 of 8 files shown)


openstf/model/objective.py
```diff
@@ -55,6 +119,11 @@ def __call__(self, trial: optuna.trial.FrozenTrial) -> float:
         Returns:
             float: Mean absolute error for this trial.
         """
+        # Check elapsed time
+        elapsed = datetime.utcnow() - self.start_time
+        if elapsed > timedelta(minutes=2, seconds=0):
+            trial.study.stop()
```
FrankKr: Very useful functionality! Minor remarks:
For maintainability; please define `MAX_DURATION_MINUTES = 2` at the top of the file, or even better: as a optional function argument, and refer to that variable here. You would have to turn the 'elapsed' then into something with minutes by: `elapsed_minutes=elapsed.total_seconds()/60.`
Also, would be nice to include a logging step if the study is stopped due to time constraint 
`logging.info('Study stopped due to time constraint')`
Suggestion instead of 'elapsed': `study_duration`
Please include this time constraint in the docstring

openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -22,13 +25,59 @@
...
+    "n_estimators": ((50, 150), False),
+    "min_split_gain": ((1e-8, 1), True),
+    "subsample_freq": ((1, 10), False),
+}
+
+
+def get_relevant_model_paramspace(
+    model: OpenstfRegressorInterface, paramspace: dict
+) -> dict:
+    """Return the parameters usefull for the model"""
```
FrankKr, line 65: -> ".. useful for the specific Regressor"

openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -68,11 +117,44 @@ def optimize_hyperparameters_pipeline(
...
+        ] = objective.eval_metric  # The objective of lgb is the eval metric
+        pruning_function = optuna.integration.LightGBMPruningCallback
+        args_eval = {"metric": objective.eval_metric, "valid_name": "valid_1"}
+        if objective.eval_metric == "mae":
+
+            args_eval["metric"] = "l1"
+    else:
+        # for other models use the default objective by not setting it.
+        pruning_function = optuna.integration.XGBoostPruningCallback
+        args_eval = {"observation_key": "validation_1-{}".format(objective.eval_metric)}
```
FrankKr, line 142: let's try and see if we can find a better place to incorporate this distinction. 
Perhaps the 'get_relevant_model_paramspace' can handle this logic, since it already receives the model as an input?

test/unit/model/test_objective.py
```diff
@@ -2,36 +2,128 @@
...
+from openstf.feature_engineering.feature_applicator import TrainFeatureApplicator
+from openstf.model.objective_creator import ObjectiveCreator
+from openstf.model.model_creator import ModelCreator
+from openstf.pipeline.optimize_hyperparameters import get_relevant_model_paramspace
+
+input_data = TestData.load("reference_sets/307-train-data.csv")
+input_data_with_features = TrainFeatureApplicator(horizons=[0.25, 24.0]).add_features(
+    input_data
+)
+input_data_with_features = input_data_with_features.iloc[::50, :]
```
FrankKr, line 18: why select ::50? I would think you would use the whole train dataset

test/unit/model/test_objective.py
```diff
@@ -2,36 +2,128 @@
...
+pj = TestData.get_prediction_job(pid=60)
```
FrankKr, line 19: not pid=307, since thats the traindata you're loading?

test/unit/model/test_objective.py
```diff
@@ -2,36 +2,128 @@
...
+
+# LGB specific
+lgb_paramspace: dict = {
+    "num_leaves": ((16, 62), False),
+    "boosting_type": ["gbdt", "dart", "rf"],
+    "tree_learner": ["serial", "feature", "data", "voting"],
+    "n_estimators": ((50, 150), False),
+    "min_split_gain": ((1e-8, 1), True),
+    "subsample_freq": ((1, 10), False),
+}
```
FrankKr, line 48: I would not expect these definitions here.
If you really need them, import them from where you already defined them

FrankKr (commented):
Looks like a good improvement.
Some minor remarks as comments in the code.

Additionally; two bigger points of interest;
- I would expect a unit test for optimize_hyperparameters_pipeline, since that has been majorly revised
- there is still some model-specific logic in optimize_hyperparameters.py, lets discuss where we could place that (see also the comment in the code)

openstf/model/objective.py
```diff
@@ -2,11 +2,13 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 import optuna
+from typing import Union, Tuple
 
+import openstf  # Needed for <loss = "openstf.metrics.metrics." + self.eval_metric>  and the line <metric = eval(loss)(test_y, forecast_y)>
```
bremme, line 7: Good that you added a comment on why this is required. But I think we can make a more clear solution to this issue.

openstf/model/objective.py
```diff
@@ -31,19 +33,83 @@ class RegressorObjective:
         study.optimize(objective)
     """
 
+    def process_tuple(
+        self,
+        trial: optuna.trial.FrozenTrial,
+        key: str,
+        value: Tuple[Union[int, float], Union[int, float]],
+    ) -> Union[int, float]:
+        """Pick hyperparameter for tuple"""
```
bremme, line 42: I will probably have to read the code a bit further. But it is not immediately clear what this function does? It does some processing on a tuple and/or it picks a hyperparameter for a tuple. Probably a good idea to bring the method name more in line with the description now it looks like two different things.

openstf/model/objective.py
```diff
@@ -31,19 +33,83 @@ class RegressorObjective:
...
+        tup, log = value
```
bremme, line 43: This is also not really clear to me. What does value, tup and log represent?

openstf/model/objective.py
```diff
@@ -31,19 +33,83 @@ class RegressorObjective:
...
+        elif isinstance(start_value, int) and isinstance(end_value, int):
+            return trial.suggest_int(key, start_value, end_value)
+
+        else:
+            raise TypeError(error)
+
+    def get_trial_parameters(
+        self, model_params: dict, trial: optuna.trial.FrozenTrial
+    ) -> dict:
+        """
```
bremme, line 62: Nice to add a docstring! A couple minor remarks. The title should go on a single line and an additional description below. Furthermore you could also describe the arguments and return optionally.

In your description you refer to "the dictionairy" its more clear to name the argument: `model_params` can contain etc. 

Also a dictionairy had key values, you are probably talking about the values in the dictionairy?

Also some lines seem to be over 88 chars. Our formatter black can do a lot but it does not split docstring or comments.

openstf/model/objective.py
```diff
@@ -31,19 +33,81 @@ class RegressorObjective:
...
+            if isinstance(value, tuple):
+                param[key] = self.process_tuple(trial, key, value)
+            # If the parameter is a list it means it is a categorical feature
+            elif isinstance(value, list):
+                param[key] = trial.suggest_categorical(key, value)
+
+            else:
+                raise TypeError("Possible input values: tuple, list")
+
+        return param
```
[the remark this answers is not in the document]
bremme: or `params` perhaps.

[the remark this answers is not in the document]
bremme: Or even better `trial_parameters` -> the `get_trial_parameters` function returns `trial_parameters`. I'm also fin with using params instead of parameters ofc.

openstf/model/objective.py
```diff
@@ -31,19 +33,83 @@ class RegressorObjective:
...
+
     def __init__(
```
bremme, line 90: The `__init__` method should always be the first method in the class, better to move to the top.

openstf/model/objective.py
```diff
@@ -31,19 +33,83 @@ class RegressorObjective:
...
         self,
         input_data,
+        model: OpenstfRegressorInterface,
+        pruning_function: optuna.integration,
```
bremme, line 94: `optune.integration` is a package I think not an interface so this seems incorrect?

openstf/model/objective.py
```diff
@@ -31,19 +33,83 @@ class RegressorObjective:
...
+        model_params: dict,
```
bremme, line 95: Is it required to use this argument? Does model.get_params() not return the same? If so, this interface is already a bit too large if we can make it smaller that would be better in my opinion.

openstf/model/objective.py
```diff
@@ -31,19 +33,83 @@ class RegressorObjective:
...
+        start_time: datetime,
```
bremme, line 96: Does this need to be an input argument? Why could the objective function itself not keep track of time?

openstf/model/objective.py
```diff
@@ -55,6 +121,11 @@ def __call__(self, trial: optuna.trial.FrozenTrial) -> float:
         Returns:
             float: Mean absolute error for this trial.
         """
+        # Check elapsed time and after n minutes end trial and stop the study
+        elapsed = datetime.utcnow() - self.start_time
```
bremme, line 125: Why is this custom stopping criteria needed? Does optuna not have all kind of functionality to determine when to stop?

openstf/model/objective.py
```diff
@@ -55,6 +121,11 @@ def __call__(self, trial: optuna.trial.FrozenTrial) -> float:
...
+        if elapsed > timedelta(minutes=2, seconds=0):
+            trial.study.stop()
+
         # Perform data preprocessing
         train_data, validation_data, test_data = split_data_train_validation_test(
             self.input_data,
```
bremme, line 131: Minor remark. Perhaps not part of this PR, but this seems a little odd place to check if the data is in the right format. Probably better to have a single unit which is responsible for this check and perhaps it makes more sense to check it at construction.

openstf/model/objective.py
```diff
@@ -78,49 +149,49 @@ def __call__(self, trial: optuna.trial.FrozenTrial) -> float:
         # Configure evals for early stopping
         eval_set = [(train_x, train_y), (valid_x, valid_y)]
 
-        model = ModelCreator.create_model(self.model_type)
-
-        params = self.get_params(trial)
+        # get the parameters used in this trial
+        param = self.get_trial_parameters(self.model_params, trial)
```
bremme, line 153: Like the previous comment, something like `trial_params` or `trial_parameters` would be more clear.

bremme: Like Frank also mentioned, if this is a method only for internal use you should prefix it with `_` and in that case you don't need to pass the `self.model_params` since the method will have access to self anyway.

openstf/model/objective.py
```diff
@@ -78,49 +149,49 @@ def __call__(self, trial: optuna.trial.FrozenTrial) -> float:
...
-            eval_metric=eval_metric,
+            eval_metric=self.eval_metric,
             callbacks=[pruning_callback],
         )
 
-        forecast_y = model.predict(test_x)
+        forecast_y = self.model.predict(test_x)
 
-        return mae(test_y, forecast_y)
+        try:
```
bremme, line 174: Perhaps we can make this a little nicer. A good rule is always to fail as early as possible, so in this case the constructor should already fail if the right evaluation metric is not available. I would suggest something like this:

```python
# at the top of the objective
from openstf.metrics import metrics

# in the RegressorObjective construction
def __init__():
    self.eval_metric_function = metrics.get_metric_eval_function(eval_metric)

# and in the __call__ method
def __call__():
    return self.eval_metric_function(test_y, forecast_y)

# in openstf.metric.metrics add a get_metric_evaluation_function
def get_metric_eval_function(metric_name):
    evaluation_function = {
        "bias": bias,
        "nsme": nsme,
        "mae": mae,
        "r_mae": r_mae,
        "frac_in_stdev": frac_in_stdev,
        "r_mae_highest": r_mae_highest,
        "r_mne_highest": r_mne_highest,
        "r_mpe_highest": r_mpe_highest,
        "r_mae_lowest": r_mae_lowest,
        "skill_score": skill_score,
        "skill_score_positive_peaks": skill_score_positive_peaks,
        "franks_skill_score": franks_skill_score,
        "franks_skill_score_peaks": franks_skill_score_peaks,
    }.get(metric_name, None)

    if evaluation_function is None:
        raise KeyError(f"Unknown evaluation metric function {metric_name}")

    return evaluation_function

```

openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -22,13 +25,59 @@
...
+    "num_leaves": ((16, 62), False),
+    "boosting_type": ["gbdt", "dart", "rf"],
+    "tree_learner": ["serial", "feature", "data", "voting"],
+    "n_estimators": ((50, 150), False),
+    "min_split_gain": ((1e-8, 1), True),
+    "subsample_freq": ((1, 10), False),
+}
+
+
+def get_relevant_model_paramspace(
```
bremme, line 62: Let try to be a bit more consistent with our naming. It looks like we use various terms for the same things? trial_params, model_params, hyper_params, relevant_model_params. We are probably talking about 1-2 things but we have more names than that it looks.

openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -22,13 +25,59 @@
...
+    "lambda": ((1e-8, 1.0), True),
+    "subsample": ((0.5, 0.99), False),
+    "min_child_weight": ((1, 6), False),
+    "max_depth": ((3, 10), False),
+    "colsample_bytree": ((0.5, 1.0), False),
+    "max_delta_step": ((1, 10), False),
+}
+# Important parameters, model specific
+# XGB specific
+xgb_paramspace: dict = {
```
bremme, line 45: I think the model specific stuff should probably not be part of the pipeline? Think about responsibilities? The pipeline should probably be only responsible for conduction the optimization and not have any knowledge about different model implementations. I guess the model specific stuff should either be part of a model specific optimizer or if its really small perhaps part of the model itself.

openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -68,11 +117,44 @@ def optimize_hyperparameters_pipeline(
...
+    model_type = pj["model"]
+    model = ModelCreator.create_model(model_type)
+
+    # Combine the default parameters with all model specific parameters
+    paramspace = default_paramspace.copy()
+    paramspace.update(**xgb_paramspace, **lgb_paramspace)
+
+    model_params = get_relevant_model_paramspace(model, paramspace)
+
+    if model_type == "lgb":
```
bremme, line 129: Like my previous comment, this model specific stuff should probably be part of the objective.

openstf/pipeline/optimize_hyperparameters.py
bremme: I would suggest to have the optimization model specific stuff in only one place and keep it out of the pipeline. The pipeline can just use the factories to get a model and objective and run the optuna optimization and does not have to be responsible for the model specific stuff.

```shell
openstf.model.objectives -> hold the objectives for the various implementations (xgb.py, lgbm.py etc.)
  can be one file for now if we want instead of a package ofc.

# Or
openstf.model.regressors.objectives > hold the objectives for the various implementations (xgb.py, lgbm.py etc.)
  can be one file for now if we want instead of a package ofc.
# Or 
openstf.model.regressors.<model-name> -> model.py and objective.py

```

test/unit/model/test_objective.py
bremme: It depends on the size of the train set I guess. Our unit test are already really slow and optimization with a large set could add a lot of time perhaps? I would say the goal of the test is to run though all paths of the code and not so much to actually deliver the best optimization on a large set.

test/unit/model/test_objective.py
```diff
@@ -2,36 +2,128 @@
...
 class TestXGBRegressorObjective(BaseTestCase):
-    def test_call(self, model_creator_mock, mea_mock, split_data_mock):
+    def test_call(self):
+        model_type = "xgb"
+        model = ModelCreator.create_model(model_type)
+
+        default_paramspace.update(**xgb_paramspace, **lgb_paramspace)
+        model_params = get_relevant_model_paramspace(model, default_paramspace)
+        start_time = datetime.utcnow()
+        objective = ObjectiveCreator.create_objective(model_type)
```
bremme, line 61: Maybe not a big deal but your also testing the creator here. You should probably just test the various objectives directly here, i.e.: `XGBRegressorObjective()` etc.

bremme:
This is a nice and important addition to our project and you made a good start! I do however have a couple remark (I've also added some comments to the code).

As far as I can tell the difference in optimization between the various models are the parameters / parameter space,
 and the pruning callback

openstf/model/objective.py
EnricoASchmitz: The custom stopping criteria allows us to use trial.study.stop() this will first finish the trial and after the trial finished it will stop the study. Using the time limit from optuna it will stop the study during a trial.

openstf/model/objective.py
EnricoASchmitz: Does this comment belong to this part of the code?
If so please explain a bit further.

openstf/model/objective.py
bremme: How do you know this: "Using the time limit from optuna it will stop the study during a trial". I would think `optune.study.Study.optimize` would check total duration after a trial?

openstf/model/objective.py
bremme: image
The comment was about the part right below. But is probably not part of this PR. I guess we can just put in on resolve for now.

openstf/model/objective.py
```diff
@@ -78,49 +92,135 @@ def __call__(self, trial: optuna.trial.FrozenTrial) -> float:
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.model_type = MLModelType.XGB
 
+    # extend the parameters with the model specific ones per implementation
     def get_params(self, trial: optuna.trial.FrozenTrial) -> dict:
+        default_params = super().get_default_params(trial)
+
+        # Compare the list to the default parameter space
+        model_parameters = self.model.get_params()
```
bremme, line 147: I'm not sure if I understand this part? Why are the parameters we want to optimize per model type not static? Or are some of the default parameters not used in all models?

openstf/model/objective.py
```diff
@@ -78,49 +92,135 @@ def __call__(self, trial: optuna.trial.FrozenTrial) -> float:
...
+    def get_pruning_callback(self, trial: optuna.trial.FrozenTrial):
+        metric = self.eval_metric
+        if metric == "mae":
+            metric = "l1"
+        return optuna.integration.LightGBMPruningCallback(
+            trial, metric=metric, valid_name="valid_1"
+        )
+
+
+class XGBQRegressorObjective(RegressorObjective):
```
bremme, line 203: I think we should use `XGBQuantileRegressorObjective` here. The one letter difference (compared to `XGBRegressorObjective`) is a little harder to see we use `XGBQuantileOpenstfRegressor` for the model as well.

bremme (approved):

PR #165, opened 2021-09-14
Feature ktps 1552 feature & importance

EnricoASchmitz:


openstf/metrics/reporter.py
```diff
@@ -91,39 +91,4 @@ def _make_data_series_figures(self, model: RegressorMixin) -> dict:
         }
 
     def _make_feature_importance_figure(self, model: RegressorMixin) -> None:
```
FrankKr, line 93: Since this function now is only a single line, and the function is 'private' (due to it starting with an underscore), perhaps you can get rid of it altogether?

openstf/model/objective.py
```diff
@@ -164,10 +164,10 @@ def get_params(self, trial: optuna.trial.FrozenTrial) -> dict:
...
 
-                                Args: trial
+            Args: trial
 
-                                Returns:
-                                    dict: {parameter: hyperparameter_value}
+            Returns:
+                dict: {parameter: hyperparameter_value}
         """
         default_params = super().get_default_params(trial)
```
FrankKr, line 172: It looks like this part is identical for all models. Is it possible to move it to the parent-class?

openstf/model/objective_creator.py
```diff
@@ -9,27 +9,24 @@
...
+    XGBQuantileRegressorObjective,
 )
 
 
 class ObjectiveCreator:
     OBJECTIVES = {
         MLModelType.XGB: XGBRegressorObjective,
         MLModelType.LGB: LGBRegressorObjective,
-        MLModelType.XGB_QUANTILE: XGBQuantileRegressor,
+        MLModelType.XGB_QUANTILE: XGBQuantileRegressorObjective,
```
FrankKr, line 20: Well spotted!

openstf/model/regressors/regressor.py
```diff
@@ -0,0 +1,32 @@
+# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
+#
+# SPDX-License-Identifier: MPL-2.0
+import numpy as np
+import pandas as pd
+from openstf.model.regressors.regressor_interface import OpenstfRegressorInterface
+
+
+class OpenstfRegressor(OpenstfRegressorInterface):
```
FrankKr, line 9: Makes sense to specify the shared functionality somewhere.
Out of curiosity, was this extra class a tip from Bram?

openstf/model/regressors/xgb_quantile.py
```diff
@@ -4,18 +4,22 @@
...
+import pandas as pd
 
 import openstf.metrics.metrics as metrics
 
 DEFAULT_QUANTILES: Tuple[float, ...] = (0.9, 0.5, 0.1)
 
 
-class XGBQuantileOpenstfRegressor(OpenstfRegressorInterface):
+class XGBQuantileOpenstfRegressor(OpenstfRegressor):
+    gain_importance_name = ("total_gain",)
```
FrankKr, line 20: It looks weird that this isn't just a string?

test/unit/pipeline/test_train_model.py
```diff
@@ -140,6 +142,32 @@ def test_train_model_pipeline_young_model(self, serializer_mock, pipeline_mock):
         )
         self.assertFalse(pipeline_mock.called)
 
+    def test_feature_importance(self):
```
FrankKr, line 145: Good idea to add a unit test for the feature importance functionality!
I think currently this is only tested for a single machine learning model type, correct?
In fact, looking at this test, I think the training pipeline altogether is only tested for a single model type. 

Perhaps you could change the test so that the `test_train_model_pipeline_core_happy_flow` iterates over all model types specified in `openstf.enums.MLModelType`? You could than add the test of `model.get_feature_importance` there as well, instead of training a model specifically for that.

FrankKr (changes requested):
Overall comments:
- Looks good, nice separation of generic and model-specific code!
- adding a small description to your PR can help the reviewer. Especially useful to already point them to certain aspects you are unsure about or would like to receive some feedback on.
- Added some minor remarks throughout
- Biggest suggestion is an improvement of the existing train_model_pipeline unit test. Let's have a call if my comment is unclear.

openstf/model/regressors/regressor.py
EnricoASchmitz: Yes, I pitched Bram an idea how to make it a shared function inside the interface and Bram said that it might be best to add a layer between it. To keep the interface an actual interface only.

test/unit/pipeline/test_train_model.py
EnricoASchmitz: test_train_model_pipeline_core_happy_flow gets the pj from prediction_jobs.json. In this file we only have a xgb model. And since it gets the model type from the pj I can't loop it.

FrankKr: Let's have a look at it together, I have good hopes that we can make it work :)

FrankKr (approved):
Approved! Please change the model.get_feature_names() method before merging, as discussed

PR #166, opened 2021-09-15
Logging now with pj_id

### LINKED ISSUE #164: Feature: add logging when suspicious moments are detected
In the Openstf pipelines, input data is checked for 'suspicious moments'. E.g. flatliners. 
These moments are corrected/left out.
New feature: log when this occurs, and for which prediction job id it occurred. This can enable the (automatic) follow-up to get broken measurement devices fixed.

Suggestion for implementation: somewhere here: https://github.com/alliander-opensource/openstf/blob/504222aa4e41cca9e5e310abeacc4dbf66ee8ceb/openstf/validation/validation.py#L24

EnricoASchmitz:
Small changes to fix the logging and add the pj_id to the logging.
Resolves #164 .

### DIFF
--- a/openstf/validation/validation.py
+++ b/openstf/validation/validation.py
@@ -23,7 +23,9 @@ FLATLINER_TRESHOLD = 24
 
 
 def validate(
-    pj_id: Union[int, str], data: pd.DataFrame, flatliner_threshold: int = FLATLINER_TRESHOLD
+    pj_id: Union[int, str],
+    data: pd.DataFrame,
+    flatliner_threshold: int = FLATLINER_TRESHOLD,
 ) -> pd.DataFrame:
     logger = structlog.get_logger(__name__)
     # Drop 'false' measurements. e.g. where load appears to be constant.
@@ -34,7 +36,8 @@ def validate(
     if num_const_load_values > 0:
         logger.warning(
             f"Changed {num_const_load_values} values of constant load to NA.",
-            pj_id=pj_id, num_const_load_values=num_const_load_values,
+            pj_id=pj_id,
+            num_const_load_values=num_const_load_values,
         )
 
     # Check for repeated load observations due to invalid measurements
@@ -47,7 +50,8 @@ def validate(
         num_nan = sum([True for i, row in data.iterrows() if all(row.isnull())])
         logger.warning(
             "Found suspicious data points, converted to NaN value",
-            pj_id=pj_id, num_nan_values=num_nan,
+            pj_id=pj_id,
+            num_nan_values=num_nan,
         )
     return data
 
--- a/openstf/validation/validation.py
+++ b/openstf/validation/validation.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: MPL-2.0
 
 from datetime import timedelta
+from typing import Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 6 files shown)


bremme:
A couple general remarks:

* You description op the PR is a bit short
    * Its not clear what you exactly fixed
    * And more important why you fixed it in a certain way
* This is not a bugfix, but a feature in my opinion (its something new I would say)
* The branchname `<type>-<ticket>-logging` does not decribe the bugfix or feature properly
    * a better alternative would be: feature-KTPS-1596-improve-data-validation-logging for example
* I would not combine Jira and github issues in the branchname.

In general I think our `validation.validate()` function mixes too many responsibilities now and hence the naming is incorrect. It does:

* Fix/manipulate the data
* It reports on which/how many manipulation have been conducted
  
But more important it does not really validate/check if the data is correct itself it seems.

In general I think we should split the responsibilities properly first. For example:

* validation.validation module
  * Contains functions which do only validation
    * don't have knowledge about logging and prediction jobs    
* preprocessing.preprocessing module
  * Contains function which only do preprocessing
    * don't have knowledge about logging and prediction jobs
* preprocessing.preprocessor module
  * Contains a Preprocessor class
    * which first validates (check only)
    * logs validation results
    * preprocess (manipulation of data)
    * takes a prediction job as an argument
    * Can optionally be configured which preprocessing steps need to run?

It could look a bit like this:

```python
# pipeline
def some_pipeline():
    preprocessor = PreProcessor()

    preprocessor.process(pj, input_data)

# PreProcessor
import validation
import preprocessing

class PreProcessor:

    def __init__(self):
        # perhaps the required preprocessing steps could be set if required here

    def process(pj, input_data):
        repeated_values_index = validation.find_repeated_values(input_data)

        if repeated_values_index:
            self.logger.warning(
                "Found repeated values", 
                preprocessing_step="repeated_values", 
                num_values=len(repeated_values_index),
                pid=pj["id"]
            )

            input_data = preprocessing.replaces_values_at_index(
                index=repeated_values_index, data=input_data, value=np.nan
            )

        nonzero_flatliners_index = validation.find_nonzero_flatliners(input_data)

        if nonzero_flatliners_index:
            self.logger.warning(
                "Found nonzero flatliner values", 
                preprocessing_step="nonzero_flatliners",
                num_values=len(nonzero_flatliners_index),
                pid=pj["id"]
            )

        input_data = preprocessing.replaces_values_at_index(
            index=nonzero_flatliners_index, data=input_data, value=np.nan
        )

        return input_data

```

But other implementation could work as well. But I think its a good idea to separate the responsibilities of data validation, data preprocessing and doing the performing the validation and preprocessing steps and logging. Perhaps we could even move more towards sklearn, and see if we can use their preprocessing interface. But we should also keep in mind to not make this story too big I guess.

openstf/validation/validation.py
```diff
@@ -22,18 +23,22 @@
...
     data = replace_repeated_values_with_nan(
         data, max_length=flatliner_threshold, column_name=data.columns[0]
     )
     num_const_load_values = len(data) - len(data.iloc[:, 0].dropna())
-    logger.debug(
-        f"Changed {num_const_load_values} values of constant load to NA.",
-        num_const_load_values=num_const_load_values,
-    )
+    if num_const_load_values > 0:
+        logger.warning(
```
bremme, line 37: In general I think we should log:
* An identifier for the job (e.g. prediction job id (like you already added)
* An identifier about the type of validation/preprocessing step which was conducted
* Some (if possible generic) metrics on that validation/preprocessing step
  * Probably the number of datapoints and or the relative number of datapoints which have been modified is a good metric.

bremme:
In sklearn every preprocessing step is modeled as a transformer, perhaps this is useful: https://scikit-learn.org/stable/modules/preprocessing.html#custom-transformers. We probably want something similar or even just use the same interface. But this does not solve the main issue of this PR ofc.

image

PR #173, opened 2021-09-22
Also log 0 if no values are incorrect during validation

EnricoASchmitz:
Now logs 0 if during data cleansing (zero flatliners or repeated values) no values are changed to NaN.


JanMaartenvanDoorn (approved):
Looks good!

PR #175, opened 2021-09-23
Bugfix include quantile models in basecase

JanMaartenvanDoorn:
Basecase forecasts were not made for quantile models, fixed this by including them in the list

### DIFF
--- a/openstf/tasks/create_basecase_forecast.py
+++ b/openstf/tasks/create_basecase_forecast.py
@@ -69,7 +69,7 @@ def main():
     taskname = Path(__file__).name.replace(".py", "")
 
     with TaskContext(taskname) as context:
-        model_type = ["xgb", "lgb"]
+        model_type = ["xgb", "xgb_quantile", "lgb"]
 
         PredictionJobLoop(context, model_type=model_type).map(
             create_basecase_forecast_task, context
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ def read_long_description_from_readme():
 
 setup(
     name="openstf",
-    version="2.4.0",
+    version="2.4.1",
     packages=find_packages(include=["openstf", "openstf.*"]),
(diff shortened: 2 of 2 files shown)


openstf/tasks/create_basecase_forecast.py
```diff
@@ -69,7 +69,7 @@ def main():
     taskname = Path(__file__).name.replace(".py", "")
 
     with TaskContext(taskname) as context:
-        model_type = ["xgb", "lgb"]
+        model_type = ["xgb", "xgb_quantile", "lgb"]
```
bremme, line 72: Since we are calling these task from another application. Perhaps it makes sense to give the `main()` function of the task input arguments. I think it could be useful to be able to run the task with different parameters for example.

```python
def main(model_type: Optional[List[str]] = None, t_behind_days: Optional[int] = None, t_ahead_days: Optional[int] = None):\
    if not model_type:
        model_type = DEFAULT_MODEL_TYPE

    if not t_behind_days:
        t_behind_days = DEFAULT_T_BEHIND_DAYS

    if not t_ahead_days:
        t_ahead_days = DEFAULT_T_AHEAD_DAYS
```

Not sure if we should change it right now, but might be a nice addition. If we had such a mechanism this PR would not be required since we could just configure the task from the client calling code.

bremme:
Looks nice. Have you tested that creating a basecase forecast actually works for a `xgb_quantile` model?

bremme (approved):

openstf/tasks/create_basecase_forecast.py
bremme: This woulc also be usefull for other (future) users of our library who don't like our hardcoded values.

JanMaartenvanDoorn: yes I agree, I will make a story for this

PR #176, opened 2021-09-23
Fixed bug

JanMaartenvanDoorn:
- Fixed bug in hyper parameter optimization
- Also fixed bug in taskcontext were the config argument was missing when the databae is initialized


FrankKr (approved):
Looks good!

PR #179, opened 2021-09-27
Changed logging to have pid as key instead of in the message.

EnricoASchmitz:
Changed the logging to not include the pid in the message and instead in use it as a key.

### DIFF
--- a/openstf/pipeline/optimize_hyperparameters.py
+++ b/openstf/pipeline/optimize_hyperparameters.py
@@ -6,6 +6,7 @@ import optuna
 from typing import List
 import structlog
 
+from openstf.exceptions import InputDataInsufficientError
 from openstf.model.model_creator import ModelCreator
 from openstf.model.objective_creator import ObjectiveCreator
 from openstf.feature_engineering.feature_applicator import TrainFeatureApplicator
@@ -54,7 +55,7 @@ def optimize_hyperparameters_pipeline(
 
     # Check if sufficient data is left after cleaning
     if not validation.is_data_sufficient(validated_data):
-        raise ValueError(
+        raise InputDataInsufficientError(
             f"Input data is insufficient for {pj['name']} "
             f"after validation and cleaning"
         )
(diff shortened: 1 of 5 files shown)


openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -54,7 +55,7 @@ def optimize_hyperparameters_pipeline(
 
     # Check if sufficient data is left after cleaning
     if not validation.is_data_sufficient(validated_data):
-        raise ValueError(
+        raise InputDataInsufficientError(
```
FrankKr, line 58: Good job replacing the generic exception for a specific one!

test/unit/pipeline/test_train_model.py
```diff
@@ -155,6 +157,21 @@ def test_train_model_pipeline_young_model(self, serializer_mock, pipeline_mock):
         )
         self.assertFalse(pipeline_mock.called)
 
+    @patch("openstf.pipeline.train_model.validation")
+    def test_train_model_InputDataInsufficientError(self, validation_is_data_sufficient_mock):
```
FrankKr, line 161: Great to add these unit tests!

test/unit/pipeline/test_train_model.py
```diff
@@ -155,6 +157,21 @@ def test_train_model_pipeline_young_model(self, serializer_mock, pipeline_mock):
...
+        validation_is_data_sufficient_mock.is_data_sufficient.return_value = False
+        with self.assertRaises(InputDataInsufficientError):
+            train_model_pipeline_core(pj=self.pj, input_data=self.train_input)
+
+    def test_train_model_InputDataWrongColumnOrderError(self):
+        input_data = self.train_input
+        input_data["load2"] = input_data["load"]
+        input_data = input_data.drop("load", axis=1)
+        input_data["load"] = input_data["load2"]
+        input_data = input_data.drop("load2", axis=1)
```
FrankKr, line 171: You could help a (future) colleague by adding a small comment of what this code does (I think it moves the first column to the end?).
Also, perhaps you can make it a bit shorter:
`input_data = input_data.iloc[:,1:].merge(input_data.iloc[:,[0]], left_index=True, right_index=True)` 
(untested code snippit)

FrankKr (approved):
Looks good!

PR #181, opened 2021-09-28
make exception more specific

FrankKr:
Relates to KTPS-1653; Raises a more specific exception when stdev is missing.
Discussed with JM, and we decided that we do want to raise an exception and not try to fix it in place.


EnricoASchmitz (approved):
Looks good.
Should it log the fact that the error was raised?

FrankKr:
> Looks good. Should it log the fact that the error was raised?

The idea is that if the exception is raised but not captured, it will show up in the logging.
And that if the exception is raised, but also captured, there is no need to log is. 
So I think it's not needed

PR #182, opened 2021-09-29
fix datetime format and remove model type selection

FrankKr:
Could not reproduce bug, perhaps the conditions do not occur anymore?
Did fix two minor other issues. Using the date() instead of datetime() gave an exception with latest version of openstf-dbc


JanMaartenvanDoorn (approved):
Looks good!

PR #208, opened 2021-10-27
add feature_names to apply_features

EnricoASchmitz:
apply_features was not getting passed the feature names


EnricoASchmitz:
Should fix #212

BenoitHardier (approved):
Nice work!  Indeed it resolves #212.

PR #210, opened 2021-10-27
Add feature to make dynamic forecasting start and end time based on forecast data instead of horizon

Davidswinkels:


openstf/pipeline/utils.py
```diff
@@ -1,17 +1,26 @@
...
 def generate_forecast_datetime_range(
-    resolution_minutes: int, horizon_minutes: int
+    forecast_data: pd.DataFrame,
 ) -> Tuple[datetime, datetime]:
-    # get current date and time UTC
-    datetime_utc = datetime.utcnow()
-    # Datetime range for time interval to be predicted
-    forecast_start = datetime_utc - timedelta(minutes=resolution_minutes)
-    forecast_end = datetime_utc + timedelta(minutes=horizon_minutes)
+    """Generate forecast range based on last cluster of null values in forecast data."""
```
FrankKr, line 13: A slightly more detailed docstring would be very helpful for someone looking at this code for the first time:)

openstf/tasks/create_forecast.py
```diff
@@ -36,7 +36,7 @@
 from openstf.tasks.utils.taskcontext import TaskContext
 
 T_BEHIND_DAYS: int = 14
-T_AHEAD_DAYS: int = 3
+T_AHEAD_DAYS: int = 2
```
FrankKr, line 39: Not wrong necessarily, but why is this variable changed in this PR?

FrankKr (approved):
Looks good.
Nice generalization

openstf/pipeline/utils.py
Davidswinkels: Added more detailed dosctring with an example :)

openstf/tasks/create_forecast.py
Davidswinkels: Before we specified the forecast horizon in 48 hours from datetime.now() when calling generate_forecast_datetime_range(). 

Now we have a dynamic forecast range based on the forecast data, and we don't use datetime.now()  and forecast horizon of 48 hours anymore. If we want to keep a forecast horizon of 48 hours, we should limit T_AHEAD_DAYS to 2 days so we retrieve data that is 2 days in the future. If we don't change this value, we will retrieve forecast data 3 days into the future and all forecasts will be 3 days ahead instead of 2.

Davidswinkels: @FrankKr Can you resolve this conversation if it's clear? Or if changes are still required, it would be nice to hear!

FrankKr: Ok, makes sense!

AlexisMignon (commented):
So, as I understand it, the forecast range is based on the last bock of nan values. If it is done this way, I assume it is because we suppose that it is possible in general to allow nan values. So I wonder what would be the expected behaviour in the following cases:

- the last nan block is not the last block of the dataframe
- what if there are missing values in the last history steps and we want to apply a model for which the forecast horizon matters (note that missing values in the target does not necessarily implies missing values in other explanatory variables).


Clearly the first case is not really expected in a forecast setting, but it could happen in a backtest setting unless, in this use case, the target column is  explicitly set to nan.

Instead of detecting nan blocks, wouldn't it be easier to take the time step following that last non-null target value ?

For the second case, while this modification indeed makes the code more flexible, it does not seem to fit all possible use cases (it is may be not necessary).

Another possibility would be to have the possibility to have a forecast_start or run_date information in the prediction job, which, if not present, could be evaluated to "now".

Davidswinkels:
Thanks for the review Alexis. Yes correct. We now select the last nan block to indicate when start of forecast is required.

In the case when last nan block still has data after it, we can handle it in two ways:

- Set forecast window end at end of last nan block
- Set forecast window end at end of given forecast data (currently implemented)

In the case when horizon matters (so start and end window), we should validate input forecast data to adhere with it's nan block to the required start and end window and does not have missing values in the target column before the required run_date.

Other option to select the last not Null value as start is possible aswell, but it's trickier to implement, because we would have to select last not Value first and then an offset. While with this nan block approach we can directly select the first time of the last nan block. Plus if we select the last not null value, we would have to validate that this last not value is not beyond the horizon. Logically thinking we want to make forecasts where we don't have data (nan data) of the target column.

Yes good point that we would want to have a forecast_start or run_date in the prediction job. That would be the next step to implemenent, and would not be part of this PR. When adding forecast_start / run_date in the prediction job we can add these steps:
- Add run_date in prediction job
- Set run_date on given run_date of prediction job. If not given, use datetime.now()
- Get input forecast data based on run_date and horizon of prediction job
- Validate input forecast data has correct missing values in target column according to the run_date and horizon
- Give input forecast data to generate_forecast_datetime_range() and continue create_forecast_pipeline_core() in openstf as normal

Davidswinkels:
Hmm even simpler solution would be to ditch the generate_forecast_datetime_range() function and to set forecast_start based on given run_date from prediction job (or datetime.now()) and to set forecast_end at run_date + horizon ^^ I did not think of that before since we we were using datetime.now() twice

AlexisMignon:
Regarding the use of the last nonnull value, assuming the values are sorted by time, it could be as simple as:

```python
# Create a dataframe with nans at the end
df = pd.DataFrame(index=pd.date_range("2021-01-01", "2021-06-01"))
np.random.seed(11)
df["x"] = np.where(
    np.random.rand(len(df)) < 0.3,
    np.random.rand(len(df)),
    np.nan
)
# Find the forecast start time
ilast = np.where(df.iloc[:, 0].notnull())[0][-1] 
# The [0] is because where returns a tuple of index arrays (1 for each dimension) when used with a condition only. With 1d arrays there is only one element in the tuple
forecast_start = df.iloc[ilast + 1].name  # Because the index in the records becomes the name of the row series
```

Davidswinkels:
Yes that would work. np.where is a nice solution to convert the series with time index to row based index.
If we do use run_date and horizon later, the generate_forecast_datetime_range() might not be needed anymore though

AlexisMignon:
I agree that a run_date based solution would not need this any more and that we need a quick fix first. So do you plan to use the `np.where` based approach or do you prefer to keep the current implementation. In the last case, I can approve the current implementation.

Davidswinkels:
Maybe keep the current implementation for now and change to run_date + horizon soon, and then remove this function

PR #213, opened 2021-11-01
Bugfix ktps 1758 optimize hyperparams inf nan

FrankKr:
Cause for bug if nan values were in target:
Data was already being validated in the pipeline, but instead of using the validated data, the non-validated input was used

bonus: added functionality in PredictionJobLoop to debug a specific pid more easily

### DIFF
--- a/openstf/pipeline/optimize_hyperparameters.py
+++ b/openstf/pipeline/optimize_hyperparameters.py
@@ -73,9 +73,9 @@ def optimize_hyperparameters_pipeline(
             f"after validation and cleaning"
         )
 
-    input_data_with_features = TrainFeatureApplicator(horizons=horizons).add_features(
-        input_data
-    )
+    validated_data_with_features = TrainFeatureApplicator(
+        horizons=horizons
+    ).add_features(validated_data)
 
     # Create objective (NOTE: this is a callable class)
     objective = ObjectiveCreator.create_objective(model_type=pj["model"])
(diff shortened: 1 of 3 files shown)


openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -73,9 +73,9 @@ def optimize_hyperparameters_pipeline(
             f"after validation and cleaning"
         )
 
-    input_data_with_features = TrainFeatureApplicator(horizons=horizons).add_features(
-        input_data
-    )
+    validated_data_with_features = TrainFeatureApplicator(
+        horizons=horizons
+    ).add_features(validated_data)
```
JanMaartenvanDoorn, line 78: Good to change this.

openstf/tasks/utils/predictionjobloop.py
```diff
@@ -60,6 +62,7 @@ def __init__(
         self.on_successful_callback = on_successful_callback
         self.on_end_callback = on_end_callback
         self.pj_kwargs = pj_kwargs
+        self.debug_pid = debug_pid
```
JanMaartenvanDoorn, line 65: Nice feature for debugging or running localy

JanMaartenvanDoorn:
I agree with the changes but I notice that some of the unit tests fail:

image

Please fix this before merging.

FrankKr:
> I agree with the changes but I notice that some of the unit tests fail:
> 
> image
> 
> Please fix this before merging.

Thanks for pointing out the failing unit test.
Turns out it really prevented me from introducing a bug; the `if is *not* None` should have been `if is None`.

JanMaartenvanDoorn (approved):
Looks good!

PR #216, opened 2021-11-02
hotfix wrong variable reference

FrankKr:
I think I accidentally introduced a bug by not pushing my last commits when fixing a bug yesterday.
This PR solves that


wfstoel (approved):
Good fix.

PR #217, opened 2021-11-02
Bug ktps 1746 horizon lag feature extract

EnricoASchmitz:
The current horizon was unused, with this fix we can and it also fixed the duplication of a lag variable 1440min=1day


charlottecvn (approved):
Looks good!

JanMaartenvanDoorn (approved):
Looks good!

PR #218, opened 2021-11-04
Change back to old filepath for trained models to fix unit test

Davidswinkels:


AlexisMignon:
Note that at least a part of the problem seems to come from the fact that the saved models are pure sklearn models not wrapped in the OpenstfRegressor base class. Another solution would be to resave a trained model based on this base class. This is what I did in the PR related to the `.feature_names` property.

Davidswinkels:
> Note that at least a part of the problem seems to come from the fact that the saved models are pure sklearn models not wrapped in the OpenstfRegressor base class. Another solution would be to resave a trained model based on this base class. This is what I did in the PR related to the `.feature_names` property.

Yes agree it would be nice to have a better solution for loading these models in the unit tests. The current unit tests where models have to be loaded sometimes can't run individually in PyCharm and can only be run from Pytest as a group. I couldn't find the PR you were referring? Do you have PR number or link?

wfstoel (approved):

Issue #184, opened 2021-10-06
[Feature] Implement linear regression model


JanMaartenvanDoorn:
WHAT: Implement a linear regression model as a openstf model.

HOW:
- Use RTE linear model and wrap in the same way as the xgb model
- Add unit tests

TEST:
- The RTE linear model is available in the pipeline
- A linear model cab be trained
- A forecast can be made using this model

PR #219, opened 2021-11-05
Feature #184 implement linear regression model

BenoitHardier:
It is the work related to #184 
 We implement a wrapper for basic linear regression that can handle missing values: 
 - The handling of missing values is performed by a metamodel that can be useful to keep aside to wrap other models that do not natively handle missing values.
 - The linear model is wrapped in a similar way to previous XGB models.
 - Unit tests have been developed to check the sanity of the model.


openstf/model/metamodels/missing_values_handler.py
```diff
@@ -0,0 +1,120 @@
...
+    Parameters
+    ----------
+
+    base_estimator: RegressorMixin
+        Regressor used in the pipeline.
+
+    missing_values: default=np.nan
+        Type of missing values handled by the imputer.
+
+    imputation_strategy: str, default=None
```
FrankKr, line 25: I guess that allowed values are defined in `SimpleImputer`?

openstf/model/metamodels/missing_values_handler.py
```diff
@@ -0,0 +1,120 @@
...
+        Imputing strategy for the imputer. If None, nan values are not allow and the linear regression will raise a ValueError.
+
+    fill_value: optional
+        Value used by the imputer for missing values.
```
FrankKr, line 29: Could you add what sensible options are? Are we looking for a specific float, or a strategy (e.g. 'mean')

openstf/model/metamodels/missing_values_handler.py
```diff
@@ -0,0 +1,120 @@
...
+                strategy=self.imputation_strategy,
+                fill_value=self.fill_value,
+            )
+            self.pipeline_ = Pipeline(
+                [("imputer", self.imputer_), ("regressor", self.regressor_)]
+            )
+
+        self.pipeline_.fit(X[self.non_null_columns_], y)
+
+        if hasattr(self.regressor_, "feature_importances_"):
```
FrankKr, line 101: Could you add a docstring to indicate what is happening here? For me it isn't directly obvious, so that would be a lot of help:)

openstf/model/metamodels/missing_values_handler.py
```diff
@@ -0,0 +1,120 @@
...
+            reg_feature_importances = self.regressor_.feature_importances_
+            self.feature_importances_ = np.zeros(self.n_features_in_)
+            j = 0
+            for i, c in enumerate(self.feature_names_):
+                if c in self.non_null_columns_:
+                    self.feature_importances_[i] = reg_feature_importances[j]
+                    j += 1
```
FrankKr, line 108: Can we make this more pythonic?
I'm not exactly sure what you want to do here, but I'm guessing that you try to only keep the feature importances of the non_null_columns_?
In that case, would this work:
`self.feature_importances_ = [x for x in self.regressor_.feature_importances if x not in self.non_null_columns_]` ?

FrankKr (commented):
Thanks for proposing this added functionality. 
For me the concept of the metamodels/missing_values_handler is not immediately clear. I've asked JM to perform a review, since he introduced the issue for which this is a fix.

I added some comments. On a sidenote; great job adding the unit tests.

Lastly, I noticed that sonarcloud complains about the variable 'X' (capital). I can see why you used it instead of a lower-case x, but my preference would be to adhere to the naming convention checked by sonar cloud. In case you didn't see it yet, the full code analysis report for your PR can be found here: https://sonarcloud.io/project/issues?id=alliander-opensource_openstf&branch=feature-%23184-Implement-linear-regression-model&resolved=false&types=CODE_SMELL

openstf/model/metamodels/missing_values_handler.py
AlexisMignon: Your proposal would not work. However a possible more pythonic way would be:
```python
reg_feature_importances_dict = dict(zip(self.non_null_columns, reg_feature_importances)
self.feature_importances_ = np.array([reg_feature_importances_dict.get(c, 0) for c in self.feature_names_])
```

openstf/model/metamodels/missing_values_handler.py
AlexisMignon: This parameter is useful when the filling stratgey is to replace by a user defined constant value. The docstring should be clearer indeed.

openstf/model/metamodels/missing_values_handler.py
```diff
@@ -0,0 +1,120 @@
...
+
+        _, y = check_X_y(X, y, force_all_finite="allow-nan", y_numeric=True)
+        if type(X) != pd.DataFrame:
+            X = pd.DataFrame(np.asarray(X))
+
+        # Remove always null columns
+        columns = X.isnull().all(0)
+        self.feature_names_ = list(X.columns)
+        self.non_null_columns_ = list(columns[~columns].index)
+        self.n_features_in_ = X.shape[1]
```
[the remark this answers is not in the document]
AlexisMignon: We introduced the meta model for missing value handling because we need the same mechanism for a future model too. The idea is that it can be used to wrap any model that does not natively deal with missing values.

There are two main fonctionalities in this model:
* Imputation of missing values
* Dealing with all-null columns so that they are not passed to the base model but everything remains consistant regarding the feature importance.

AlexisMignon:
Regarding the X variable name, it's because we followed the sklearn naming conventions. We will change it

openstf/model/metamodels/missing_values_handler.py
```diff
@@ -0,0 +1,120 @@
+# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
```
JanMaartenvanDoorn, line 1: I noticed an __init__.py is missing in the metamodels folder. Could you add it?

openstf/model/regressors/linear.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self, missing_values=np.nan, imputation_strategy=None, fill_value=0):
+        super().__init__(
+            _LinearRegressor(),
+            missing_values=missing_values,
+            imputation_strategy=imputation_strategy,
+            fill_value=fill_value,
+        )
+
+
+class LinearOpenstfRegressor(LinearRegressor, OpenstfRegressor):
```
JanMaartenvanDoorn, line 33: It took me a while to understand all the inheritance that is going on in this file. I think for now it is fine but can you think of a way simplify this in the future?

test/unit/model/regressors/test_linear.py
```diff
@@ -0,0 +1,73 @@
+# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
```
JanMaartenvanDoorn, line 1: Nice unittests and coverage!

openstf/model/metamodels/missing_values_handler.py
```diff
@@ -0,0 +1,120 @@
...
+# SPDX-License-Identifier: MPL-2.0
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator, RegressorMixin, MetaEstimatorMixin, clone
+from sklearn.pipeline import Pipeline
+from sklearn.impute import SimpleImputer
+from sklearn.utils.validation import check_is_fitted, check_X_y, check_array
+
+
+class MissingValueHandler(BaseEstimator, RegressorMixin, MetaEstimatorMixin):
```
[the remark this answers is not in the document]
JanMaartenvanDoorn: One very minor thing: This class is named MissingValueHandler while the file is name missing_value**s**_handler.py For consinteccy reasons: could you change the name of the class to MissingValuesHandler?

JanMaartenvanDoorn (commented):
Overall I think this is a nice addition to the code. For some parts it took some time for me to understand. Maybe see if you can add a bit more documentation/comments to the parts that have quite a lot going on. Especially the fit method in the MissingValueHandler and the regressors/linear.py file in general. That said I have a few additional comments/sugestions, let me know what you think. Also if you have any questions do not hesitate to reach out.

openstf/model/metamodels/missing_values_handler.py
```diff
@@ -0,0 +1,120 @@
...
+            X = pd.DataFrame(np.asarray(X))
+
+        # Remove always null columns
+        columns = X.isnull().all(0)
+        self.feature_names_ = list(X.columns)
+        self.non_null_columns_ = list(columns[~columns].index)
+        self.n_features_in_ = X.shape[1]
+
+        self.regressor_ = clone(self.base_estimator)
+        if self.imputation_strategy is None:
```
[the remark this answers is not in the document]
BenoitHardier: For us, the possibility to have no imputation is a safeguard to ensure that an error is raised if data has missing values in the default case. That is why we prefer to keep it like this.
Another possibility is to have 2 linear models one with imputation and another without (for this last model we would not wrap it with the MissingValuesHandler) and thus we could remove this default "no imputation" behaviour.

openstf/model/metamodels/missing_values_handler.py
BenoitHardier: The idea was to get feature importance from the underlying base estimator. But now, it seems to me that it should be way better to delegate this task to the LinearOpenstfRegressor thus we could remove the weird `_LinearRegressor` class and simplify the heritage.

openstf/model/metamodels/missing_values_handler.py
BenoitHardier: done.

openstf/model/metamodels/missing_values_handler.py
BenoitHardier: done.

openstf/model/metamodels/missing_values_handler.py
BenoitHardier: It has been clarified in the docstring.

openstf/model/regressors/linear.py
BenoitHardier: LinearRegressor is the sklearn-like class (wrapped in the metamodel).
LinearOpenstfRegressor inherits from this sklearn-like class and implements the OpenstRegressor interface.
The weird class _LinearRegression has been removed because feature_importance calculation has been moved in LinearOpenstfRegressor. 
It seems to me that now the inheritance is quite similar to lgbm or xgb models.

FrankKr (commented):
Thanks for incorporating the changes! Makes it more clear.
I did not have time to check the PR thoroughly, sorry! But I see you invited reviews from two other persons, so that will be fine.

openstf/model/metamodels/missing_values_handler.py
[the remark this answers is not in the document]
JanMaartenvanDoorn: Check, I dont think two different models are nescesarry. Maybe good to add some documentation that this is a double check/safe guard.

openstf/model/regressors/linear.py
JanMaartenvanDoorn: Seems much more clear, thank you for adding these clarifications!

JanMaartenvanDoorn (approved):
Looks good! Great that we have a linear model now!

jtwild:
Hi guys,

Would you be able to add an example to the repo where a linear model is trained? 
I tried to do it by taking the supplied examples of the OpenSTEF workshop and simply changing the model type to 'linear'. However, this does not work because the model throws an error about missing data. I see from your code that you have implemented a way to handle this, but I cannot understand how I should call it. From the documentaiton it is also not clear to me how I can use this.

```
ValueError: Input X contains NaN.
LinearRegression does not accept missing values encoded as NaN natively. For supervised learning, you might want to consider sklearn.ensemble.HistGradientBoostingClassifier and Regressor which accept missing values encoded as NaNs natively. Alternatively, it is possible to preprocess the data, for instance by using an imputer transformer in a pipeline or drop samples with missing values. See https://scikit-learn.org/stable/modules/impute.html You can find a list of all estimators that handle NaN values at the following page: https://scikit-learn.org/stable/modules/impute.html#estimators-that-handle-nan-values
```

PR #220, opened 2021-11-05
Feature ktps 1682 max n models

FrankKr:
After training a model, make sure that not more than MAX_N_MODELS are stored for the given prediction_job.

Added a unit test for this functionality.
The test makes use of


EnricoASchmitz (approved):
Exactly as I had in mind, looks good.

PR #227, opened 2021-11-12
Feature ktps 1775 generalize serializer

EnricoASchmitz:
Clean up of the serializer, the abstract serializers contain the only function that are directly called.
When creating a custom serializer make sure you cover all abstract functions.

For the reviewer: about half of the files are MLflow files, the unittests need a example MLflow run. I would recommend filtering the files on .py and only review the python files


EnricoASchmitz:
First merge base!
https://github.com/alliander-opensource/openstf/pull/226

PR #228, opened 2021-11-15
Bugfix ktps 1805 remove old models from disk

FrankKr:
Fixes the `serializer.remove_old_models` so artifacts are also removed from disk (and not only marked as 'deleted').


Davidswinkels (approved):
Nice fix and test!

PR #229, opened 2021-11-16
Feature custom models

BenoitHardier:
We propose to develop a mechanism to add custom models without strong coupling with the existing codebase.
The idea is to keep core models (namely already developed models like XGB) but to permit others to use training and forecasting pipelines with custom models that implement the bare minimum of the regressor interface and objective.

This would close issue #230.


openstf/model/model_creator.py
```diff
@@ -102,18 +103,22 @@ def create_model(model_type: Union[MLModelType, str], **kwargs) -> OpenstfRegres
...
             model_type = MLModelType(model_type)
+            model_class = ModelCreator.MODEL_CONSTRUCTORS[model_type]
+            valid_kwargs = valid_model_kwargs[model_type]
         except ValueError as e:
-            valid_types = [t.value for t in MLModelType]
-            raise NotImplementedError(
-                f"No constructor for '{model_type}', "
-                f"valid model_types are: {valid_types}"
-            ) from e
+            if is_custom_type(model_type):
```
Davidswinkels, line 109: This custom type will become standard behavior. So it's nicer if we can load the custom model in the try statement instead of in except statement. Then the except statement has the logic for error handling

openstf/model/objective_creator.py
```diff
@@ -34,11 +38,15 @@ def create_objective(model_type: Union[MLModelType, str]) -> RegressorObjective:
...
             # and nothing when a MLModelType enum is used.
             model_type = MLModelType(model_type)
+            objective = ObjectiveCreator.OBJECTIVES[model_type]
         except ValueError as e:
-            valid_types = [t.value for t in MLModelType]
-            raise NotImplementedError(
-                f"No objective for '{model_type}', "
-                f"valid model_types are: {valid_types}"
-            ) from e
+            if is_custom_type(model_type):
```
Davidswinkels, line 43: This custom type will become standard behavior. So it's nicer if we can load the custom objective in the try statement instead of in except statement. Then the except statement has the logic for error handling

openstf/model/regressors/custom_regressor.py
```diff
@@ -0,0 +1,69 @@
+# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+from typing import List, Type
+from abc import abstractmethod
+from importlib import import_module
+import pandas as pd
+import inspect
+from .regressor import OpenstfRegressor
```
Davidswinkels, line 10: Preferably we use full paths for package/module imports and not relative paths. It's more readable and works better

openstf/model/regressors/custom_regressor.py
```diff
@@ -0,0 +1,69 @@
...
+        ...
+
+    @property
+    @abstractmethod
+    def objective(self) -> Type[RegressorObjective]:
+        ...
+
+
+def load_custom_model(custom_model_path) -> CustomOpenstfRegressor:
+    path_elts = custom_model_path.split(".")
```
Davidswinkels, line 32: What is elts? Maybe we can use longer variable name or add comment to explain? Will make code more readable in long term

test/unit/model/test_custom_models.py
```diff
@@ -0,0 +1,74 @@
...
+    def predict(self, X, **kwargs):
+        return np.zeros(len(X))
+
+
+class InvalidRegressor:
+    pass
+
+
+class TestCustomModels(unittest.TestCase):
+    def test_create_custom_model(self):
```
Davidswinkels, line 49: Nice tests! :)

test/unit/model/test_custom_models.py
```diff
@@ -0,0 +1,74 @@
...
+            load_custom_model(model_path)
+
+        with self.assertRaises(ValueError):
+            model_path = __name__ + ".InvalidRegressor"
+            load_custom_model(model_path)
+
+    def test_create_custom_objective(self):
+        model = DummyRegressor()
+        model_path = __name__ + ".DummyRegressor"
+        obj = create_custom_objective(model, None)
```
Davidswinkels, line 68: maybe use objective instead of obj. Now it's still clear in this code, but if we expand these tests it might become unclear later. Therefore just call it objective already :)

Davidswinkels (approved):
Some small changes that can be fixed, but nice addition to make a start for custom regressors!

openstf/model/regressors/custom_regressor.py
Davidswinkels: Same for row below (line 11)

Davidswinkels:
Thanks for the extra fixes to make code nicer. If merge conflicts are resolved, this can go to main :)

PR #237, opened 2021-11-22
fixed bug in getting model age

JanMaartenvanDoorn:
- fixed bug that prevented hyperparameter optimisation to start

### DIFF
--- a/openstf/model/serializer.py
+++ b/openstf/model/serializer.py
@@ -140,8 +140,12 @@ class MLflowSerializer(AbstractSerializer):
             # return the latest run
             prev_run = self._find_models(pj["id"], n=1)
             # Use [0] to only get latest run id
-            prev_run_id = str(prev_run["run_id"])
-        except LookupError:
+            if not prev_run.empty:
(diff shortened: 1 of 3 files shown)


openstf/model/serializer.py
```diff
@@ -140,8 +140,12 @@ def save_model(
...
             prev_run = self._find_models(pj["id"], n=1)
             # Use [0] to only get latest run id
-            prev_run_id = str(prev_run["run_id"])
-        except LookupError:
+            if not prev_run.empty:
+                prev_run_id = prev_run["run_id"][0]
+            else:
+                self.logger.info("No previous model found in MLflow", pid=pj["id"])
+                prev_run_id = None
+        except KeyError:
```
wfstoel, line 148: This seems like it does the same thing twice. Shouldn't it be either the if-statement or the try-except, instead of both?

JanMaartenvanDoorn: good one, fixed it now

wfstoel (approved):
Good fix

Issue #187, opened 2021-10-06
[Feature] Review/give feedback to FeatureAdder implementation


JanMaartenvanDoorn:
WHAT: Now features can only be added in groups. A feature adder can help making a more specific selection of the features we want to add. This can also help to add feature that are not yet available in openstf. 

HOW:
- @BenoitHardier opens a PR
- Lets have a discussion on whats is the best way to parameterize this functionality

TEST:
- PR approved or clear plan for further implemmentation

JanMaartenvanDoorn:
I think this one can be closed as well

PR #246, opened 2021-11-24
Feature #187 add feature adder

BenoitHardier:
We propose to develop a mechanism to add custom features without strong coupling with the existing codebase.
The mechanism relies on 2 classes:

- _FeatureAdder_ is the basic block that implements the logic for computing the specific feature and the syntactic sugar to load properly the FeatureAdder according to the feature name. 

- _FeatureDispatcher_ regroups a list of FeatureAdders, scans the feature_names to assign to each feature the proper FeatureAdder and orchestrates the effective computing of the features.

Each new feature (or group of features) you want to add could be wrapped in a FeatureAdder. The parameters of a feature can be incorporated in the name of that feature and retrieved by the FeatureAdder with a regex. 

---
Example: dayofweek_(?P\<type\>num|name) 
- if the feature is 'dayofweek_num' -> output day of the week number -> 4
- if the feature is 'dayofweek_name' -> output day of the week name -> Friday
---

For now, the mechanism is implemented but not added in the pipelines.
The idea is to add the FeatureDispatcher in FeatureApplicator classes. 
**However, we think that we need a way to indicate from where to load custom featureAdders (probably we need to point to the file(s) where to find FeatureAdders and load all these classes)**

This would close issue #187.


FrankKr:
I think I fixed all issues caused by the renaming of openstf to openstef. 
The build does not succeed, but this was already the case before the renaming

JanMaartenvanDoorn (approved):
Nice new feature for openSTEF! I thinkwe should consider moving all features to this system.
For a few functions typehints are missing. Please add them before you merge.

PR #247, opened 2021-11-25
Improve find models function in mlflow serializer

Davidswinkels:


openstf/model/serializer.py
```diff
@@ -279,39 +281,39 @@ def _find_models(
         pid: Union[int, str],
         n: Optional[int] = None,
         filter_string: str = None,
-    ) -> Union[pd.Series, pd.DataFrame]:
+    ) -> Union[pd.DataFrame]:
```
JanMaartenvanDoorn, line 284: This does not have to be an union anymore

JanMaartenvanDoorn (approved):
Nice fix!

PR #250, opened 2021-11-25
Feature ktps 1706 implement proloaf model

JanMaartenvanDoorn:


openstf/feature_engineering/feature_applicator.py
```diff
@@ -34,17 +35,22 @@ def __init__(
...
             df: pd.DataFrame with input data to which the features have to be added
+            pj (PredictionJobDataClass): Prediction job.
         """
         pass
 
 
 class TrainFeatureApplicator(AbstractFeatureApplicator):
-    def add_features(self, df: pd.DataFrame, latency_config=None) -> pd.DataFrame:
+    def add_features(
+        self, df: pd.DataFrame, pj: PredictionJobDataClass = None, latency_config=None
```
JanMaartenvanDoorn, line 52: I think we should not make pj optional here

openstf/feature_engineering/feature_applicator.py
```diff
@@ -97,8 +107,11 @@ def add_features(self, df: pd.DataFrame, latency_config=None) -> pd.DataFrame:
 
         # NOTE this is required since apply_features could add additional features
         if self.feature_names is not None:
-            # Add horizon to requested features else it is removed
-            features = self.feature_names + ["horizon"]
+            # Add horizon to requested features else it is removed, and if needed the proloaf feature (historic_load)
+            if pj is not None and pj["model"] == "proloaf":
```
JanMaartenvanDoorn, line 111: if pj is not optional you dont have to check wheter it is none. Besides that I dont think this check will work in case pj is none as the evaluation of pj["model"] == "proloaf" will raise an exception

openstf/feature_engineering/proloaf_historic_features.py
```diff
@@ -0,0 +1,37 @@
+# SPDX-FileCopyrightText: 2017-2021 Alliander N.V. <korte.termijn.prognoses@alliander.com> # noqa E501>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+""" This module contains all proloaf related functions used for feature engineering.
```
JanMaartenvanDoorn, line 5: I think it is better remove proloaf from the name of this file as other models might need the same features in the future. However it is good to mention in the documentation that this feature is intended for models that generate their own lag features such as proloaf

openstf/pipeline/optimize_hyperparameters.py
```diff
@@ -81,7 +81,18 @@ def optimize_hyperparameters_pipeline(
 
     validated_data_with_features = TrainFeatureApplicator(
         horizons=horizons
-    ).add_features(validated_data)
+    ).add_features(validated_data, pj=pj)
+
+    if pj["model"] == "proloaf":
+        # Adds additional proloaf features to the input data, historic_load (equal to the load)
+        if "historic_load" not in list(validated_data_with_features.columns):
+            validated_data_with_features[
```
JanMaartenvanDoorn, line 89: Sonarcloud complains about this part, it is probably better to replace this double if statement with:

if pj["model"] == "proloaf" and "historic_load" not in list(validated_data_with_features.columns):

charlottecvn (approved):

PR #256, opened 2021-11-27
Hotfix confidence interval applicator stdev

Davidswinkels:
Original fix for confidence interval applicator:
- Add tests to confidence_interval_applicator
- Improve catch of lookuperror when finding old model for training a model
- Raise error when standard deviation is empty

Ran into issues when running other unit tests. Therefore also fixed:
- formatting code with black and isort
- Cleaning up old data and trained_model runs
- reading trained_models and utils from unit test directory instead of test directory. Work directory is base openstf directory (this is how pytest runs it in ci/cd pipeline. In PyCharm we can set the working directory to use same work directory)

image


openstf/pipeline/train_model.py
```diff
@@ -62,7 +62,7 @@ def train_model_pipeline(
         old_model_age = (
             old_model.age
         )  # Age attribute is openstf specific and is added by the serializer
-    except (AttributeError, FileNotFoundError):
+    except (AttributeError, FileNotFoundError, LookupError):
```
FrankKr, line 65: Good fix!

openstf/pipeline/train_model.py
```diff
@@ -119,17 +119,6 @@ def train_model_pipeline_core(
...
-    required anymore.
-
-    For training a model the following keys in the prediction job are
-    expected:
-        "id"            Only used for logging
-        "model"         Model type, any of "xgb", "lgb",
-    And in modelspecs:
-        "hyper_params"  Hyper parameters dictionairy specific to the model_type
-        "feature_names"      List of features to train model on or None to use all features
-
```
FrankKr, line 132: Good cleanup

test/unit/model/test_confidence_interval_applicator.py
```diff
@@ -11,20 +12,86 @@
...
+                1.1089453301,
+                2.9508361224,
+                1.6705985526,
+                1.9371298645,
+                3.2265538352,
+            ],
+            "hour": [0, 1, 2, 3, 4],
+            "horizon": [47, 47, 47, 47, 47],
+        }
+    )
```
FrankKr, line 27: I don't see how this is needed in the unit test. A small docstring on why this is required would help the next person who looks at it

test/unit/model/test_confidence_interval_applicator.py
```diff
@@ -11,20 +12,86 @@
...
+            pd.Timestamp(2012, 5, 1, 2, 00),
+        ]
+        stdev_forecast = self.stdev_forecast
+        actual_stdev_forecast = ConfidenceIntervalApplicator(
+            MockModel(), stdev_forecast
+        )._add_standard_deviation_to_forecast(forecast)
+        self.assertTrue("stdev" in actual_stdev_forecast.columns)
+        self.assertEqual(actual_stdev_forecast["stdev"][0], 2.9508361224)
+        self.assertEqual(actual_stdev_forecast["stdev"][1], 2.9508361224)
+        self.assertEqual(actual_stdev_forecast["stdev"][2], 1.6705985526)
```
FrankKr, line 89: For next time, you could use self.assertDataFrameEqual

FrankKr (approved):

test/unit/model/test_confidence_interval_applicator.py
Davidswinkels: Yes good suggestion. Added a small docstring.

Issue #244, opened 2021-11-24
[Feature] Implement grouped regressor meta model


JanMaartenvanDoorn:
WHAT: We want to have the posibility to group data by a certain column in the input data and train a model on each group. This can be for example different substations, hour of the day, day of the week. 

HOW:
- make a sklearn metamodel (similar to the missing values model with linear regression)
- check how this works with hyper parameter optimization
- make new hypeparameter entry for the grouping columns

TEST:
- model can be trained on multiple groups of data

PR #259, opened 2021-11-30
Feature #244 grouped regressor metamodel

BenoitHardier:
We propose to develop a metamodel that can group data by somes columns in the input data and train a model on each group.
We add unit tests with the linear model wrapped in a GroupedRegressor.
The possible issues about hyperparameters tuning should probably be resolved in the custom model definition using a GroupedRegressor.

This would close issue #244.


openstef/model/metamodels/grouped_regressor.py
```diff
@@ -0,0 +1,141 @@
...
+from sklearn.utils.validation import check_is_fitted
+
+
+class GroupedRegressor(BaseEstimator, RegressorMixin, MetaEstimatorMixin):
+    """Meta-model that trains an instance of the base estimator for each key of a groupby operation applied on the data.
+    The base estimator is a sklearn regressor, the groupby is performed on the columns specified in parameters.
+    Moreover fit and predict methods can be performed in parallel for each group key thanks to joblib.
+    Example:
+
+    data = | index | group | x0 | x1 | x3 | y |
```
JanMaartenvanDoorn, line 16: Nice documentation!

openstef/model/metamodels/grouped_regressor.py
```diff
@@ -0,0 +1,141 @@
...
+    ----------
+        feature_names_: list(str)
+            All input feature (without group_columns).
+
+        estimators_: dict(str, RegressorMixin)
+            Dictionnary that stocks fitted estimators for each group.
+            The keys are the keys of grouping and the values are the regressors fitted on the grouped data.
+    """
+
+    def __init__(self, base_estimator, group_columns, n_jobs=1):
```
JanMaartenvanDoorn, line 51: A general remark: I see the methods do not have typehints (both input and output), could you please add these to all mehtods?

openstef/model/metamodels/grouped_regressor.py
```diff
@@ -0,0 +1,141 @@
...
+    @classmethod
+    def grouped_compute(cls, df, group_columns, func, n_jobs=1):
+        index_name = df.index.name or "index"
+        df_reset = df.reset_index()
+
+        df_res = pd.DataFrame(index=df_reset.index)
+
+        gb = df_reset.groupby(group_columns)
+
+        if n_jobs > 1:
```
JanMaartenvanDoorn, line 91: Nice to make this parallel, do you think it is a good idea to also implement this in de xgb_quantile model?

openstef/model/metamodels/grouped_regressor.py
```diff
@@ -0,0 +1,141 @@
...
+        X = df_group.loc[:, self.feature_names_]
+        y = df_group.loc[:, "__target__"]
+        return (group, estimator.fit(X, y))
+
+    def _partial_predict(self, group, df_group):
+        return self.estimators_[group].predict(df_group)
+
+    @classmethod
+    def grouped_compute(cls, df, group_columns, func, n_jobs=1):
+        index_name = df.index.name or "index"
```
JanMaartenvanDoorn, line 84: This function seems to do quite a lot, could you add a bit more documentation on what it is doing and how the DataFrames look during every step?

JanMaartenvanDoorn (commented):
Nice addition to openSTEF! In general I like your implementation. I have two general remarks:
- Could you add type hints to the methods?
- Could you add a bit more documentation to the methods in the class, especially when converting the DataFrames in different formats?

Davidswinkels:
We at Alliander can help to look into sonarcloud, signing and some tests

Davidswinkels:
@BenoitHardier black is fixed, and merged latest main branch into this branch. Unit tests for grouped regressor should still be added to overcome sonarcloud code analysis and the comments from JM still need to be addressed. It would be great if you can fix those so we can review this branch and merge it to main :)

BenoitHardier:
> @BenoitHardier black is fixed, and merged latest main branch into this branch. Unit tests for grouped regressor should still be added to overcome sonarcloud code analysis and the comments from JM still need to be addressed. It would be great if you can fix those so we can review this branch and merge it to main :)

Hi @Davidswinkels, I've already written unit test here  https://github.com/OpenSTEF/openstef/blob/7c0617860d151935497c6674ca36c20953e2d703/test/unit/model/metamodels/test_grouped_regressor.py#L13 
So I am puzzled by the result of the coverage analysis. Could it be possible that the sonarcloud skip this test file?

openstef/model/metamodels/grouped_regressor.py
BenoitHardier: Yes we think that the xgb_quantile could be implemented with the grouped_regressor. However if the current implementation is working properly we suggest that it stays like that until you have time to test the switch. (If we have time we can propose to you an implementation but it is not in our priorities)

Davidswinkels:
Hmm very strange that there is unit tests, but that they don't get used for the code coverage. @BenoitHardier You could try to run pytest with a coverage check locally to see if the unit tests are included locally. If locally coverage does not work too, then it could be possible to switch to pytest instead of unittest, and then recheck if tests are included in coverage. Or potentially it could be a setting in the CICD (github action) that your folder is not included. I will check the CICD pipeline for that

Davidswinkels:
CICD pipeline does include test/unit for pytest coverage: https://github.com/OpenSTEF/openstef/blob/7c0617860d151935497c6674ca36c20953e2d703/.github/workflows/python-build.yaml#L53-L61

BenoitHardier:
```
        <package name="model.metamodels" line-rate="0.9912" branch-rate="0" complexity="0">
			<classes>
				<class name="__init__.py" filename="model/metamodels/__init__.py" complexity="0" line-rate="1" branch-rate="0">
					<methods/>
					<lines/>
				</class>
				<class name="grouped_regressor.py" filename="model/metamodels/grouped_regressor.py" complexity="0" line-rate="0.9857" branch-rate="0">
```

The pytest coverage check in local seems to detect my unit tests the coverage `line-rate` is high.

BenoitHardier:
@Davidswinkels Is it possible that the coverage.xml file from the component test overwrite the coverage file from the unit test ?

Davidswinkels:
Pytest locally shows 99% coverage in grouped_regressor
openstef\model\metamodels\grouped_regressor.py           70      1    99%

Still trying to figure out why sonarcloud does not see this coverage

Davidswinkels:
@BenoitHardier The sonar cloud issue seems fixed. You can try to resolve the comments from JM. Then I can review and approve :) So we can get this branch merged to main

openstef/model/metamodels/grouped_regressor.py
Davidswinkels: Great that we could later switch to use grouped_regressor too

openstef/model/metamodels/grouped_regressor.py
```diff
@@ -0,0 +1,214 @@
...
+
+        index_name = df.index.name or "index"
+        df_reset = df.reset_index()
+
+        df_res = pd.DataFrame(index=df_reset.index)
+
+        gb = df_reset.groupby(group_columns)
+
+        if n_jobs > 1:
+            group_res = joblib.Parallel(n_jobs=n_jobs)(
```
Davidswinkels, line 149: Preferably parallel scaling is done at a higher level (e.g. cluster level with unlimited multiple pods/jobs that run openstef jobs simultaneously in a kubernetes cluster) instead of at process level, that is often limited by amount of available CPU processors. Scaling at both levels (cluster + process) becomes complex too since it is unclear what is scaled when. This function is okay now, but let's add a comment that preferred parallel scaling is at cluster level instead of process level
```suggestion
            # Preferred scaling is at cluster level (e.g. k8s/serverless) instead of process level
            group_res = joblib.Parallel(n_jobs=n_jobs)(
```

test/unit/model/metamodels/test_grouped_regressor.py
```diff
@@ -0,0 +1,92 @@
...
+import sklearn
+from lightgbm import LGBMRegressor
+
+from openstef.model.regressors.linear import LinearRegressor
+from openstef.model.metamodels.grouped_regressor import GroupedRegressor
+from test.unit.utils.base import BaseTestCase
+from test.unit.utils.data import TestData
+
+
+class TestGroupedRegressor(BaseTestCase):
```
Davidswinkels, line 13: Nice unit tests with good code coverage!

Davidswinkels (changes requested):
Hi @BenoitHardier. Thanks for adding GroupedRegressor. I think this branch is nearly ready to merge to main. There is one last comment about scaling that needs to be resolved.

openstef/model/metamodels/grouped_regressor.py
BenoitHardier: You are right, we could even log a warning to be sure that it is at the risk of the user. However my idea is that we can keep it, if someone use it out of the box the parallel scaling could be useful.

BenoitHardier (commented):
Everything all right for me !
@AlexisMignon  has the same issue with Black and the coverage in #333. Do we wait for this PR to be merged in main and then merge main into #333 ?

Davidswinkels:
> Everything all right for me ! @AlexisMignon has the same issue with Black and the coverage in #333. Do we wait for this PR to be merged in main and then merge main into #333 ?

Great. Yeah we can merge this first and then marge main into #333

Davidswinkels (approved):
Good feature to finally have 🏆

PR #264, opened 2021-12-08
Replaced hardcoded horizon by the time resolution

### LINKED ISSUE #206: [BUG] Hardcoded horizon for the operational feature applicator in create forecast pipeline
**Describe the bug**
You need to pass a list of **one** horizons to `OperationalPredictFeatureApplicator`.
https://github.com/alliander-opensource/openstf/blob/3a40da41c55acb84556fb0ae79462b339f7a8a2f/openstf/feature_engineering/feature_applicator.py#L116
Thus, in the forecast pipeline, this list of horizons is hardcoded with the value `0.25` as 15min seems to be the default time resolution in openstf.
This horizon is used in the downstream function `extract_lag_features` to filter impossible lag features for the forecast.
https://github.com/alliander-opensource/openstf/blob/3a40da41c55acb84556fb0ae79462b339f7a8a2f/openstf/feature_engineering/lag_features.py#L94
If the resolution used is lower than 15min, namely 10 min, this filtering will be incorrect.

**Expected behavior**
It should accept arbitrary _horizon_ coming from the prediction job and the default value shoud be `horizon=None` where no filtering is performed on lag features.

AlexisMignon:
Replaced the hard coded 0.25 horizon by the resolution information in the prediction job object.

Close #206

### DIFF
--- a/openstef/pipeline/create_forecast.py
+++ b/openstef/pipeline/create_forecast.py
@@ -83,7 +83,7 @@ def create_forecast_pipeline_core(
     # Add features
     data_with_features = OperationalPredictFeatureApplicator(
         # TODO use saved feature_names (should be saved while training the model)
-        horizons=[pj["resolution_minutes"] / 60.],
+        horizons=[pj["resolution_minutes"] / 60.0],
         feature_names=model.feature_names,
     ).add_features(validated_data)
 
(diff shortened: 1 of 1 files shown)


Davidswinkels (approved):
Good fix! :)

Davidswinkels:
> Note that there are some hardcoded default horizons in the TrainFeatureApplicator.add_features methods.
> Is it really a good idea?

Good idea. Maybe we can change it at other places as well?

PR #267, opened 2021-12-08
Bugfix ktps 1857 loading models

FrankKr:
fix bug in loading of models (currently alliander acc)
also, make a nice function to separate the generation of model file, so this can be mocked in unit tests

### DIFF
--- a/openstef/model/serializer.py
+++ b/openstef/model/serializer.py
@@ -190,11 +190,8 @@ class MLflowSerializer(AbstractSerializer):
                     f"Model couldn't be found for pid {pid}. First train a model!"
                 )
 
-            # Fix artifact_uri to also work when unit testing remote
-            model_uri_orig = os.path.join(latest_run.artifact_uri, "model/")
-            # replace first part so locally loaded models also work remote
-            model_uri = f"{self.mlflow_folder}{model_uri_orig.split('test/unit/trained_models/mlruns')[-1]}"
-
+            # Get model uri
+            model_uri = _get_model_uri(latest_run.artifact_uri)
             loaded_model = mlflow.sklearn.load_model(model_uri)
 
             # Add model age to model object
@@ -536,3 +533,10 @@ class MLflowSerializer(AbstractSerializer):
                     pid=pid,
                 )
         return modelspecs.feature_names
+
+
+def _get_model_uri(artifact_uri: str) -> str:
+    """Set model uri based on latest run.
+    Note that this function is primarily useful
+    so it can be mocked during unit tests"""
+    return os.path.join(artifact_uri, "model/")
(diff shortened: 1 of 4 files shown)


openstef/model/serializer.py
```diff
@@ -536,3 +531,10 @@ def _get_feature_names(
                     pid=pid,
                 )
         return modelspecs.feature_names
+
+
+def _get_model_uri(artifact_uri: str) -> str:
+    """Set model uri based on latest run.
+    Note that this function is primarily useful
+    so it can be mocked during unit tests"""
```
Davidswinkels, line 539: Nice to clean up how we get the path to the model! If we only use this `get_model_uri` function for mocking, then I think we could also mock os.path.join and lose the extra function for unit tests potentially. Will be a bit cleaner

Davidswinkels (commented):
Nice fix! We already had quite some issues loading the model from the path before. It would still be nice if we can add a unit test, where we check what loaded_model.path is set if we have a mocked lastest_run.artifact_uri and run `load_model` with the expected correct paths. Hopefully that will help with breaking issues with mlflowserializer in the future

openstef/model/serializer.py
FrankKr: Good suggestion. I've kept it as-is for now, I like it better as a separate function, but I guess that's just taste:)

FrankKr:
> Nice fix! We already had quite some issues loading the model from the path before. It would still be nice if we can add a unit test, where we check what loaded_model.path is set if we have a mocked lastest_run.artifact_uri and run `load_model` with the expected correct paths. Hopefully that will help with breaking issues with mlflowserializer in the future

Good idea, I've added a test for the mocked model path explicitely

test/unit/tasks/test_create_forecast.py
```diff
@@ -14,14 +15,28 @@
...
+        # Determine absolute location where already stored model is, based on relative path.
+        # This is needed so the model stored in the repo can be found when running remote
+        rel_path = "test/unit/trained_models/mlruns/0/d7719d5d316d4416a947e4f7ea7e73a8/artifacts/model"
+        _get_model_uri_mock.return_value = Path(rel_path).absolute().as_uri()
         # Use MLflowSerializer to load a model
         self.model, _ = self.serializer.load_model(pid=307)
 
+    def test_mocked_model_path(self):
+        """This test explicitely tests if the model path is mocked correctly"""
+        assert (
```
Davidswinkels, line 35: Good extra test. It would be even better to test if the path is correctly created in the `MLflowSerializer.load_model` function without mocking all the path creation functions. We then would mock `mlflow.sklearn.load_model`, `self._determine_model_age_from_mlflow_run` and `self._get_model_specs`. Then we can really check if whole path is correctly calculated and set by the `MLflowSerializer.load_model` function

Davidswinkels (approved):
Nice extra test. Now we test the mocked path directly, but it would be better to test the calculated path and mock all other function's that don't calculate the model path.

test/unit/tasks/test_create_forecast.py
FrankKr: Good idea, I added the test

Issue #268, opened 2021-12-08
Cannot train model with low number of training samples


wfstoel:
The first two weeks of training data get removed in [this line](https://github.com/OpenSTEF/openstef/blob/78f8ba332370cf132ab2acc2dd192afd88ff3492/openstef/validation/validation.py#L73). This makes it impossible to train a new model with low number of training samples.

What is the reasoning behind this line? And should we remove it or not?

FrankKr:
Reasoning was that for many load-time-series the T-7d and T-14d features are A: important, but B: not available for the first 7 (respectively 14) days. 
But perhaps good idea to not remove this by default, or only remove them if plenty of inputdata is available?

wfstoel:
I agree that the T-7d and T-14d features are important. However, this removes the first two weeks from the dataset **before** calculating these features. This functionality probably changed during development. For now I propose to remove this line. If desired we could add removing first two weeks of data **after** calculating the laggs in the feature adding process.

PR #269, opened 2021-12-09
No longer remove first two weeks from dataset

wfstoel:
See #268. Also slightly improved logging of validation functions.


FrankKr (approved):
Looks good!

PR #270, opened 2021-12-14
Bugfix train quantile models

FrankKr:
As discussed over the phone.
Fixes bug we currently have on prd, which in rare occasions can occur.

Added a unit test to test this bug as well.

Also fixes KTPS-1865


test/unit/pipeline/test_optimize_hyperparameters.py
```diff
@@ -23,10 +23,20 @@ def setUp(self) -> None:
 
     @patch("openstef.model.serializer.MLflowSerializer.save_model")
     def test_optimize_hyperparameters_pipeline(self, save_model_mock):
+        """Also check if non-default quantiles are processed correctly"""
```
JanMaartenvanDoorn, line 26: Nice test!

JanMaartenvanDoorn:
Maybe good idea to also increase the version number in setup.py for a new release

PR #281, opened 2022-01-06
fixed unit tests due to bevrijdingsdag issue

JanMaartenvanDoorn:


test/unit/feature_engineering/test_apply_features.py
```diff
@@ -105,7 +106,7 @@ def test_train_feature_applicator(self):
 
         self.assertDataframeEqual(
             input_data_with_features,
-            TestData.load("input_data_multi_horizon_features.csv"),
+            TestData.load("input_data_multi_horizon_features.csv").drop(columns=['is_bevrijdingsdag', 'is_bridgedaybevrijdingsdag']),
```
FrankKr, line 109: It looks a little 'hacky' to drop these columns here.
Would it make more sense to change the 'apply_holidays_features' so the output columns are clearly defined?

FrankKr, line 109: It looks a little 'hacky' to drop these columns here.
Would it make more sense to change the 'apply_holidays_features' so the output columns are clearly defined?

Also, will this unit test bug pop-up again when we are in a year where 'bevrijdingsdag' is not an additional holiday?

FrankKr:
Also, I noticed the 'black-format-code' action returns an error. This seems unrelated.
I did a quick google to see if I could fix this, but no luck.
However, we do not have to deal with that here I suppose.

test/unit/feature_engineering/test_apply_features.py
JanMaartenvanDoorn: True, I will update the csv.

I checked and I think the holidays package changed some things as Bevrijdingsdag is not returned anymore for any year. I think this will not cause an issue anymore.

JanMaartenvanDoorn:
> Also, I noticed the 'black-format-code' action returns an error. This seems unrelated. I did a quick google to see if I could fix this, but no luck. However, we do not have to deal with that here I suppose.

hopefully this is fixed now as well

FrankKr (approved):
ok, great!

PR #284, opened 2022-01-12
Feature/1749 setup mlflow for distributed mlflow tracking server

Davidswinkels:
This feature makes it possible to connect to an optional MLFlow server uri:
- Adding optional mlflow tracking uri from environment variable, which makes it possible to connect to distributed mlflow tracking server
- extra: Removed abstractserializer for now since we don't have multiple serializers
- extra: Fixed small bug by always setting experiment id and tracking uri in init

There were some nice to haves to make mlflow serializer implementation even cleaner without relying too much on prediction job, but I'm not sure if we want to pick that up in this user story. That's a nice optimization for later if everything works.


wfstoel (approved):
Nice changes to introduce MLFlow with S3 buckets! Instead of using an environment variable, it would be a bit cleaner to use an argument in the __init__ function.

openstef/model/serializer.py
```diff
@@ -81,11 +86,13 @@ def save_model(
             self._log_figure_with_mlflow(report)
         self.logger.debug(f"MLflow path after saving= {self.mlflow_folder}")
 
-        # Also store report files in easy-to-find location, which are updated on each new model
-        # Easy for web visualisation, e.g. through grafana
-        location = os.path.join(self.web_volume, f'{pj["id"]}')
-        Reporter.write_report_to_disk(report, location=location)
-        self.logger.info(f"Stored report to disk: {location}")
+        if save_report_to_disk:
```
wfstoel, line 89: Good idea to make this an option.

wfstoel (approved):

openstef/model/serializer.py
Davidswinkels: Yes that makes it possible to not use a persistent volume

PR #303, opened 2022-02-14
Update create_basecase_forecast.py

FrankKr:
Signed-off-by: FrankKr <fpmkreuwel@gmail.com>

### DIFF
--- a/openstef/pipeline/create_basecase_forecast.py
+++ b/openstef/pipeline/create_basecase_forecast.py
@@ -5,6 +5,7 @@ from pathlib import Path
 
 import pandas as pd
 import structlog
+import numpy as np
 from openstef_dbc.services.prediction_job import PredictionJobDataClass
 
 from openstef.feature_engineering.feature_applicator import (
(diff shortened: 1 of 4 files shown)


openstef/pipeline/create_basecase_forecast.py
```diff
@@ -41,10 +42,10 @@ def create_basecase_forecast_pipeline(
     logger = structlog.get_logger(__name__)
 
     logger.info("Preprocessing data for basecase forecast")
-    # Validate and clean data - use a very long flatliner threshold.
-    # If a measurement was constant for a long period, a basecase should still be made.
+    # Validate data
+    # Currently effectively disabled by giving np.inf
     validated_data = validation.validate(
-        pj["id"], input_data, flatliner_threshold=4 * 24 * 14 + 1
+        pj["id"], input_data, flatliner_threshold=np.inf
```
Davidswinkels, line 48: Just wondering why we don't remove the validate method? What are reasons to keep it with np.inf?

FrankKr: I thought about that as well. I kept it in so we explicitly show that we did not 'forget' this step, and so that in the future we can more easily switch to arguments defined in the prediction_job (KTPS-1963)

FrankKr:
I'll fix the failing unit tests

openstef/pipeline/create_basecase_forecast.py
Davidswinkels: Ok good plan and good reasons :)

FrankKr:
Test coverage is measured inaccurately.

Issue #294, opened 2022-01-26
Let model classes declare whether they can deal with quantiles predictions themselves


AlexisMignon:
WHAT:  Currently, there are basically two quantiles mechanism:
  1. If the model name contains "quantile" then the model is called using the XGBoost API
  2. Otherwise, quantiles are computed separately.

Rather to rely on the class name, we move the decision regarding the quantile mechanism to use in an abstract method declared in the base class of OpenSTEF models.

HOW:
- add property to the openSTEF baseclass that indicated whether a model can provide its own quantile predictions
- Implement a check for this property in the train and predict pipelines.
- remove checks on the name of the modelype

TEST:
- Models can indicate wheter they can provide quantile predictions.
- Pipelines ask for quantile predictions based on this property

PR #305, opened 2022-02-22
Add the property indicating if a model can predict quantiles

BenoitHardier:
We propose to add a property that indicates if the model can provide a prediction for the quantiles.
This property replaces the check on the name of the model. Indeed the current behavior checks if "quantile" is in the name of the model to choose between the default compute of the quantiles or the use the model's predictions.

This would close issue #294 
Signed-off-by: Benoit Hardier <benoit.hardier@probayes.com>

### DIFF
--- a/openstef/model/basecase.py
+++ b/openstef/model/basecase.py
@@ -70,3 +70,7 @@ class BaseCaseModel(BaseEstimator, RegressorMixin):
         ]
 
         return basecase_forecast.sort_index()
+
+    @property
+    def can_predict_quantiles(self):
+        return False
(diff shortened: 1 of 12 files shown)


FrankKr (changes requested):
Hi Benoît, Nice addition to make this explicit.
Sonar complains the test-coverage is low; I've made a PR to increase this:
https://github.com/OpenSTEF/openstef/pull/309
Could you have a look and merge that PR into this one to fix the issue?

FrankKr (approved):
I see that the coverage check still complains, however, I approve the PR for merge; I'm confident the code is tested, and the numer of ' uncovered lines'  is very small.

PR #312, opened 2022-03-14
Update requirements and fix unit tests

wfstoel:
Updates some of the requirements to the newest version. Also removes some unnecessary requirements.

The core requirements (xgboost, sklearn) are not updated as to preserve current default behavior. Updating these might be a good idea as well, but requires more work to make sure performance is not affected.

Signed-off-by: Willem Frederik Stoel <frederik.stoel@alliander.com>


FrankKr (approved):
Looks good!

PR #313, opened 2022-03-17
Bug update default xgb quantile hyperparams

JanMaartenvanDoorn:
- Added default hyper parameters to xgb_quantile that are more in line with standard xgb hyperparameters


wfstoel (approved):
Good!

PR #319, opened 2022-03-18
WIP: Added possibility to save the forecasts on the train datasets

### LINKED ISSUE #316: Possibility to save the forecasts on the datasets used in training.
WHAT: When chaining forecast models, we need to be able to retrieve the forecasts of previous models on the train/val/test sets.

HOW:
* Add an option `save_train_forecasts` in the `PredictionJobDataClass`

There are at least two possibilities to implement retrieve the forecasts:

1. Everything in the report :
  * Return the forecasts through the report object by adding in to the stored data frames.
  * Make the entry point of the pipeline return the report so we can retrieve the forecasts.

2. Return model and report
* Both are returned by the pipeline entry point
* we recompute the forecasts in the task

3. Use report to transfer the forecasts
* As point 1. but we remove it from report at the end and just return the forecast

How to store the forecasts:

1. Database call with a specific method (what would be the proper signature)
2. Keep it in memory for instance with a dictionnary indexed by the pj id (might lead memory issues).

TEST: Check that the forecasts are correctly returned when required.

AlexisMignon:
Closes #316 

### DIFF
--- a/openstef/data_classes/prediction_job.py
+++ b/openstef/data_classes/prediction_job.py
@@ -19,6 +19,7 @@ class PredictionJobDataClass(BaseModel):
     name: str
     description: Optional[str]
     quantiles: Optional[List[float]]
+    save_train_forecasts: False
 
     def __getitem__(self, item):
         """Allows us to use subscription to get the items from the object"""
--- a/openstef/data_classes/prediction_job.py
+++ b/openstef/data_classes/prediction_job.py
@@ -19,7 +19,7 @@ class PredictionJobDataClass(BaseModel):
     name: str
     description: Optional[str]
     quantiles: Optional[List[float]]
-    save_train_forecasts: False
+    save_train_forecasts: bool = False
 
     def __getitem__(self, item):
         """Allows us to use subscription to get the items from the object"""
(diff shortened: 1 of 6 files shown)


JanMaartenvanDoorn:
I havent looked thoroughly to the PR but I noticed some unit tests are failing: 
image
Could you please fix them first?

BenoitHardier:
I fix the bug but now I have a problem on the SonarCloud Scan stage of the build pipeline. Could you check-it @JanMaartenvanDoorn ? (I have the same problem on the #320) 

> ERROR: Caused by: Error 500 on https://sonarcloud.io/api/alm_integration/show_pullrequest?project=OpenSTEF_openstef&pullrequestKey=320 : {"errors":[{"msg":"An unexpected error occurred. Please try again later."}]}  

And the link outputs this:
>  {"errors":[{"msg":"Authentication is required"}]}

AlexisMignon:
@JanMaartenvanDoorn. The raised error is related to the test of the train_test_split function, which is unrelated to our modifications, or at least that's what we assume. Could you have a look ?

JanMaartenvanDoorn:
I think I fixed the test. The default split functions uses stratification as the test was testing the normal splitting. I now changed the tests such that it is testing the regular plitting. The stritification is tests in the next test.

JanMaartenvanDoorn (approved):
Looks good! Black is failing but that is not caused by the code formatting

PR #322, opened 2022-03-18
Same end time for all PJs of a task

### LINKED ISSUE #318: Fix the end time for the train task to be the same for all prediction jobs.
WHAT: In the current implementation, the end time of the input data given to the train pipeline is computed just before calling it. If the pipelines take a lot of time to complete, then this time can be different for all the pipelines execution. In order to keep consistent datasets across pipelines (for chained models), we need to have the same end time.

HOW:
* Compute the end time in the main function and pass it to actual task
* Should it be an option in the configuration ?

TEST: ?

AlexisMignon:
Closes #318 

### DIFF
--- a/openstef/tasks/train_model.py
+++ b/openstef/tasks/train_model.py
@@ -40,6 +40,8 @@ def train_model_task(
     pj: PredictionJobDataClass,
     context: TaskContext,
     check_old_model_age: bool = DEFAULT_CHECK_MODEL_AGE,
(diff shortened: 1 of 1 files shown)


JanMaartenvanDoorn (approved):
Looks good! 
Small thing: I ran black as the check was failing. For your information: I am running
black==22.1.0 in the default configuration. Probably the check will pass if you run this before comitting.

PR #323, opened 2022-03-18
Bugfix train full quantile model after hyperparameter optimization

JanMaartenvanDoorn:
- fixed bug that caused quantile models to only be trained for one quantile after hyperparameter optimization


FrankKr (approved):
Great!

PR #326, opened 2022-03-28
chore - fix depr warning

FrankKr:
compare same types
Signed-off-by: Frank Kreuwel <frank.kreuwel@alliander.com>


FrankKr:
does not work

PR #328, opened 2022-03-28
Fix quantile xgb ojective function

wfstoel:
Signed-off-by: Willem Frederik Stoel <frederik.stoel@alliander.com>

### DIFF
--- a/openstef/metrics/metrics.py
+++ b/openstef/metrics/metrics.py
@@ -362,6 +362,9 @@ def xgb_quantile_obj(preds, dmatrix, quantile=0.2):
     left_mask = errors < 0
     right_mask = errors > 0
 
+    # The factor `* errors` is different from the original implementation, however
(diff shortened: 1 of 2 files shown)


openstef/metrics/metrics.py
```diff
@@ -362,7 +362,7 @@ def xgb_quantile_obj(preds, dmatrix, quantile=0.2):
     left_mask = errors < 0
     right_mask = errors > 0
 
-    grad = -quantile * left_mask + (1 - quantile) * right_mask
+    grad = (quantile * left_mask + (1 - quantile) * right_mask) * errors
```
FrankKr, line 365: Great!
Maybe add a little docstring explaing the resizing by errors (since this is different from the original, referenced loss function

wfstoel: Good idea, I will add it.

PR #329, opened 2022-03-29
Add a dependency mechanism between PJs

### LINKED ISSUE #317: Add a dependency mechanism between prediction jobs
WHAT: When chaining forecast models, we need to be able to order the prediction jobs to respect the chain order.

HOW:
* Add a `depends_on` field to the `PredictionJobDataClass` defaulting to `[]` to specify the ids of the prediction jobs the current job depends on.
* Use a graph-based algorithm to sort the prediction jobs.
*  Modify the `PredictionJobLoop` mechanism to use this order.

TEST: Check that predictions jobs are indeed executed in an order that respects the constraints

AlexisMignon:
Closes #317 


AlexisMignon:
The test failure seems to be due to a problem with a Black dependency.

AlexisMignon:
Need to 
1. Add a version to the `networkx` requirements
2. Use a specific database method to save forecasts with ids.

AlexisMignon:
@Davidswinkels 

Finally, I think there is no need to have a new method to write forecasts when the prediction job order is modified by the dependency mechanism. As I understand it from:

https://github.com/OpenSTEF/openstef/blob/86ef57f0191ceb8deec122fcf372a617d79ef656/openstef/pipeline/create_forecast.py#L123

The prediction job id is added as a column to the forecast data frame.

openstef/tasks/utils/predictionjobloop.py
```diff
@@ -5,6 +5,9 @@
 import random
 import sys
 from collections import defaultdict
+from itertools import chain
+
+from .dependencies import has_dependencies, find_groups
```
Davidswinkels, line 10: In general, absolute imports are preferred over relative imports

Davidswinkels:
> @Davidswinkels
> 
> Finally, I think there is no need to have a new method to write forecasts when the prediction job order is modified by the dependency mechanism. As I understand it from:
> 
> https://github.com/OpenSTEF/openstef/blob/86ef57f0191ceb8deec122fcf372a617d79ef656/openstef/pipeline/create_forecast.py#L123
> 
> The prediction job id is added as a column to the forecast data frame.

Agree. Jan Maarten and I discussed this dependency mechanism a bit too, and we think we can keep writing forecasts to database after each forecast.

Davidswinkels (approved):
Nice way to setup dependency mechanism with graph structure! This should not affect the way we use OpenStef at Alliander, and will make it possible for you to have chaining of predictions

Davidswinkels:
@AlexisMignon Good monday morning! If you feel the branch/feature is ready for main branch, then it can be merged. There is some black/unit test coverage fixes on main branch, but I think we don't need it to merge this branch to main.

AlexisMignon:
@Davidswinkels. It is all good for me. You can merge if it's ok for you.

PR #330, opened 2022-03-29
Bugfix quantile hyperparam optim

FrankKr:
Fixes the bug discussed with JM.
A hyperparam optimization of a model which could predict quantiles, would result in the incorrect quantiles being stored.
This fixes that.

### DIFF
--- a/test/unit/pipeline/test_optimize_hyperparameters.py
+++ b/test/unit/pipeline/test_optimize_hyperparameters.py
@@ -37,9 +37,6 @@ class TestOptimizeHyperParametersPipeline(BaseTestCase):
             "quantiles"
         ]
         self.assertTupleEqual(stored_quantiles, predefined_quantiles)
-        # And Assert that the quantile attribute of the stored model is also correct
-        quantile_attribute = save_model_mock.call_args[0][0].quantiles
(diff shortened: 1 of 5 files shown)


test/unit/pipeline/test_optimize_hyperparameters.py
```diff
@@ -65,6 +68,25 @@ def test_optimize_hyperparameters_pipeline_no_load_data(self):
...
+    ):
+        """If the regressor can predict quantiles explicitely,
+        the model should be retrained for the desired quantiles"""
+        pj = self.pj
+        predefined_quantiles = (0.001, 0.5)
+        pj["quantiles"] = predefined_quantiles
+        pj["model"] = "xgb_quantile"
+
+        parameters = optimize_hyperparameters_pipeline(
+            pj, self.input_data, "./test/unit/trained_models", n_trials=2
```
JanMaartenvanDoorn, line 83: Doesnt this slow down the unit tests by a lot? Even with 2 trials?

JanMaartenvanDoorn (approved):

test/unit/pipeline/test_optimize_hyperparameters.py
FrankKr: Good idea, I reduced it to 1

PR #334, opened 2022-04-11
KTPS-1963-make-validation-configurable

JanMaartenvanDoorn:
- remove default settings from validation.validation
- added default settings to predictionjob dataclass
- renamed clean function


FrankKr (approved):
I like it a lot!:)

PR #337, opened 2022-04-12
Bugfix/change skillscore signature

FrankKr:
Stumbled upon this during fix of other bug

- Skill score was calculated scaled to the basecase forecast, but the basecaseforecast can be empty. Makes more sense to scale using `realised`
- Method to store new KPIs did not work, added this to the TestData class


JanMaartenvanDoorn (approved):
Looks good! Please change "load" back to "basecase"

PR #341, opened 2022-04-14
Make the train pipeline return the datasets with forecasts

### LINKED ISSUE #340: The forecasts on the train, val and test datasets are not returned in the report object
WHAT: 
When the `save_train_forecasts` option is activated, the forecasts are added to the datasets and the datasets are passed to a `Reporter` object. This `Reporter` object generates a `Report` object report and this report is returned by the `train_pipeline`. However, the report  does not contain the data sets any more. So we cannot retrieve the forecasts.

HOW:
Make the pipeline functions return the datasets along the call chain.

TEST:
Check that we indeed retrieve the forecasts as columns in the data frames.

AlexisMignon:
The datasets with forecasts information were not returned through the report object. They are now directly returned by the pipeline.

This implied changing the return signature of the pipeline core function.

Closes #340


AlexisMignon:
I have also completed the tests for the train task.

JanMaartenvanDoorn (approved):
Good to fix this bug! Also good to move the report one level down to the pipeline where it is actually saved.

AlexisMignon:
@JanMaartenvanDoorn @Davidswinkels Should we merge ?

PR #352, opened 2022-04-27
Add the serialization of feature_modules

### LINKED ISSUE #347: `feature_modules` are not serialized, thus featureAdder doesn't work at forecast time.
WHAT:
For forecast task, the modelspecs' attributes are loaded from mlflow tags. The current implementation is lacking both the serialization of the `feature_modules` at the end of training during the serialization step. And the fetching of the `features_modules` at forecast time to feed the `OperationalPredictFeatureApplicator`. Thus, at forcast time the features are not computed by the `FeatureAdder`s because no modules are specified to the dispatcher instantiated in the `OperationalPredictFeatureApplicator`.

Reminder: `feature_modules` is the list of python files that contains the concrete implementation of `FeatureAdder` needed to compute the features.

HOW:
- Serialize `features_modules` as a mlflow tag likewise `feature_names` when saving a trained model.
- Add the fetching of `feature_modules` in `_get_model_specs` when loading a trained model.

TEST:
Check that we indeed serialize correctly the `feature_modules` during serialization step. And check that we fetch the `feature_modules` at forecast time.

BenoitHardier:
resolve #347 

Signed-off-by: Benoit Hardier <benoit.hardier@probayes.com>

### DIFF
--- a/openstef/model/serializer.py
+++ b/openstef/model/serializer.py
@@ -339,9 +339,9 @@ class MLflowSerializer:
             )
 
         # if feature modules is none, see if we can retrieve them from the old model
-        if model_specs.feature_modules is None:
+        if not model_specs.feature_modules:
             try:
-                if loaded_model.feature_modules is not None:
+                if loaded_model.feature_modules:
                     model_specs.feature_modules = loaded_model.feature_modules
                     self.logger.info(
                         "feature_modules retrieved from old model with an attribute",
--- a/openstef/model/serializer.py
+++ b/openstef/model/serializer.py
@@ -84,11 +84,12 @@ class MLflowSerializer:
         mlflow.set_tag("model_type", model_type)
         mlflow.set_tag("prediction_job", experiment_name)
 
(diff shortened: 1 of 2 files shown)


openstef/model/serializer.py
```diff
@@ -300,6 +305,55 @@ def _get_feature_names(
                 )
         return model_specs.feature_names
 
+    def _get_feature_modules(
```
JanMaartenvanDoorn, line 308: Nice addition and nice exception handling with clear messagages! As there is some logic in this functions, could you please make some unit tests that check the if statements in lines 342 and 344?

JanMaartenvanDoorn (changes requested):
I agree with the changes, however I am missing some unit tests that test the logic in lines 342 and 344

openstef/model/serializer.py
BenoitHardier: The tests has been added. Coverage is 100% now. Thank you.

JanMaartenvanDoorn (approved):

PR #354, opened 2022-05-03
Chore/black with signoff and fix append to concat

Davidswinkels:


Davidswinkels:
Duplicate from https://github.com/OpenSTEF/openstef/pull/355

PR #356, opened 2022-05-04
made core version of hyperparameter pipeline

JanMaartenvanDoorn:
-made a core pipeline of the hyperparameter optimization without serializer depedency

### DIFF
--- a/openstef/pipeline/optimize_hyperparameters.py
+++ b/openstef/pipeline/optimize_hyperparameters.py
@@ -62,6 +62,41 @@ def optimize_hyperparameters_pipeline(
     Returns:
         dict: Optimized hyperparameters.
     """
+
+    (
+        best_model,
+        model_specs,
+        report,
+        trials,
+        best_trial_number,
+        best_params,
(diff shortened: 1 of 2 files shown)


openstef/pipeline/optimize_hyperparameters.py
```diff
@@ -62,6 +63,64 @@ def optimize_hyperparameters_pipeline(
...
+        phase="Hyperparameter_opt",
+        trials=trials,
+        trial_number=best_trial_number,
+    )
+    if artifact_folder:
+        Reporter.write_report_to_disk(report=report, artifact_folder=artifact_folder)
+    return best_params
+
+
+def optimize_hyperparameters_pipeline_core(
```
Davidswinkels, line 95: Nice to split the pipeline into a core without model saving and a pipeline with model saving! Can you add one simple unit test for optimize_hyperparameters_pipeline_core? Just one simple unit test will do in my opinion and probs existing tests can be reused :)

Davidswinkels (changes requested):
Change looks good! Unit test for new pipeline core will still be good

Davidswinkels (approved):
Nice change!

PR #358, opened 2022-05-09
fix-make-fold-randomization-backtest-configurable

JanMaartenvanDoorn:
- made fold randomization configurable 
- changed name of inpur_data to data in default splitting function to make is more consistent with requirements for custom splitting functions
- changed default behavior to 
image
instead of 
image


FrankKr (approved):
Nice improvement!

PR #375, opened 2022-06-01
Fix logging of hyper parameters

Davidswinkels:


Davidswinkels:
Problem was already resolved by pinning protobuf in mlflow tracking server and in openstef location

PR #376, opened 2022-07-05
Improve performance of replace repeated values with nan

Davidswinkels:
Plus run isort and black

New function to replace nan has improved performance from:
725 ms ± 111 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
to:
2.41 ms ± 164 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

### DIFF
--- a/openstef/data_classes/prediction_job.py
+++ b/openstef/data_classes/prediction_job.py
@@ -2,10 +2,12 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-from typing import Union, Optional, List, Dict, Any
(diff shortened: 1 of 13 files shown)


FrankKr (approved):
Nice improvement!
If you could nake it possible to keep the first 24 flatline values that would be great.
Or even better, if you could make the `keep_first` an input argument for the replace_repeated_values_with_nan, which maybe uses 0 by default

I did not check the unit tests, I assumed you tested the code yourself.

Davidswinkels:
Kept the first 24 flatline values :)

Issue #380, opened 2022-07-18
trained model figures are stored one level too high


FrankKr:
https://github.com/OpenSTEF/openstef/blob/35e49e32b2cabf4610a2342f446e677abb10238f/openstef/metrics/reporter.py#L129

This should be {artifact_folder}/{pid}/{name}.html.
With this implementation, e.g. 'Predictors_47.0.html' is stored in ../trained_models/ and is overwritten by the next trained model.

PR #382, opened 2022-07-19
Bugfix ktps 2272 #380 reporter location

FrankKr:
Fix location where reports (artifacts) are stored.
The current implementation (made when adopting the mlflow functionality) wrote the reports one level too high, resulting in them being overwritten for each new training that happened (also for other prediction_jobs). 

Fixed the location where they are stored back to what we used, and added tests for this.

As a bonus I changed CountryHolidays to country_holidays, as that gave a FutureWarning and I had to wait for the build to complete.


Davidswinkels (approved):
LGTM

PR #387, opened 2022-07-27
Bugfix ktps 2298 feature mismatch

FrankKr:


openstef/pipeline/create_forecast.py
```diff
@@ -82,7 +82,7 @@ def create_forecast_pipeline_core(
     data_with_features = OperationalPredictFeatureApplicator(
         # TODO use saved feature_names (should be saved while training the model)
         horizons=[pj["resolution_minutes"] / 60.0],
-        feature_names=model.feature_names,
+        feature_names=model.feature_importance_dataframe.index.values,
```
Davidswinkels, line 85: Good fix! Is it possible to add a test (or multiple) to make sure we can avoid the bug in the future?

Davidswinkels (changes requested):
See comment

openstef/pipeline/create_forecast.py
FrankKr: The tests were there, but the previously stored model was not compatible with those tests, as at was pretty old. 
I've updated the stored model to allow for accurate testing.

Davidswinkels (approved):
LGTM

PR #388, opened 2022-08-01
Bugfix/ktps 2133 score attr xgbquantile

FrankKr:
Used composition to replace the .score method (which used to be inherited).
Also added a unit test to test this method for all models explicitely

### DIFF
--- a/openstef/model/regressors/regressor.py
+++ b/openstef/model/regressors/regressor.py
@@ -15,7 +15,7 @@ class OpenstfRegressor(OpenstfRegressorInterface):
         self.feature_importance_dataframe = None
         self.feature_importances_ = None
         # This `score` attribute makes `score` method from RegressorMixin available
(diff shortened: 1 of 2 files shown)


openstef/model/regressors/regressor.py
```diff
@@ -13,6 +14,7 @@ class OpenstfRegressor(OpenstfRegressorInterface):
     def __init__(self):
         self.feature_importance_dataframe = None
         self.feature_importances_ = None
+        self.score = RegressorMixin.score
```
Davidswinkels, line 17:
```suggestion
        # This `score` field makes `score` method from RegressorMixin available
        self.score = RegressorMixin.score    
```

Davidswinkels (changes requested):
Score method can be more clear

Davidswinkels (approved):
Thanks for the fix

PR #389, opened 2022-08-01
integrate abstractions regressor interface

FrankKr:
Removed the abstract-model layer of the openstefregressors.
Situation was:
AbstractOpenstefClass -> OpenstefClass -> SpecificModelClass (e.g. XGB)

I think the additional layer with the abstractmethods made everything less clear. Stumbled upon this while fixing another bug and had to have a couple of looks of which class did what.
What do you think?

Signed-off-by: Frank Kreuwel <fpmkreuwel@gmail.com>


wfstoel (approved):
LGTM!

Issue #339, opened 2022-04-13
Bad typehint for PredictionJobDataClass.train_horizons_mintues


AlexisMignon:
WHAT:
The current typehint is: `Optional[List[int]]` while it should be `Optional[Union[str, List[int]]]`

HOW: Simply modify the typehint

TEST: None needed. (Should have triggered a code smell ?)

FrankKr:
The addition of being able to specify the horizon as *str* was added in this commit: https://github.com/OpenSTEF/openstef/commit/8bc20f955c98d00354dd712a069093c6610acdc9 I don't completely understand why this was added, or why this was added to the feature applicator, but not to the PJ datamodel. @BenoitHardier (author)/ @JanMaartenvanDoorn (reviewer) could you clarify?

A PR to revert back to the more strict situation where the horizons should be a list of floats is here: #390

AlexisMignon:
We introduced the possibility to indicate a column name instead of a list of horizons to address the case when the horizon information is already in the input data. This can be typically the case when forecast information is used as explanatory variable (e.g. from a weather forecast service).

PR #390, opened 2022-08-02
Related to #339 - remove str option for horizon_minutes 

FrankKr:
See issue #339


FrankKr:
Closed after discussing with Benoît

PR #397, opened 2022-09-07
fixed some issues

JanMaartenvanDoorn:
- removed splitting coefficients from task
- drop nans for PTU's where no weather data could befound
- Added check that at leat for two days ahead a forecast should be available.
- Adapted unit test to new situation and added some more explicit testing on a mock dataframe

### DIFF
--- a/openstef/tasks/create_components_forecast.py
+++ b/openstef/tasks/create_components_forecast.py
@@ -31,6 +31,7 @@ Attributes:
 """
 from datetime import datetime, timedelta
 from pathlib import Path
+import pytz
 
 import structlog
 
@@ -95,7 +96,9 @@ def create_components_forecast_task(pj: PredictionJobDataClass, context: TaskCon
     # Make forecast for the demand, wind and pv components
     forecasts = create_components_forecast_pipeline(pj, input_data, weather_data)
 
-    if forecasts.index.max() < datetime.utcnow() + timedelta(days=2):
+    if forecasts.index.max() < datetime.utcnow().replace(tzinfo=pytz.utc) + timedelta(
+        days=2
+    ):
         raise ValueError(
             "Could not make component forecast for two days ahead, probably input data is missing."
         )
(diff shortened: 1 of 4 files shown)


openstef/tasks/create_components_forecast.py
```diff
@@ -92,17 +93,15 @@ def create_components_forecast_task(pj: PredictionJobDataClass, context: TaskCon
...
-        logger.warning(f"No Coefs found. Skipping pid", pid=pj["id"])
-        return
-
     # Make forecast for the demand, wind and pv components
-    forecasts = create_components_forecast_pipeline(
-        pj, input_data, weather_data, split_coefs
-    )
+    forecasts = create_components_forecast_pipeline(pj, input_data, weather_data)
+
+    if forecasts.index.max() < datetime.utcnow().replace(tzinfo=pytz.utc) + timedelta(
```
FrankKr, line 99: 2 days is too long I think.
We retrieve Harm_Arome data for more or less 48 hours ahead, but if it is only available up to 36 hours ahead every now and again, that is to be expected

openstef/tasks/create_components_forecast.py
```diff
@@ -92,17 +93,15 @@ def create_components_forecast_task(pj: PredictionJobDataClass, context: TaskCon
...
+        days=2
+    ):
+        raise ValueError(
```
FrankKr, line 102: Better to define a custom exception. e.g:
```python
class ForecastTooShortHorizonError(Exception):
    pass
#...
raise ForecastTooShortHorizonError("Could not make component forecast far enough ahead, probably input data is missing.")
```

FrankKr (approved):
Looks like a good improvement. Made two suggestions

openstef/tasks/create_components_forecast.py
JanMaartenvanDoorn: Good tip!, I set it to 30 hours now.

openstef/tasks/create_components_forecast.py
JanMaartenvanDoorn: Yes, good suggestion!

PR #406, opened 2022-10-25
use the attribute feature_names of the model instead of the index of …

BenoitHardier:
…the feature_importance_dataframe


JanMaartenvanDoorn (approved):
Looks good!

PR #415, opened 2022-11-07
405 use model from other prediction job and pipelines_to_run pj field

BenoitHardier:


openstef/data_classes/prediction_job.py
```diff
@@ -81,6 +82,10 @@ class PredictionJobDataClass(BaseModel):
     """Only required for create_wind_forecast task"""
     hub_height: Optional[float]
     """Only required for create_wind_forecast task"""
+    pipelines_to_run: PipelineType = PipelineType.BOTH
+    """The pipelines to run for this pj"""
+    forecast_pid: Optional[Union[int, str]]
```
JanMaartenvanDoorn, line 87: forecast_pid seems a bit too generic to me, to avoid confusion what do you think of renaming "forecast_pid" to "alternative_forecast_model_pid". I think this describes the id better.

openstef/pipeline/create_forecast.py
```diff
@@ -41,10 +41,12 @@ def create_forecast_pipeline(
         DataFrame with the forecast
 
     """
+    pid = pj["id"] if pj.forecast_pid is None else pj.forecast_pid
```
JanMaartenvanDoorn, line 44: Good to add a small comment on what is happening here e.g.: Selecting alternative forecast model if it exists

openstef/data_classes/prediction_job.py
```diff
@@ -8,6 +8,7 @@
 
 from .model_specifications import ModelSpecificationDataClass
 from .split_function import SplitFuncDataClass
+from ..enums import PipelineType
```
JanMaartenvanDoorn, line 11:
```suggestion
from openstef.enums import PipelineType
```

openstef/data_classes/prediction_job.py
```diff
@@ -8,6 +8,7 @@
 
 from .model_specifications import ModelSpecificationDataClass
 from .split_function import SplitFuncDataClass
```
JanMaartenvanDoorn, line 10:
```suggestion
from openstef.data_classes.split_function import SplitFuncDataClass
```

openstef/data_classes/prediction_job.py
```diff
@@ -8,6 +8,7 @@
 
 from .model_specifications import ModelSpecificationDataClass
```
JanMaartenvanDoorn, line 9:
```suggestion
from openstef.data_classes.model_specifications import ModelSpecificationDataClass
```

openstef/pipeline/create_forecast.py
JanMaartenvanDoorn: Maybe also good to make pid more specific, e.g. prediction_model_pid

openstef/data_classes/prediction_job.py
```diff
@@ -81,6 +82,10 @@ class PredictionJobDataClass(BaseModel):
     """Only required for create_wind_forecast task"""
     hub_height: Optional[float]
     """Only required for create_wind_forecast task"""
+    pipelines_to_run: PipelineType = PipelineType.BOTH
```
JanMaartenvanDoorn, line 85: What do you think of implementing this as a list of PipelineTypes? This way we do not have to define a separate enum catergory for each pipeline combination we want to run.

openstef/tasks/optimize_hyperparameters.py
```diff
@@ -44,6 +44,10 @@ def optimize_hyperparameters_task(
         context: Task context
 
     """
+    if pj.pipelines_to_run == PipelineType.FORECAST:
```
JanMaartenvanDoorn, line 47: See earlier comment about implementing the pipelines_to_run as a list. In that case the check here would be to check if the pipelines_to_run contains a hyper_parameters type.

openstef/tasks/train_model.py
```diff
@@ -55,6 +55,10 @@ def train_model_task(
         datetime_end: End
 
     """
+    if pj.pipelines_to_run == PipelineType.FORECAST:
```
JanMaartenvanDoorn, line 58: Similare to earliers comments

JanMaartenvanDoorn (changes requested):
Nice addition!, I have some recommendations, have a look at the comments, if something is not clear or you want to discuss feel free to give me a call.

openstef/data_classes/prediction_job.py
BenoitHardier: Yes you are right, I'll clarify the name.

openstef/pipeline/create_forecast.py
BenoitHardier: If the names of the variables are more precise it'll be clearer. I make that change too.

openstef/data_classes/prediction_job.py
BenoitHardier: Yes, It's probably better with a list. I follow your suggestion.

openstef/data_classes/prediction_job.py
```diff
@@ -2,12 +2,13 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 """Specifies the prediction job dataclass."""
-from typing import Optional, Union
+from typing import Optional, Union, List
```
JanMaartenvanDoorn, line 5:
```suggestion
from typing import Optional, Union
```

openstef/data_classes/prediction_job.py
```diff
@@ -81,6 +82,14 @@ class PredictionJobDataClass(BaseModel):
     """Only required for create_wind_forecast task"""
     hub_height: Optional[float]
     """Only required for create_wind_forecast task"""
+    pipelines_to_run: List[PipelineType] = [
```
JanMaartenvanDoorn, line 85:
```suggestion
    pipelines_to_run: list[PipelineType] = [
```

openstef/pipeline/create_forecast.py
```diff
@@ -41,10 +41,15 @@ def create_forecast_pipeline(
         DataFrame with the forecast
 
     """
+    prediction_model_pid = pj["id"]
+    # Use the alternative forecast model if it's specify in the pj
+    if pj.alternative_forecast_model_pid:
```
JanMaartenvanDoorn, line 46: Nice!

openstef/tasks/create_basecase_forecast.py
```diff
@@ -42,6 +43,10 @@ def create_basecase_forecast_task(
         context: Contect object that holds a config manager and a database connection
 
     """
+    if pj.pipelines_to_run == PipelineType.TRAIN:
```
JanMaartenvanDoorn, line 46: Will this work? As pj.pipelines_to_run is a list now I think this will break.

JanMaartenvanDoorn: I think nicest is to add a pipelinetype for the basecase forecast and do a check similar to the ones in the train and forecasts pipelines

JanMaartenvanDoorn (changes requested):
Nice changes, I however found something that might break but is probably easy to fix.

openstef/tasks/create_basecase_forecast.py
BenoitHardier: Sure, I forgot to change it. Nice catch!

JanMaartenvanDoorn (approved):
Looks good!

PR #421, opened 2022-11-28
fix issue

JanMaartenvanDoorn:
-fixed issue with component forecasts being positive despite postprocessing


MartijnCa:
The changes you made look good 👍. 

There is something I do not understand though, which does not relate to the changes you made: The function `post_process_wind_solar` recognizes the sign of the forecast. I would expect the function to either return the recognized sign or to adjust the sign according to what was recognized.  Is this not necessary?

MartijnCa (approved):

PR #427, opened 2022-12-01
Making sure model_specs from previous model are always used if availa…

MartijnCa:
…ble when retraining. To train with new model_specs, a hyperparameter optimization needs to be dones first.

Signed-off-by: Martijn Cazemier <martijn.cazemier@alliander.com>


JanMaartenvanDoorn (approved):
Looks good!

PR #431, opened 2022-12-28
added error

JanMaartenvanDoorn:
- added check on datetime index of input dataframe
- raises a Valuerror in the rare case this happens


Luukhalsema (approved):
Nice!

PR #432, opened 2023-01-02
Update depricated xgboost functionality

JanMaartenvanDoorn:
- updated dependencies, including update to mlflow >2 -> Does this cause any problems?
- updated deprecated way of picking best ieteration in quantile models
- xgboost could be upgraded from 1.6.1 to 1.7.2 but this causes problems when loading lold models, I will make an issue to tackle this.


wfstoel (approved):
LGTM!

PR #433, opened 2023-01-16
upgrade xgboost version

FrankKr:
Included additional attributes when loading an old model for compatibility. Added to the docstrings that that code is for compatibility


JanMaartenvanDoorn (approved):
Looks good!

PR #437, opened 2023-01-30
434 autoregressive model and more generic feature applicator

### LINKED ISSUE #434: Autoregressive model (SARIMAX) in OpenSTEF and more generic FeatureApplicator
WHAT: 
As discussed in some refinement sessions, we want to add the possibility to train and perform forecast for autoregressive models (ie _timeseries oriented_ models contrary to _pointwise_ models already available in OpenSTEF). We have in mind to begin to implement _(S)ARIMA(X)_ models thanks to the [statmodels](https://www.statsmodels.org/stable/index.html) package.  
The workflow for  such model, especially for the  forecast, has some discrepancies with the classical _sklearn-like_ worflow. The main difference is the need of the past targets `y_past` (and the past exogenous/feautres data `X_past`) at forecast time, but such past data are, now by deflaut, filtered out with the function `generate_forecast_datetime_range` that keeps the timestamps with `load` column empty (more precisely with NaN values). Moreover we also need to deactivate the feature engineering generation, especially for the lag-features that are already handle by the model itself.   
  
HOW:
1. Thus, we propose to replace the FeatureApplicator (and the filtering of forcast datetime interval) by a more generic object with the same default behaviors than the existing FeatureApplicator but it can have different behaviors accordings the parametrization of the model, the prediction job and/or the modelspecs. We also think that the splitting between _Operational_ and _Train_ feature applictor is not compulsary and we can simplify the way it is done. Finally, we are convinced that eventually most of the logic/controls flow in the Feature applicator could be simplified with this new generic mecanism.  
This solution have been designed to prevent most of code duplication and deep change in the framework but we are obviously open to suggestions for refining it.

2. The second step is to implement the OpenSTEFRegressor that wrap the proper ARIMA model.

I'll try to implement a proof of concept in the PR linked with this issue.
 
TEST:

- Be sure that default behavior is the current behavior (no regression).

- Test simple Autoregressive model with train and forecast pipeline.

BenoitHardier:
close #434

### DIFF
--- a/openstef/data_classes/data_prep.py
+++ b/openstef/data_classes/data_prep.py
@@ -1,0 +1,99 @@
+# SPDX-FileCopyrightText: 2017-2022 Contributors to the OpenSTEF project <korte.termijn.prognoses@alliander.com> # noqa E501>
+#
+# SPDX-License-Identifier: MPL-2.0
+"""Specifies the split function dataclass."""
+import inspect
+import json
+from importlib import import_module
+from typing import Any, Sequence, Union, TypeVar
+
+from pydantic import BaseModel
+
+DataPrepClass = TypeVar("DataPrepClass")
+
+
+class DataPrepDataClass(BaseModel):
+    """Class that allows to specify a custom class to prepare the data (feature engineering , etc ...)."""
+
+    klass: Union[str, type[DataPrepClass]]
+    arguments: Union[
+        str, dict[str, Any]
+    ]  # JSON string holding the function parameters or dict
+
+    def __getitem__(self, key: str):
+        """Allows us to use subscription to get the items from the object."""
+        return getattr(self, key)
+
+    def __setitem__(self, key: str, value: any):
+        """Allows us to use subscription to set the items in the object."""
+        if hasattr(self, key):
+            self.__dict__[key] = value
+        else:
+            raise AttributeError(f"{key} not an attribute of prediction job.")
+
+    def _load_klass(
+        self, required_arguments: Sequence[str] = None
+    ) -> type[DataPrepClass]:
+        """Load data prep class from path.
+
+        Args:
+            klass_path (str): The path to the data prep class
+
+        Returns:
+            klass (type[AbstractDataPreparation]): The loaded data prep class
+
+        """
+        if isinstance(self.klass, str):
+            path_elements = self.klass.split(".")
+            module_path = ".".join(path_elements[:-1])
+            module = import_module(module_path)
+            klass_name = path_elements[-1]
+            klass = getattr(module, klass_name)
+        else:
+            klass = self.klass
+
+        # Check that the klass accepts mandatory arguments
+        if not inspect.isclass(klass):
+            raise ValueError("The loaded object is not a class: {klass!r}")
+
+        if required_arguments is not None:
+            klass_params = set(inspect.signature(klass).parameters)
+
+            if len(set(required_arguments) - klass_params) > 0:
+                raise ValueError(
+                    "The loaded data prep class does not have the required arguments"
+                )
+
+        return klass
+
+    def _load_arguments(self) -> dict[str, Any]:
+        """Load the arguments.
+
+        Convert the arguments from JSON if they are given as strings or simply return them otherwise.
+
+        Returns:
+            arguments (dict[str, Any]): The additional arguments to be passed to the class
+
+        """
+        if isinstance(self.arguments, str):
+            return json.loads(self.arguments)
+        else:
+            return self.arguments
+
+    def load(
+        self, required_arguments: Sequence[str] = None
+    ) -> tuple[type[DataPrepClass], dict[str, Any]]:
+        """Load the function and its arguments.
+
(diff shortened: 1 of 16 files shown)


openstef/data_classes/data_prep.py
```diff
@@ -0,0 +1,99 @@
...
+        return getattr(self, key)
+
+    def __setitem__(self, key: str, value: any):
+        """Allows us to use subscription to set the items in the object."""
+        if hasattr(self, key):
+            self.__dict__[key] = value
+        else:
+            raise AttributeError(f"{key} not an attribute of prediction job.")
+
+    def _load_klass(
```
JanMaartenvanDoorn, line 34:
```suggestion
    def _load_class(
```

openstef/model/regressors/arima.py
```diff
@@ -0,0 +1,163 @@
...
+
+from sklearn.metrics import r2_score
+from sklearn.model_selection import TimeSeriesSplit
+from openstef.model.regressors.regressor import OpenstfRegressor
+
+
+class ARIMAOpenstfRegressor(OpenstfRegressor):
+    """Wrapper around statmodels implementation of (S)ARIMA(X) model.
+
+    To make prediction, it needs to update its historic, applying the fitted parameters to new data unrelated to the
```
JanMaartenvanDoorn, line 17:
```suggestion
    To make a prediction, it needs to update its historic, applying the fitted parameters to new data unrelated to the
```

JanMaartenvanDoorn: I do not understand the word historic here, do you mean historic data?

openstef/model/regressors/arima.py
```diff
@@ -0,0 +1,163 @@
...
+        seasonal_order=(0, 0, 0, 0),
+        trend=None,
+    ):
+        self.backtest_max_horizon = backtest_max_horizon
+        self.order = order
+        self.seasonal_order = seasonal_order
+        self.trend = trend
+
+    def fit(self, x, y, **kwargs):
+        exog = x
```
JanMaartenvanDoorn, line 35: Maybe add a link to https://www.statsmodels.org/stable/endog_exog.html to give a bit more background on the exog terminology as it is not that common.

openstef/model/regressors/arima.py
```diff
@@ -0,0 +1,163 @@
...
+            dates=dates,
+            order=self.order,
+            seasonal_order=self.seasonal_order,
+            trend=self.trend,
+        )
+        self.results_ = self.model_.fit()
+        self.feature_in_names_ = list(x.columns)
+        return self
+
+    def update_historic(self, y_past, x_past):
```
JanMaartenvanDoorn, line 49: Also here I do not understand the historic terminology, do you mean history or maybe historic data?

openstef/model/regressors/arima.py
```diff
@@ -0,0 +1,163 @@
...
+        ys_pred = []
+
+        freq = pd.infer_freq(x.index)
+        if not (freq[0].isdigit()):
+            freq = f"1{freq}"
+        test_size = pd.Timedelta(self.backtest_max_horizon, "minutes") // pd.Timedelta(
+            freq
+        )
+        n_splits = (x.shape[0] // test_size) - 1
+        tscv = TimeSeriesSplit(n_splits=n_splits, test_size=test_size)
```
JanMaartenvanDoorn, line 152: Please use the full term here and not the abbreviation. I think you mean time_series_cross_validation ?

openstef/model/regressors/arima.py
```diff
@@ -0,0 +1,163 @@
...
+        for apply_index, test_index in tscv.split(x):
+            arima_results = self.results_.apply(
+                y.iloc[apply_index], x.iloc[apply_index]
+            )
+            x_ts = x.iloc[test_index]
```
JanMaartenvanDoorn, line 157: Same comment about the _ts postfix, please use the full term

JanMaartenvanDoorn (changes requested):
Looks good, nice to have ARIMA available in openSTEF! I have a few comments regarding readability.

openstef/model/regressors/arima.py
BenoitHardier: I refactored this part to be more readable. Names are more explicit.

openstef/model/regressors/arima.py
BenoitHardier: Great suggestion: 
- I added multiple links to relevant documentations of statsmodels
- I removed the renaming of variables and explicitly passed the arguments as keyword argument to be clearer.

openstef/model/regressors/arima.py
BenoitHardier: Yes I enhanced the documentation, it should be clearer.

BenoitHardier:
> Looks good, nice to have ARIMA available in openSTEF! I have a few comments regarding readability.

Thank you for the remarks, I refactored a bit the class and added more documentation to enhance the readability.

JanMaartenvanDoorn (approved):
Thanks for adding the documentation and updating the names, this makes it more clear!

PR #446, opened 2023-03-01
default split for train_pipeline and scoring

BenoitHardier:


openstef/model_selection/model_selection.py
```diff
@@ -173,9 +174,8 @@ def split_data_train_validation_test(
         test_data = data_[start_date_test:]
         train_val_data = data_[:start_date_test]
     else:
-        start_date_val = start_date + np.round(number_indices * test_fraction) * delta
-        test_data = data_[:start_date_val]
-        train_val_data = data_[start_date_val:]
+        test_data = data_.copy(deep=True)
```
JanMaartenvanDoorn, line 177: I remember disusing this but on a closer look I do not understand this anymore. For the split_test_train_validation data I would expect three different datasets to come out (may also be empty datasets).

What is the problem again with just using `pd.concat([train_data, validation_data])` on the place were you want to use the test_data_set?

openstef/model_selection/model_selection.py
```diff
@@ -141,6 +141,7 @@ def split_data_train_validation_test(
         - Test data.
 
     """
+    test_fraction = test_fraction if back_test else 0
```
JanMaartenvanDoorn, line 144: This seems logical

openstef/pipeline/train_model.py
```diff
@@ -199,14 +199,13 @@ def train_model_pipeline_core(
 
     # Check if new model is better than old model
     if old_model:
-        combined = pd.concat([train_data, validation_data]).reset_index(drop=True)
         # skip the forecast column added at the end of dataframes
         if pj.save_train_forecasts:
-            combined = combined.iloc[:, :-1]
+            test_data = test_data.iloc[:, :-1]
```
JanMaartenvanDoorn, line 204: See question above

test/unit/pipeline/test_train.py
```diff
@@ -73,10 +73,9 @@ def test_split_data_train_validation_test_stratification_no_backtest(self):
             delta=2,
         )
```
JanMaartenvanDoorn, line 74: Nice improvements in this file, make is more clear!

JanMaartenvanDoorn (commented):
Upon closer look I think I still do not understand why we should change the test_data set. Can you explain a bit more about this?

openstef/model_selection/model_selection.py
BenoitHardier: The main reason is related to the use of custom split function. If you implement your own split procedure it seems very akward to design it with this concat operation in mind. Moreover it seems very sensible to compute scores/metrics on the test set.   
My changes do not affect the fact that the split function returns 3 datasets (they are just overlapping when `back_test=False`).   
The second issue is the fact that this concat can introduce weird behaviours when the order of train and val are shuffled. This has no consequences for pointwise models like xbgm/lightgbm because lag features has been computed before splitting. But with ARIMA (or other ar models like prophete) it can be very painfull, that why we prefer to keep this operation behind the split function.

PR #449, opened 2023-03-09
added operational score set

JanMaartenvanDoorn:
- Added an extra output of the splitting function to make extra clear which data is going to be used for what purpose

### DIFF
--- a/openstef/model/objective.py
+++ b/openstef/model/objective.py
@@ -87,6 +87,7 @@ class RegressorObjective:
             self.train_data,
             self.validation_data,
             self.test_data,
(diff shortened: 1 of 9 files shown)


BenoitHardier:
Hi @JanMaartenvanDoorn ,
with @AlexisMignon we are not convinced by adding this 4th dataset. It means that one of the 4 dataset is not used depending the context. 
Can we just keep the fact that the test set is  not same depending the value of the argument `back_test` ? 
We can add some extra comments and rename some variables to be very clear about the behavior of the split function.

PR #451, opened 2023-03-09
set n estimators if none

FrankKr:
Fixes current bug due to version upgrade xgboost and old models not having a required param


JanMaartenvanDoorn (approved):
Good fix!

PR #456, opened 2023-04-25
Update requirements

wfstoel:
Noticeable updates:
- Removed patch version from dependencies, this means openstef is more tolerable to updates.
- Removed pytz in favor of native python datetime.timezone.
- Updated pandas to 2.0 and rewrote unit tests to work with the changes.
- Updated attached trained_model.

The large amount of file changes is largely caused by the removal of unnecessary white space.


FrankKr (approved):
LGTM!

PR #457, opened 2023-04-28
Feature/hyperparam trains new model

FrankKr:
minor improvement, see comment in code why this is useful.
Caused by KTPS-2825


MartijnCa (approved):
Nice improvement!

PR #462, opened 2023-05-12
Bugfix/ktps 2838 stdev explodes

FrankKr:
Adds a cap on the extrapolation of the standard deviation for non-trained horizons.
Also, improved the (flawed) unit test of the stdev calculation.


JanMaartenvanDoorn (approved):
Nice fix!

PR #465, opened 2023-08-04
add uncertainty estimation docs

FrankKr:


FrankKr:
recreate pr so that latest changes from main are included

PR #466, opened 2023-08-04
skip failing test

FrankKr:
This unit test has been spontaneously failing since a number of commits. It does work offline, making it very hard to debug.
Agreed to skip it?

Alternatively; do you know an easy way to only skip it online, but still run it offline? I think I saw a construction for that somewhere.


wfstoel (approved):
Im fine by setting the unit test to skip. Alternatively you could consider changing the assert check on line 50 to use `set` instead of `list`. Possibly the order of the columns is not guaranteed?

FrankKr:
Thanks for having a more detailed look as well. 
Unfortunately, the assertion before that one fails already.
I'm marking the unittest and will create an issue for it.

PR #472, opened 2023-08-11
remove incorrect report creation

FrankKr:
Verified locally that this solves the issue.
I did not add a unit test, since this involved retrieving data from the figure which can be done, but gets quite complex rather quickly.


openstef/pipeline/optimize_hyperparameters.py
```diff
@@ -188,7 +188,7 @@ def optimize_hyperparameters_pipeline_core(
     )
 
     best_hyperparams = study.best_params
-    best_model = study.user_attrs["best_model"]
+    # the best_model could be accessed via study.user_attrs["best_model"]
```
JanMaartenvanDoorn, line 191:
```suggestion
    # The best_model could be accessed via study.user_attrs["best_model"]
```

JanMaartenvanDoorn (approved):
Looks good! Added a minor suggestion.

PR #475, opened 2023-09-12
Make model not found uniform

wfstoel:
This change makes the model not found exception uniform. Goal is to make the teams alerts contain a single exception with multiple pids instead of an exception per pid.


bartpleiter (approved):
LGTM!

PR #476, opened 2023-09-15
Feature/ktps 3015 remove old mlflow artifacts

bartpleiter:
As for the unit test: mlflow does not support relative paths for the artifacts in the meta.yaml files. Furthermore, creating four new models takes too long for a unit test. Therefore, I choose to dynamically update the yaml files during the test in the temp directory, using a helper function (I wonder if the placement of this function is okay)


wfstoel (approved):
LGTM!

PR #478, opened 2023-10-02
Known zero flatliners exception handling

MartijnCa:


FrankKr (approved):
Looks good, provided two suggestions

PR #479, opened 2023-10-17
Upgrade pydantic to v2 while referencing v1

GBBBAS:
The LF Energy RTDIP team leverage pydantic v2 and currently working on integration with Openstef. One particular package that we would like to upgrade in Openstef is pydantic. 

To not break code, pydantic has provided a [migration procedure](https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features) that allows upgrading of the package but still being able to reference pydantic using `from pydantic.v1 import ....`. 

This PR therefore upgrades the pydantic version in requirements.txt and also updates any imports in the code to use `from pydantic.v1`

### DIFF
--- a/openstef/data_classes/data_prep.py
+++ b/openstef/data_classes/data_prep.py
@@ -7,7 +7,7 @@ import json
 from importlib import import_module
 from typing import Any, Sequence, Union, TypeVar
 
-from pydantic import BaseModel
+from pydantic.v1 import BaseModel
 
 DataPrepClass = TypeVar("DataPrepClass")
 
--- a/openstef/data_classes/model_specifications.py
+++ b/openstef/data_classes/model_specifications.py
@@ -4,7 +4,7 @@
 """Specifies the dataclass for model specifications."""
 from typing import Optional, Union
 
-from pydantic import BaseModel
+from pydantic.v1 import BaseModel
 
 
 class ModelSpecificationDataClass(BaseModel):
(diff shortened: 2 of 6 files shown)


FrankKr:
@GBBBAS Thanks for making this PR!
Could you please bump the version in the setup.py to 3.3.4?

I don't have access to make this change on your fork :)

GBBBAS:
> @GBBBAS Thanks for making this PR! Could you please bump the version in the setup.py to 3.3.4?
> 
> I don't have access to make this change on your fork :)

Done!

wfstoel:
@GBBBAS We had some issues with our GitHub actions build & test pipeline in combination with forks (as we never use those ourselves). It should be fixed now, but we cannot run them afterward. Could you repush the PR to force the pipelines to run?

My apologies for the inconvience.

GBBBAS:
> @GBBBAS We had some issues with our GitHub actions build & test pipeline in combination with forks (as we never use those ourselves). It should be fixed now, but we cannot run them afterward. Could you repush the PR to force the pipelines to run?
> 
> My apologies for the inconvience.

Good to chat again! I see you are getting the sonarcloud error on forked repos. We had the same issue on RTDIP. If its easier, let me know if you want us to be granted access to the repo, and we work on a branch directly on this repo instead of a fork?

wfstoel (approved):

wfstoel:
I added you as to this repository with write access. So for future changes you could create a branch instead. We'll have a look to make this process a better for the future.

For now I'll force merge this PR. Thanks for the contribution!

PR #482, opened 2023-10-30
Added fix and unit test for detecting flatliners on prediction data w…

bartpleiter:
…ith NaN load values in the future.

### DIFF
--- a/test/unit/validation/test_validation_detect_ongoing_zero_flatliner.py
+++ b/test/unit/validation/test_validation_detect_ongoing_zero_flatliner.py
@@ -3,26 +3,28 @@
 # SPDX-License-Identifier: MPL-2.0
 
 from datetime import datetime, timedelta
+from freezegun import freeze_time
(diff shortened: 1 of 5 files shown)


test/unit/validation/test_validation_detect_ongoing_zero_flatliner.py
```diff
@@ -14,6 +14,10 @@
     start=now - timedelta(minutes=180), end=now, freq="0.25H"
 )
 
+four_hour_range_predict_setting = pd.date_range(
+    start=now - timedelta(minutes=180), end=now + timedelta(minutes=60), freq="0.25H"
+)
+
```
wfstoel, line 20: It's better to move this code into the test itself. If for some reason there is big delay between initializing this module and running the test, the datetime will not have the correct value anymore.

wfstoel (approved):

wfstoel (approved):
Nice use of freezegun to make the unit test more stable!

PR #485, opened 2023-12-07
Use train model horizons specified in prediction job correctly

MartijnCa:


bartpleiter (approved):
Nice changes! And good unit tests 👌

PR #487, opened 2023-12-08
Allow for input of databricks workspace path

rodalynbarce:
Changes to allow users to specify a path to save/load experiments within a Databricks workspace as Databricks requires experiment names to be an absolute path, e.g. "/Users/{username}/my-experiment"


MartijnCa (approved):
Nice improvement Rodalyn! Now that the MLFLow serializer is compatible with Databricks, OpenSTEF has become more accessible to use for people who rely on Databricks.

I took the liberty to make a small commit to consolidate the logic and execute it once at initialization of the serializer. 

I wish you happy holidays!

PR #491, opened 2024-01-03
Update python-upload-package.yaml

FrankKr:
Uses improved method for publishing to pypi.

For future reference: https://docs.pypi.org/trusted-publishers/using-a-publisher/

->
The year change from 2023 to 2024 made it so that the feature engineering of holidays broke down.
Turns out that that part of the codebase is very messy and poorly (improperly) tested.
I made a story for improving that part seperately:
https://alliander.atlassian.net/browse/KTPS-3375


bartpleiter (approved):
Good change to use gh action instead of twine

PR #492, opened 2024-01-10
New dazls updates

ThomasNijsen:
Added a new DAZLs model (with license), and updated the DAZLs model class with a more elaborate string method for model checking.


.gitattributes
```diff
@@ -0,0 +1 @@
+*.sav filter=lfs diff=lfs merge=lfs -text
```
wfstoel, line 1: This file is probably not needed anymore since we are not using lfs, right?

openstef/pipeline/create_component_forecast.py
```diff
@@ -53,12 +55,20 @@ def create_input(
...
+    input_df["sem0"] = input_df["total_load"].sem()
     input_df["sem1"] = input_df["radiation"].sem()
+    input_df["sem2"] = input_df["windspeed_100m"].sem()
+
+    # Features for the new model
+    # Periodic Month feature
+    c = (1 / 11) * np.pi - (1 / 365)
+    n = pd.to_datetime(input_df.index).strftime("%m").tolist()
+    n = np.array([float(x) for x in n])
+    input_df["month_ff"] = np.sin(c * (n - 1))
```
wfstoel, line 71: What prevents you from using n directly from `strftime` (without `.tolist()`) as the input for the sine function?

openstef/pipeline/create_component_forecast.py
```diff
@@ -92,9 +102,29 @@ def create_components_forecast_pipeline(
...
+            DAZLS_STORED + "adaptation_model_scaler.z"
+        )
+        dazls_model.adaptation_model_input_columns = joblib.load(
+            DAZLS_STORED + "adaptation_model_features.z"
+        )
+
+        dazls_model.target_columns = joblib.load(DAZLS_STORED + "target.z")
+        dazls_model.target_scaler = joblib.load(DAZLS_STORED + "target_scaler.z")
+
+        print(dazls_model)
```
wfstoel, line 127: It is not a good practice to directly print strings. If you think it is beneficial to log certain information, you should use a logger with a static string. You can add key-value pairs to add dynamic information to the log message. E.g.:
```
import structlog

logger = structlog.get_logger(__name__)

logger.info("DAZLS model loaded", dazls_model=str(dazls_model))
```

openstef/data/dazls_stored_new_adaptation_model.z
wfstoel, line 1: Currently you are using the string `new` to indicate that this is a newly trained DAZLS model. It is better practice to give a proper version to the model. My preference would be to use a folder for this. So the final structure would look like:
`openstef/data/dazls/v3.4.0/*`
Then you can put all the files in this folder.

Do not forget to change the import path in the code after making this change!

openstef/data/dazls_stored_new.sav.license
wfstoel, line 1: Every file that cannot have a header in the file itself should have a licence file similar to this. So you need to add a licence file for every `.z` file with the exact same name with the `.licence` suffix added to it.

Since the `dazls_stored_new.save` file no longer exists, you can remove this specific licence file.

wfstoel (changes requested):
Nice job adding a new and better DAZLS model! I left some remarks for improvements regarding best practices.

wfstoel (approved):
Nice!

PR #495, opened 2024-01-15
Fixed replacing of repeated values.

bartpleiter:


test/unit/preprocessing/test_preprocessing.py
```diff
@@ -2,55 +2,45 @@
...
-            df, 5, "Column2"
+        # Act
+        df_no_repeated_sequences = preprocessing.replace_repeated_values_with_nan(
+            df, 2, "Column1"
         )
-        self.assertFalse(df_no_repeated.isna().values.any())
 
-        # No change, since there are 4 seq. numbers.
-        df_no_repeated = preprocessing.replace_repeated_values_with_nan(
+        df_repeated_sequences_below_threshold = (
```
wfstoel, line 30: Tiny optional improvement would be to split this unit test into two. In principle a single unit test should only test a single test case. Thus the first unit test can test that a sequence with repetitions indeed removes them properly. A second unit test can test that a sequence without repetitions is not changed.

wfstoel (approved):
Nice change! I like the easy-to-read code.

PR #497, opened 2024-01-16
Updated XGBoost version, added check for early stopping.

bartpleiter:


JonitaRuiter (approved):
Nice upgrade of XGBoost

PR #499, opened 2024-01-18
New dazls updates with version and package updates

ThomasNijsen:


setup.py
```diff
@@ -43,7 +43,12 @@ def read_long_description_from_readme():
     # for more information
     package_data={
         # Include anything in the data directory
-        "openstef": ["data/*", "data/dazls_model_3.4.0/*", "*.license"]
+        "openstef": [
+            "data/*",
+            "data/dazls_model_3.4.0/*",
+            "data/dazls_model_3.2.49/*",
```
wfstoel, line 49: This model we do not want to include as it is not required on pypi. The unit test will still be able to access the model.

wfstoel (approved):

PR #503, opened 2024-02-01
New dazls updates

ThomasNijsen:


openstef/model/regressors/dazls.py
```diff
@@ -103,14 +103,20 @@ def fit(self, features, target):
         )
         self.adaptation_model.fit(adaptation_model_input, y_train)
 
-    def predict(self, x: np.array):
+    def predict(self, x: np.array, return_sub_preds=False):
```
wfstoel, line 106: Nice idea to add this argument!

Could you add a type_hint: `return_sub_preds: bool = False`.

wfstoel (changes requested):
Nice to see another iteration of the DAZLS model.

Could you remove all older versions of the DAZLS model from the git repo? The existence of these models makes the repo quite large.

wfstoel (approved):
LGTM!

PR #505, opened 2024-02-01
Pandas bug patch

ThomasNijsen:


.github/workflows/docs-check.yaml
bartpleiter, line 1: Good changes! Also good suggestions, although I would suggest to move the comments to a confluence page to keep the code clean.

bartpleiter (changes requested):
Good fix! See comment for suggestion.

bartpleiter (approved):
LGTM

PR #513, opened 2024-03-11
Added cyclic temporal feature based on sin/cosine of the time of the day.

egordm:
Creates a new feature of based on sine and cosine of the time of the day. 

temporal_features


FrankKr (approved):
Looks good to me!
You could consider updating the expected_result dataframes to include the New features.
And very small thing, perhaps the name `time_of_day_sine` is a little bit easier to see what the feature is about when te name is abbreviated in de feature-importance-plots then when it starts with sine/cosine

egordm:
Does not have as much effect on the forecasts as hoped after evaluation. I am closing this.

Issue #278, opened 2022-01-05
[chore] Proloaf warning should only be raised if appropriate


FrankKr:
https://github.com/OpenSTEF/openstef/blob/c5ce486dd32a556e7451af06b2e3b30206d58087/openstef/model/model_creator.py#L16-L22

The warning that the optional proloaf dependency is not installed should only be raised if it is required. Currently it is always logged as a warning, causing for unneeded concern:
![image](https://user-images.githubusercontent.com/18208480/148240421-386eaf07-fee1-42e9-b899-c1ee246a3806.png)

egordm:
Closing this issue. The pr already removed all the references and in openstef 4 we don't have such regressor anymore. If the need arises for one, the a followup should be opened.

PR #517, opened 2024-03-27
remove all proloaf references

FrankKr:
Signed-off-by: FrankKr <frank.kreuwel@alliander.com>

**Fixes issue**: #278 by removing all proloaf references since this model is not used and requires a lot of custom fixes.


JonitaRuiter (approved):
LGTM!

PR #519, opened 2024-03-27
Improve readme

wfstoel:


JonitaRuiter (commented):
Looks good to me, but let's see if we can fix the 'failing' action badges before removing all of them.

clara-de-smet (approved):

PR #522, opened 2024-03-27
Refactored column names of dataframes

clara-de-smet:
Mostly related to weather dataframes


openstef/app_settings.py
```diff
@@ -9,9 +9,38 @@
...
+    weather_column_name_windspeed_100m_extrapolated: str = "windspeed_100mExtrapolated"
+    weather_column_name_saturation_pressure: str = "saturation_pressure"
+    weather_column_name_vapour_pressure: str = "vapour_pressure"
+    weather_column_name_dewpoint: str = "dewpoint"
+    weather_column_name_air_density: str = "air_density"
+    weather_column_name_humidity: str = "humidity"
+    weather_column_name_pressure: str = "pressure"
+    weather_column_name_wind_power_fit_extrapolated: str = "windPowerFit_extrapolated"
+    weather_column_name_wind_power_fit_harm_arome: str = "windpowerFit_harm_arome"
+    weather_column_name_wind_hub_height: str = "hub_height"
```
FrankKr, line 30: These are derived features - I don't there is a usecase where a users want to specify something else?

FrankKr, line 30: Perhaps instead of specifying it in the settings, OpenSTEF can just be very clear to the user how it expects certain (optional) input that it needs?

openstef/app_settings.py
```diff
@@ -9,9 +9,38 @@
...
+
+    weather_column_name_turbine_type: str = "turbine_type"
+    weather_column_name_number_turbines: str = "n_turbines"
```
FrankKr, line 33: This is not really 'weather' data?

openstef/app_settings.py
```diff
@@ -9,9 +9,38 @@
...
+
+    location_column_name_latitude: str = "lat"
+    location_column_name_longitude: str = "lon"
```
FrankKr, line 36: For me this is more explicit than needed. Lat and Lon are pretty common as-is, so as long as OpenSTEF is clear it needs this, I don't think its needs to be a custom setting? 
Also, are these terms not already defined in the Prediction Job the user needs to input?

openstef/app_settings.py
```diff
@@ -9,9 +9,38 @@
...
+
+    forecast_column_name_pid: str = "pid"
+    forecast_column_name_customer: str = "customer"
+    forecast_column_name_description: str = "description"
+    forecast_column_name_type: str = "type"
+    forecast_column_name_general_type: str = "algtype"
+    forecast_column_name_horizon_minutes: str = "horizon_minutes"
```
FrankKr, line 43: The user can rename these themselves from the output of OpenSTEF if they'd desire in my opinion

PR #527, opened 2024-03-27
General code improvements

wfstoel:


openstef/model/regressors/custom_regressor.py
```diff
@@ -28,7 +28,7 @@ def valid_kwargs() -> list[str]:
 
     @classmethod
     @abstractmethod
-    def objective(self) -> Type[RegressorObjective]:
+    def objective() -> Type[RegressorObjective]:
```
egordm, line 31: Should this be staticmethod?
Classmethod takes cls as argument

egordm (commented):
Two small sugggestions

openstef/model/regressors/custom_regressor.py
wfstoel: Good catch.

Jonasvdbo:
@bartpleiter and @daanvanes, could you please provide an update on the status of this PR? It seems to have been open for quite some time now.

PR #529, opened 2024-03-29
KTP-762 Implemented Linear Quantile Regression Model

egordm:
Following changes/features were added:
* New LinearQuantile model. Used by specifying model="linear_quantile" in PredictionJob
* Added feature filtering to ignore features deemed irrelevant.
* Added data imputer transform to replace nans in data as linear model can not work with these.
* Added tests for Linear Quantile Model.

Notes:
* The model training quite a bit slower than XGBoost Quantile.
* For each quantile a new SKLearn QuantileRegression model is trained.
* Adding more features significantly impacts training time.

Here are some insights from the comparison notebook:

### Feature Importance
Feature importance is more balanced now, possibly also because of the feature filtering:
image

image

Model also sees that AMI, AZI etc features make a nice base signal.

### Forecasting
* Model follows the load target more closely.
* The quantiles do not cross. 
* The biggest quantile (0.9 in this image is better at forecasting peaks.
* There are more false positives.
image

### DIFF
--- a/openstef/feature_engineering/missing_values_transformer.py
+++ b/openstef/feature_engineering/missing_values_transformer.py
@@ -1,0 +1,69 @@
+from typing import Union, List
+
+import numpy as np
+import pandas as pd
+from sklearn.impute import SimpleImputer
+from sklearn.preprocessing import FunctionTransformer
+from sklearn.utils.validation import check_array
+
+
+class MissingValuesTransformer:
+    in_feature_names: List[str] = None
+    _n_in_features: int = None
+
+    non_null_feature_names: List[str] = None
+
+    def __init__(
+            self,
+            missing_values: Union[int, float, str, None] = np.nan,
+            imputation_strategy: str = None,
+            fill_value: Union[str, int, float] = None,
+    ):
+        """Initialize missing values handler."""
+        self.missing_values = missing_values
+        self.imputation_strategy = imputation_strategy
+        self.fill_value = fill_value
+
+    def fit(self, x, y=None):
+        _ = check_array(x, force_all_finite="allow-nan")
+        if type(x) != pd.DataFrame:
+            x = pd.DataFrame(np.asarray(x))
+
+        self.in_feature_names = list(x.columns)
+        self._n_in_features = x.shape[1]
+
+        # Remove always null columns
+        is_column_null = x.isnull().all(axis="index")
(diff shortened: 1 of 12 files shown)


openstef/feature_engineering/missing_values_transformer.py
```diff
@@ -0,0 +1,72 @@
...
+
+import numpy as np
+import pandas as pd
+from sklearn.impute import SimpleImputer
+from sklearn.preprocessing import FunctionTransformer
+from sklearn.utils.validation import check_array
+
+
+class MissingValuesTransformer:
+    in_feature_names: List[str] = None
```
FrankKr, line 14: Maybe a small docstring where someone can use this class for would be helpful:)

openstef/model/regressors/linear_quantile.py
```diff
@@ -0,0 +1,236 @@
...
+        solver: str = "highs",
+        missing_values: Union[int, float, str, None] = np.nan,
+        imputation_strategy: str = None,
+        fill_value: Union[str, int, float] = None,
+    ):
+        """Initialize LinearQuantileOpenstfRegressor.
+
+        Model that provides quantile regression with SKLearn QuantileRegressor.
+        For each desired quantile an QuantileRegressor model is trained,
+        these can later be used to predict quantiles.
```
FrankKr, line 55: Perhaps nice to add that certain features are ignored

openstef/pipeline/train_model.py
```diff
@@ -450,6 +450,7 @@ def train_pipeline_step_train_model(
     model = ModelCreator.create_model(
         pj["model"],
         quantiles=pj["quantiles"],
+        **(pj.model_kwargs or {}),
```
FrankKr, line 453: nice addiditon

FrankKr (approved):
Looks good Egor!
Two minor comments. 
And if you are feeling clean-code-minded, you could solve the issues SonarCloud lists:
https://sonarcloud.io/project/issues?id=OpenSTEF_openstef&pullRequest=529&resolved=false&inNewCodePeriod=true

egordm:
Reevaluated benchmark for the linear model after recent change (code style fixes). And change in y scaling:

image

openstef/feature_engineering/missing_values_transformer.py
egordm: Nice suggestions. Fixed!

PR #533, opened 2024-04-23
feature(KTP-767): Implemented XGBoost Multioutput Arctan Loss

egordm:
Reference implementation: https://github.com/LaurensSluyterman/XGBoost_quantile_regression/tree/master

Improvements:
* Shorter training time
* Better quantiles
* Similar performance as xgboost_quantile

image

image

### DIFF
--- a/openstef/metrics/metrics.py
+++ b/openstef/metrics/metrics.py
@@ -453,6 +453,7 @@ def arctan_loss(y_true, y_pred, taus, s=0.1):
     Returns:
         grad: An array containing the (negative) gradients with respect to y_pred.
         hess: An array containing the second derivative with respect to y_pred.
+
     """
     size = len(y_true)
     n_dim = len(taus)  # The number of columns
--- a/openstef/metrics/metrics.py
+++ b/openstef/metrics/metrics.py
@@ -431,3 +431,53 @@ def xgb_quantile_obj(
     hess = np.ones_like(preds)
 
     return grad, hess
+
+
+def arctan_loss(y_true, y_pred, taus, s=0.1):
+    """Compute the arctan pinball loss.
+
+    Note that XGBoost outputs the predictions in a slightly peculiar manner.
+    Suppose we have 100 data points and we predict 10 quantiles. The predictions
+    will be an array of size (1000 x 1). We first resize this to a (100x10) array
+    where each row corresponds to the 10 predicted quantile for a single data
+    point. We then use a for-loop (over the 10 columns) to calculate the gradients
+    and second derivatives. Legibility was chosen over efficiency. This part
+    can be made more efficient.
+
+    Parameters:
+        y_true: An array containing the true observations.
+        y_pred: An array containing the predicted quantiles
+        taus: A list containing the true desired coverage of the quantiles
+        s: A smoothing parameter
+
+    Returns:
+        grad: An array containing the (negative) gradients with respect to y_pred.
+        hess: An array containing the second derivative with respect to y_pred.
+    """
+    size = len(y_true)
+    n_dim = len(taus)  # The number of columns
+    n_rows = size // n_dim
+
+    # Resize the predictions and targets.
+    # Each column corresponds to a quantile, each row to a data point.
+    y_pred = np.reshape(y_pred, (n_rows, n_dim))
+    y_true = np.reshape(y_true, (n_rows, n_dim))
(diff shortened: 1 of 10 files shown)


openstef/metrics/metrics.py
```diff
@@ -431,3 +431,54 @@ def xgb_quantile_obj(
...
+    where each row corresponds to the 10 predicted quantile for a single data
+    point. We then use a for-loop (over the 10 columns) to calculate the gradients
+    and second derivatives. Legibility was chosen over efficiency. This part
+    can be made more efficient.
+
+    Parameters:
+        y_true: An array containing the true observations.
+        y_pred: An array containing the predicted quantiles
+        taus: A list containing the true desired coverage of the quantiles
+        s: A smoothing parameter
```
bartpleiter, line 451:
```suggestion
        y_pred: An array containing the predicted quantiles.
        taus: A list containing the true desired coverage of the quantiles.
        s: A smoothing parameter.
```

openstef/metrics/metrics.py
bartpleiter, line 1: I see some inconsistencies with using a full stop after comments/lines in the docstring. I know it is only a minor detail and was already in the example implementation, but I think it is nice if we are consistent with our full stop/period usage :)

openstef/model/objective.py
```diff
@@ -349,6 +349,36 @@ def get_pruning_callback(self, trial: optuna.trial.FrozenTrial):
         )
 
 
+class XGBMultioutputQuantileRegressorObjective(RegressorObjective):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.model_type = MLModelType.XGB_QUANTILE
+
+    def get_params(self, trial: optuna.trial.FrozenTrial) -> dict:
+        """Get parameters for XGBQuantile Regressor Objective with objective specific parameters.
```
bartpleiter, line 358:
```suggestion
        """Get parameters for XGB Multioutput Quantile Regressor Objective with objective specific parameters.
```

openstef/model/objective.py
bartpleiter, line 1: Nice to add the smoothing parameter to the hyperparameter search as Laurens suggested!

openstef/model/regressors/xgb_multioutput_quantile.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def __init__(
+        self,
+        quantiles: tuple[float, ...] = DEFAULT_QUANTILES,
+        gamma: float = 0.0,
+        colsample_bytree: float = 1.0,
+        subsample: float = 1.0,
+        min_child_weight: int = 0,
+        max_depth: int = 6,
+        learning_rate: float = 0.22,
```
bartpleiter, line 68: Given the last point in the docstring, would'nt it be better to use 0.05 as learning rate?

test/unit/model/regressors/test_xgb_multioutput_quantile.py
```diff
@@ -0,0 +1,103 @@
...
+        with self.assertRaises(ValueError):
+            model = XGBMultiOutputQuantileOpenstfRegressor((0.2, 0.3, 0.5, 0.6, 0.7))
+            model.predict("test_data", quantile=0.8)
+
+    def test_set_params(self):
+        # Check hyperparameters are set correctly and do not cause errors
+
+        model = XGBMultiOutputQuantileOpenstfRegressor((0.2, 0.3, 0.5, 0.6, 0.7))
+
+        hyperparams = {
```
bartpleiter, line 81: Maybe good idea to add a check for the smoothing parameter as well so we know for sure it is being used

bartpleiter (approved):
Very  nice! Made some small suggestions

openstef/model/objective.py
egordm: Good catch

openstef/metrics/metrics.py
egordm: It does look nicer indeed, but in google docstring standard they also mix dot endings and not so it's kinda hard to know or check which is right.

I have changed the args with two more words to use .

test/unit/model/regressors/test_xgb_multioutput_quantile.py
egordm: Nice suggestion.

Issue #514, opened 2024-03-15
ERROR: Failed building wheel for statsmodels


lbbartels:
When I want to install openstef I get the following error:

note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for statsmodels
Failed to build statsmodels
ERROR: Could not build wheels for statsmodels, which is required to install pyproject.toml-based projects

How do I fix this?

With kind regards,
Luuk

clara-de-smet:
Hi Luuk, I ran into the same problem a while ago. The issue for me was that statsmodels requires a C++ redistributable, which pip cannot install. I believe that may be the issue you are encountering as well? You could installing the C++ redistributable yourself or you could try installing statsmodels with conda.

egordm:
As far as I know this this has been fixed in v3 and no longer relevant for v4. If issue persists, consider opening a new one targeting v4.

PR #540, opened 2024-06-18
allow newer statsmodels versions

MartinoMensio:
Hi,
Is it possible to allow more recent versions of statsmodels package?

the 0.14 version has more pre-built wheels, and building the 0.13 ones requires additional libraries, otherwise you get errors such as https://github.com/OpenSTEF/openstef/issues/514

Let me know.

Best,
Martino

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,6 @@ pydantic-settings~=2.3
 pymsteams~=0.2.2   # TODO remove teams dependency
 scikit-learn~=1.3
 scipy~=1.10
-statsmodels~=0.13.5
+statsmodels>=0.13.5
 structlog>=23.1,<25
 xgboost~=2.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -17,6 +17,6 @@ pydantic-settings~=2.3
 pymsteams~=0.2.2   # TODO remove teams dependency
 scikit-learn~=1.3
 scipy~=1.10
-statsmodels~=0.13.5
+statsmodels>=0.13.5
```
egordm, line 20: Sounds good. I don't see a problem with that, but we should still add some constraint to the versioning in case they release a breaking version.

```suggestion
statsmodels>=0.13.5,<1.0.0
```

egordm (changes requested):

MartinoMensio:
Thank you @egordm! I added a commit to address your comment :)

egordm (approved):
Awesome, thanks for the change. Will merge it in a bit together with another change.

PR #542, opened 2024-07-17
fix: Made use of non_null params consistent for linear quantile model so null columns are handled correctly.

egordm:
The issue was that if you supply a null column to linear model that
* Fit function throws error
* Feature weight calculation throws an error

These changes fix it and add unit tests on missing_values_transformer

The second commit fixes current test errors due to changes in dependencies.


bartpleiter (approved):
LGTM!

PR #551, opened 2024-09-27
Numpy 2.0 compatibility

clara-de-smet:


requirements.txt
lschilders, line 1: Also force numpy version to be >2

lschilders (changes requested):

requirements.txt
clara-de-smet: Good catch!

bartpleiter (approved):
Unit tests seem to work again :)

lschilders (approved):

PR #558, opened 2024-09-30
Fallback for component splitting

clara-de-smet:


bartpleiter (approved):
Nice fix with unit test!

PR #561, opened 2024-10-01
Add Apple Silicon installation instructions to README.md

twebr:
If you want to install the openstef package on Apple Silicon (Mac with M1-chip or newer), you can encounter issues with the dependencies, such as xgboost. With this pull request, I add instructions for solving these issues.

Credits to Lars Schilders for discovering these installation problems.


bartpleiter (approved):
LGTM! Did not test because I have no Apple devices though

PR #562, opened 2024-10-02
[fix] disable imputation on future data

lschilders:
In the `MissingValuesTransformer`, data that had missing "future" values, i.e. features with trailing `null` values was imputed with the `imputation_strategy`.  This is not desirable since it does not make sense to impute data when it is not in between two known values. Implemented a pre-processing step to remove rows with any trailing null values. Filtered the same rows in the labels accordingly.


openstef/feature_engineering/missing_values_transformer.py
```diff
@@ -57,6 +58,11 @@ def __init__(
             ).set_output(transform="pandas")
             self.imputer_._validate_params()
 
+    def remove_trailing_null_rows(self, df: pd.DataFrame) -> pd.DataFrame:
+        """Remove rows with trailing null values in a DataFrame."""
+        self.non_trailing_null_rows = ~df.bfill().isnull().any(axis="columns")
+        return df.loc[self.non_trailing_null_rows]
+
```
egordm, line 65: I think that it would be nice to have a helper method to remove trailing null rows.
It would be nice if it was self contained (no refs to self). And private _remove_trailing_...

Also, I think it isn't used anywhere right?

openstef/feature_engineering/missing_values_transformer.py
```diff
@@ -83,17 +94,25 @@ def transform(self, x) -> pd.DataFrame:
...
 
         """
         self.fit(x, y)
-        return self.transform(x)
+
+        if y is not None:
+            y = y.loc[self.non_trailing_null_rows]
+
+        x = self.transform(x)
+        x = x.loc[self.non_trailing_null_rows]
```
egordm, line 114: We probably should using self.non_trailing_null_rows and instead just recalculate non_trailing_null_rows.

Mainly because if you set it to a member variable `non_trailing_null_rows` then it will also be baked into model's pickle checkpoint even though it is not used beyond the fitting process.

openstef/feature_engineering/missing_values_transformer.py
```diff
@@ -69,9 +75,14 @@ def fit(self, x, y=None):
         # Remove always null columns
         is_column_null = x.isnull().all(axis="index")
         self.non_null_feature_names = list(x.columns[~is_column_null])
+        x = x[self.non_null_feature_names]
+
+        # Remove rows with trailing null values
+        self.non_trailing_null_rows = ~x.bfill().isnull().any(axis="columns")
```
egordm, line 81: I believe there are some feature that are null and on which forward imputation should be "okay".
Specifically, I don't think APX feature is ignored. Meaning that when it's used it's always nan for future values.

We should probably use a whitelist for a few features on which forward filling is not okay. For example radiation, windspeed.

image

egordm (changes requested):
Nice idea to put it in the missing values transformer, feels like that it was missing a feature indeed.
I added a few comments to handle a few edge cases it thing we might face.

openstef/feature_engineering/missing_values_transformer.py
lschilders: I added an option to the `MissingValueTransformer`: `no_fill_future_values_features` that defines this whitelist, agree on this setup?

openstef/feature_engineering/missing_values_transformer.py
```diff
@@ -37,11 +38,16 @@ def __init__(
...
+            no_fill_future_values_features: The features for which it does not make sense
+                to fill future values. Rows that contain trailing null values for these
+                features will be removed from the data.
         """
         self.missing_values = missing_values
         self.imputation_strategy = imputation_strategy
         self.fill_value = fill_value
+        if no_fill_future_values_features is None:
+            no_fill_future_values_features = []
+        self.no_fill_future_values_features = no_fill_future_values_features
```
egordm, line 50: A small nitpick, there is a shorthand for this.
```suggestion
        self.no_fill_future_values_features = no_fill_future_values_features or []
```

egordm (changes requested):
It looks clean now. I think there is one more thing I missed in previous review and we can merge.

For non_null_feature_names to be passable, it should be added here:
https://github.com/OpenSTEF/openstef/blob/main/openstef/model/model_creator.py#L118

egordm (approved):
Awesome, approved!

PR #565, opened 2024-10-04
feature(KTP-1279): Changed linear model scaling and improved sample weighting feature.

egordm:
* Minmax scaling doesn't weigh low peaks good enough.
* Using quantile scaling now, where everything above p95 and below p5 has a sample weight of 1.
* Using standard scaling for model features.

### DIFF
--- a/openstef/model/regressors/linear_quantile.py
+++ b/openstef/model/regressors/linear_quantile.py
@@ -206,9 +206,10 @@ class LinearQuantileOpenstfRegressor(OpenstfRegressor, RegressorMixin):
         return self
 
     def _calculate_sample_weights(self, y: np.array):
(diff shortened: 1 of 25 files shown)


openstef/model/regressors/linear_quantile.py
```diff
@@ -191,6 +195,16 @@ def fit(self, x: pd.DataFrame, y: pd.Series, **kwargs) -> RegressorMixin:
 
         return self
 
+    def _calculate_sample_weights(self, y: np.array):
```
clara-de-smet, line 198: Could your document this function?

clara-de-smet (approved):

openstef/model/regressors/linear_quantile.py
egordm: Good one. Added!

PR #569, opened 2024-11-21
Make holiday features configurable

JanMaartenvanDoorn:
- Added possibility to specify electricity bidding zone
- Based on bidding zone use holidays of the country of that bidding zone
- If country is NL (default) Dutch school holidays will be applied for any other country school holidays are not available
- Ran black formatting, this may cause some extra unrelated changes

Black seems to fail but I ran black locally, do you know what i can do to fix this?


clara-de-smet:
Are you using the same black version as used in the PR (23.9.1)? There seems to be a permissions issue regardless, but let's first check the version you're running locally.

clara-de-smet:
@JanMaartenvanDoorn Running black with right version was the answer :) Normally the workflow should be able to push formatting changes into the PR again, but it appears that it doesn't have the right to do so (anymore). I've put it on our backlog!

bartpleiter (approved):
LGTM!

PR #574, opened 2024-12-02
WIP feature(KTP-1470): Added data balancing feature.

egordm:


majidkhoshrou (approved):
lgtm

test/unit/tasks/test_train_model.py
```diff
@@ -46,6 +48,33 @@ def test_create_train_model_task_happy_flow(self, train_model_pipeline_mock):
...
+
+        # Assert
+        self.assertEqual(train_model_pipeline_mock.call_count, 1)
+        self.assertEqual(
+            train_model_pipeline_mock.call_args_list[0][0][0]["id"], self.pj["id"]
+        )
+        self.assertEqual(
+            len(train_model_pipeline_mock.call_args_list[0][0][1]),
+            len(test_data) * 2,
+        )
```
lschilders, line 77: The concatenation of the year-old data and recent data is not really tested here, might be better to check whether the correct portion of data is from the one set and the other portion of data is from the other

lschilders (changes requested):
I would extend the test, for the rest it looks good.

test/unit/tasks/test_train_model.py
egordm: Good point. I have added code to test for data that should be present.

lschilders (approved):
Nice solution for the test!

PR #577, opened 2024-12-06
Update black github action to make review suggestions

bartpleiter:
Ready for review&merge. 
Old behavior: new commit when black fails, new commit does not trigger build actions.
New behavior: pr review suggestions are added using black diffs when black fails. Will prevent merging as comments need to be resolved. User can either accept the suggestions from the PR, or run black locally instead with a new commit.

### DIFF
--- a/.github/workflows/black-format-code.yml
+++ b/.github/workflows/black-format-code.yml
@@ -27,32 +27,15 @@ jobs:
       contents: write
 
     steps:
-      # Checkout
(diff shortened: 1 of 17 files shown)


clara-de-smet (approved):
Nice improvement on the github action!

sgatto (commented):
Its very neat. It's simpler than before.  
I'm glad the solution worked.

PR #578, opened 2024-12-06
Replace apx with more general name

JanMaartenvanDoorn:
- removed APX
- ensured backwards compatibility

PR #580, opened 2024-12-12
feature(KTP-1459): Added feature clipping support to limit extrapolation on configured features such as APX.

egordm:


openstef/model/regressors/linear_quantile.py
```diff
@@ -89,6 +92,9 @@ def __init__(
                 "Cannot train quantile model as 0.5 is not in requested quantiles!"
             )
 
+        if clipped_features is None:
+            clipped_features = ["APX"]
```
lschilders, line 96: We need to be wary of https://github.com/OpenSTEF/openstef/pull/578 which changes "APX" everywhere, so we woud need to do this here as well

openstef/model/regressors/linear_quantile.py
```diff
@@ -177,6 +184,7 @@ def fit(self, x: pd.DataFrame, y: pd.Series, **kwargs) -> RegressorMixin:
             y = pd.Series(np.asarray(y), name="load")
 
         x = self._remove_ignored_features(x)
+        x = self.feature_clipper_.fit_transform(x)
```
lschilders, line 187: Why would we `fit_transform` here rather than just `fit`. I know transform doesn't change anything here, but then why would we add it?

lschilders (commented):
Very neat and incredibly fast implementation!

openstef/model/regressors/linear_quantile.py
egordm: Good one. Fixed, transforming during training indeed doesn't do anything.

openstef/model/regressors/linear_quantile.py
egordm: Good to know. Once this is merged I will push an addition to that PR.

lschilders (approved):

PR #585, opened 2025-01-30
feat: added gblinear model as regressor.

bartpleiter:


egordm (approved):
Overall looks good to me!
I notice in GBLinear we have a lot of duplicate logic. I wonder if we should bundle up the preprocessing logic, or we should wait with that until we have a more formal model data prep pipeline.

PR #592, opened 2025-02-17
remove T-1D and T-7D from TO_KEEP_FEATURES in gblinear_quantile

lschilders:


egordm (approved):
Looks good. T-7d might still be usable. So we might want to keep it if it adds value.

PR #593, opened 2025-02-18
fix calculation of rolling aggregates by removing NaNs from window

lschilders:
* rolling window aggregate is now forward filled for where the variable is unknown, such that forecasts further into the future can use that feature as well


egordm (approved):
Nice, looks good!

PR #594, opened 2025-02-25
updated dataclasses to pydantic v2 models

JanMaartenvanDoorn:
- updated pydantic v1 models to v2 models
- added license because REUSE complained it was missing


LICENSES/CC-BY-2.5.txt
JanMaartenvanDoorn, line 1: Added this license file because REUSE complained it was missing

lschilders: This issue is now fixed in https://github.com/OpenSTEF/openstef/pull/595. Let's merge that one first and then this license can be deleted again.

openstef/data_classes/prediction_job.py
lschilders, line 1: "It is also used to determine the holidays that should be used. Currently only ENTSO-E bidding zones are supported." is removed from the description of `electricity_bidding_zone`. It is still relevant, right? At least the first sentence.

lschilders (commented):
Nice that we do not use v1 interchangeably with v2 anymore!

openstef/data_classes/prediction_job.py
JanMaartenvanDoorn: yes is still relevant, I will add it.

lschilders (approved):

PR #596, opened 2025-02-26
fix: compute feature importance GBLinear with absolute weight values

lschilders:


bartpleiter (approved):
Nice to add a test for this!

PR #597, opened 2025-02-26
Add possibility to install cpu only version of xgboost

JanMaartenvanDoorn:
- added possibility to install cpu only version of xgboost as the standard version of XGboost installs a lot of large packages, this will help keeping docker images small


lschilders (changes requested):
When I run `pip install openstef[cpu]` I get:
`ERROR: No matching distribution found for xgboost-cpu~=2.0; extra == "cpu"`

README.md
```diff
@@ -64,19 +72,31 @@ If you want to install the `openstef` package on Apple Silicon (Mac with M1-chip
...
 mkdir -p /usr/local/opt/libomp/
 ln -s /opt/brew/Cellar/libomp/{your_version}/lib /usr/local/opt/libomp/lib
 ```
+
 3. Uninstall `xgboost` with `pip` (`pip uninstall xgboost`) and install with `conda-forge` (`conda install -c conda-forge xgboost`)
 4. If you encounter similar issues with `lightgbm`: uninstall `lightgbm` with `pip` (`pip uninstall lightgbm`) and install later version with `conda-forge` (`conda install -c conda-forge 'lightgbm>=4.2.0'`)
 
+### InsRemark regarding installation with minimal Xgboost dependency
+
+It is possible to install openSTEF with a minimal XGboost (CPU-only) package. This only works on x86_64 (amd64) Linux and Windows platforms. Advantage is that significantly smaller dependencies are installed. In that case run:
```
lschilders, line 86:
```suggestion
### Remark regarding installation with minimal XGBoost dependency

It is possible to install openSTEF with a minimal XGBoost (CPU-only) package. This only works on x86_64 (amd64) Linux and Windows platforms. Advantage is that significantly smaller dependencies are installed. In that case run:
```

bartpleiter (approved):
Nice size reduction!

lschilders (approved):

PR #598, opened 2025-03-07
Only use old model for model specs and score checking if ignore_existing_models is set to False

JanMaartenvanDoorn:
- added extra option to model training that will ignore any existing models and will just train a new model.
- added unit test for this new fucntionality

### DIFF
--- a/openstef/pipeline/train_model.py
+++ b/openstef/pipeline/train_model.py
@@ -46,7 +46,7 @@ def train_model_pipeline(
     check_old_model_age: bool,
     mlflow_tracking_uri: str,
     artifact_folder: str,
-    start_with_new_model: bool = False,
+    ignore_existing_models: bool = False,
 ) -> Optional[tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]]:
     """Middle level pipeline that takes care of all persistent storage dependencies.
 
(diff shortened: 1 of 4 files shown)


openstef/pipeline/train_model.py
```diff
@@ -46,6 +46,7 @@ def train_model_pipeline(
     check_old_model_age: bool,
     mlflow_tracking_uri: str,
     artifact_folder: str,
+    start_with_new_model: bool = False,
```
bartpleiter, line 49: For me the argument name is a bit confusing. What do you think of "ignore_existing_models"?

test/unit/pipeline/test_pipeline_train_model.py
```diff
@@ -564,6 +564,51 @@ def test_train_model_No_old_model(self, serializer_mock, save_model_mock):
         )
         self.assertEqual(len(serializer_mock_instance.method_calls), 3)
 
+    @patch("openstef.model.serializer.MLflowSerializer.save_model")
+    @patch("openstef.pipeline.train_model.MLflowSerializer")
+    def test_train_model_start_with_new_model(self, serializer_mock, save_model_mock):
+        # Mock an old model which is better than the new one.
+
```
bartpleiter, line 571: looks like a copy paste leftover ;)

bartpleiter (approved):
Nice improvement and test! Small suggestion on a different variable name, but otherwise looks good

JanMaartenvanDoorn:
Thanks @Godelaine  and @bartpleiter!

PR #606, opened 2025-04-11
Bump lightgbm to 4.0

MvLieshout:


openstef/model/objective.py
```diff
@@ -130,26 +132,29 @@ def __call__(
...
+        if pruning_callback is not None:
+            callbacks.append(pruning_callback)
 
         # validation_0 and validation_1 are available
         self.model.fit(
             train_x,
             train_y,
             eval_set=eval_set,
-            early_stopping_rounds=EARLY_STOPPING_ROUNDS,
-            verbose=self.verbose,
```
MvLieshout, line 152: Not actually sure we want to remove this verbose option for the XGBoost calls, but LGBM does not accept it in v4.0 and higher.

openstef/model/objective.py
```diff
@@ -272,6 +280,9 @@ def get_pruning_callback(self, trial: optuna.trial.FrozenTrial):
             trial, observation_key=f"validation_1-{self.eval_metric}"
         )
 
+    def get_early_stopping_callback(self):
+        return EarlyStopping(rounds=EARLY_STOPPING_ROUNDS, metric_name=self.eval_metric, data_name=f"validation_1", maximize=False, save_best=True)
```
MvLieshout, line 284: Do we always want to maximize?

openstef/model/objective.py
```diff
@@ -130,28 +132,37 @@ def __call__(
...
         # validation_0 and validation_1 are available
         self.model.fit(
             train_x,
             train_y,
             eval_set=eval_set,
-            early_stopping_rounds=EARLY_STOPPING_ROUNDS,
-            verbose=self.verbose,
             eval_metric=self.eval_metric,
             callbacks=callbacks,
+            **fit_kwargs,
```
MvLieshout, line 165: `early_stopping_rounds` and `verbose` is not an argument of `fit()` for LGBM>=v4.0. For newer versions of XGB, `early_stopping_rounds` is also deprecated. 
As an alternative for `early_stopping_rounds`, we can use a  callback.
As for `verbose`, I only removed it as argument for the LGB model.

openstef/model/objective.py
```diff
@@ -272,6 +286,15 @@ def get_pruning_callback(self, trial: optuna.trial.FrozenTrial):
             trial, observation_key=f"validation_1-{self.eval_metric}"
         )
 
+    def get_early_stopping_callback(self):
+        return EarlyStopping(
+            rounds=EARLY_STOPPING_ROUNDS,
+            metric_name=self.eval_metric,
+            data_name=f"validation_1",
+            maximize=False,
```
MvLieshout, line 294: I am not sure minimisation is always wanted.

openstef/model/objective.py
```diff
@@ -272,6 +288,15 @@ def get_pruning_callback(self, trial: optuna.trial.FrozenTrial):
...
```
MvLieshout, line 296: Not sure if this should always be false...

test/unit/model/test_objective.py
```diff
@@ -29,7 +29,7 @@
 
 class TestRegressorObjective(BaseTestCase):
     def test_call(self):
-        model_type = "xgb"
+        model_type = "flatliner"
```
MvLieshout, line 32: If I keep using "xgb" here, I run into issues where the model type is not defined for the RegressorObjective.

clara-de-smet (approved):

PR #608, opened 2025-04-16
Rename openstf to openstef

JanMaartenvanDoorn:
- rename OpenstfRegressor to OpenstefRegressor

@bartpleiter Be sure to check if this causes trouble when loading models generated with older versions


JanMaartenvanDoorn:
> Looks good, however the unit tests are failing with an indication that there are still places with openstf
@bartpleiter 
Yes, I looked into this a bit deeper, models that are saved with older/current versions of openstef have still the openstf regressor. Renaming this would mean retraining all existing models which is quite painfull. 

Lets wait with this and take this as a thing we fix in openSTEF v4. In that case we can make clear that retraining is required. I close this PR for now.

PR #609, opened 2025-04-16
Feature: Add median regressor

ylvab:
Add new type of regressor where the prediction is the median over a window of time lags.

### DIFF
--- a/openstef/model/regressors/median.py
+++ b/openstef/model/regressors/median.py
@@ -34,7 +34,7 @@ class MedianRegressor(OpenstfRegressor, RegressorMixin):
             The list of feature names
 
         """
-        check_is_fitted(self)
+        check_is_fitted(self, "feature_names_")
         return self.feature_names_
 
     @property
@@ -53,6 +53,7 @@ class MedianRegressor(OpenstfRegressor, RegressorMixin):
             If any lag feature is NaN, this will be ignored.
             If all lag features are NaN, the regressor will return NaN.
         """
+
         lag_df = x.loc[:, self.feature_names]
         median = lag_df.median(axis=1, skipna=True)
 
@@ -71,7 +72,7 @@ class MedianRegressor(OpenstfRegressor, RegressorMixin):
         self.feature_importances_ = np.ones(len(self.feature_names_)) / (
             len(self.feature_names_) or 1.0
         )
-        return RegressorMixin()
+        return self
 
     def __sklearn_is_fitted__(self) -> bool:
         return True
--- a/openstef/model/regressors/median.py
+++ b/openstef/model/regressors/median.py
@@ -1,0 +1,77 @@
+"""This module contains the median regressor."""
+
+import numpy as np
+import pandas as pd
+
+from sklearn.base import RegressorMixin
+from sklearn.utils.validation import check_is_fitted
+
+from openstef.model.regressors.regressor import OpenstfRegressor
+
+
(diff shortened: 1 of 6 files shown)


test/unit/pipeline/test_pipeline_train_model.py
```diff
@@ -157,6 +157,11 @@ def _train_model_pipeline_core_happy_flow(self, data_prep_class):
                             function=split_dummy_arima,
                             arguments={},
                         )
+                
+                if model_type == ModelType.MEDIAN:
```
ylvab, line 161: Is this an acceptable way of handling this? Is it desirable that all models should run under these DataPrepClasses? In that case I might need some guidance on how to think about this.

openstef/model/regressors/median.py
```diff
@@ -0,0 +1,77 @@
...
+
+from openstef.model.regressors.regressor import OpenstfRegressor
+
+
+class MedianRegressor(OpenstfRegressor, RegressorMixin):
+    """
+    Median regressor implementing the OpenSTEF regressor API.
+
+    This regressor is good for predicting two types of signals:
+    - Signals with very slow dynamics compared to the sampling rate, possibly
```
JanMaartenvanDoorn, line 17: Good docs and use cases!

openstef/model/regressors/median.py
```diff
@@ -0,0 +1,77 @@
...
+        Which lag features are used is determined by the feature engineering step.
+        """
+        self.feature_names_ = list(x.columns[x.columns.str.startswith("T-")])
+        if len(self.feature_names_) == 0:
+            raise ValueError("No lag features found in the input data.")
+
+        self.feature_importances_ = np.ones(len(self.feature_names_)) / (
+            len(self.feature_names_) or 1.0
+        )
+        return RegressorMixin()
```
JanMaartenvanDoorn, line 74: Wouldnt it be better to return self here? In that case we can save the state, including the featurenames to mlflow. As it is now we will save an empty regressor, possibly with empy featurenames which can cause problems in the inference stage.

JanMaartenvanDoorn (changes requested):
We should look into the RegressorMixIn return

test/unit/model/regressors/test_median.py
```diff
@@ -0,0 +1,130 @@
...
+
+import numpy as np
+import pandas as pd
+
+from openstef.model.regressors.median import MedianRegressor
+
+
+class TestMedianRegressor(BaseTestCase):
+    def setUp(self):
+        self.train_input = TestData.load("reference_sets/307-train-data.csv")
```
clara-de-smet, line 12: I don't see `train_input` being used anywhere. Could you remove it?

test/unit/pipeline/test_pipeline_train_model.py
clara-de-smet: DataPrepClasses are optional, so to me this is acceptable as well.

clara-de-smet (changes requested):
Nice model addition! Just the single comment, but looks good overall.

test/unit/model/regressors/test_median.py
ylvab: You are right, I ended up generating training data instead. Will remove!

JanMaartenvanDoorn (approved):
Looks good!

clara-de-smet (approved):
Looks good!

PR #611, opened 2025-04-28
Make median model autoregressive

ylvab:


openstef/model/regressors/median.py
```diff
@@ -54,10 +54,16 @@ def predict(self, x: pd.DataFrame, **kwargs) -> np.array:
             If all lag features are NaN, the regressor will return NaN.
         """
 
-        lag_df = x.loc[:, self.feature_names]
-        median = lag_df.median(axis=1, skipna=True)
+        lags_t0 = x.loc[0, self.feature_names]
```
ylvab, line 57: What is the features are not evenly spaced and the feature steps do not correspond to a time step. Do we need to do a check between prediction resolution and lag features somewhere?

JanMaartenvanDoorn (approved):
Yes, this looks like what we discussed.

openstef/model/regressors/median.py
JanMaartenvanDoorn: Yes good idea, maybe we should do this check under "training" so that we do not pick impossible lags

openstef/model/regressors/median.py
```diff
@@ -68,6 +122,34 @@ def fit(self, x: pd.DataFrame, y: pd.DataFrame, **kwargs) -> RegressorMixin:
...
+        diffs = np.diff(lags_in_min)
+        unique_diffs = np.unique(diffs)
+        if len(unique_diffs) > 1:
+            raise ValueError(
+                "Lag features are not evenly spaced. "
+                f"Got lags with differences: {unique_diffs} min. "
+                "Please ensure that the lag features are generated correctly."
+            )
+        # Check that the lags are in the right order
+        if not all(np.diff(lags_in_min) > 0):
```
JanMaartenvanDoorn, line 147: Can we make a static method that automatically enforeces a certain order? (e.g. alphabetically)

JanMaartenvanDoorn (commented):
Looks good, robust checks. One sugestion to take care of lag order.

openstef/model/regressors/median.py
```diff
@@ -53,21 +109,81 @@ def predict(self, x: pd.DataFrame, **kwargs) -> np.array:
...
+        if len(feature_names) == 0:
             raise ValueError("No lag features found in the input data.")
+        
+        # Check that the lags are evenly spaced, i.e. T-1, T-2, T-3, ...
+        feature_to_lags_in_min = []
+        for feature in feature_names:
+            if feature.endswith("min"):
+                lag_in_min = int(feature.split("-")[1].split("min")[0])
+            elif feature.endswith("d"):
+                lag_in_min = int(feature.split("-")[1].split("d")[0]) * 60 * 2
```
JanMaartenvanDoorn, line 156:
```suggestion
                lag_in_min = int(feature.split("-")[1].split("d")[0]) * 60 * 24
```

JanMaartenvanDoorn (approved):
Looks good!

JanMaartenvanDoorn:
@clara-de-smet @bartpleiter We found some issues with the median model. Making it autoregressive solve these issues. Please have another look.

lschilders (commented):
Very well documented! I'd suggest to split some isolated functionality from the `fit` and `predict` methods to separate (static)methods, since the methods are quite large now. For example the autoregressive step could be a separate method.

ylvab:
> Very well documented! I'd suggest to split some isolated functionality from the `fit` and `predict` methods to separate (static)methods, since the methods are quite large now. For example the autoregressive step could be a separate method.

Thank you for the review! I have broken the autoregressive step out to a static method, as you suggested, and also moved much of the validation in fit into a static method. 

I cannot break into several static methods, since they would depend on each other and no longer be static. It would be possible to put parts as functions outside the class, but I'm not sure that would improve readability.

lschilders (approved):
Good improvement!

PR #614, opened 2025-05-02
Pin run_id in prediciton job

JanMaartenvanDoorn:
- added possibility to pin a certain run_id in the prediction job, this makes it easier to locally itterate on models.
- updated unit tests

### DIFF
--- a/openstef/model/serializer.py
+++ b/openstef/model/serializer.py
@@ -18,6 +18,7 @@ from mlflow.store.artifact.artifact_repository_registry import get_artifact_repo
 from xgboost import XGBModel  # Temporary for backward compatibility
 
 from openstef.data_classes.model_specifications import ModelSpecificationDataClass
+from openstef.data_classes.prediction_job import PredictionJobDataClass
 from openstef.logging.logger_factory import get_logger
 from openstef.metrics.reporter import Report
 from openstef.model.regressors.regressor import OpenstfRegressor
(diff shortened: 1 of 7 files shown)


openstef/model/serializer.py
```diff
@@ -142,20 +143,31 @@ def _log_figures_with_mlflow(self, report) -> None:
 
     def load_model(
         self,
-        experiment_name: str,
+        pj: PredictionJobDataClass,
```
ylvab, line 146: This is not very backward compatible. Would it be better to add the run_id as optional?``
```suggestion
        experiment_name: str,
        model_run_id: typing.Optional[str] = None,
```

ylvab (commented):
What is the standard for backward compatibility in these cases?

openstef/model/serializer.py
JanMaartenvanDoorn: yes good point, now changed this.

openstef/model/serializer.py
```diff
@@ -143,20 +144,29 @@ def _log_figures_with_mlflow(self, report) -> None:
...
-        """Load sklearn compatible model from MLFlow.
+        """ Load an sklearn-compatible model from MLflow.
 
-        Args:
-            experiment_name: Name of the experiment, often the id of the predition job.
+        This method retrieves a trained model and its specifications from MLflow 
+        based on the provided PredictionJobDataClass instance. It supports loading 
+        a specific model run if a run number is provided.
+
+            pj (PredictionJobDataClass): The prediction job data containing the
```
ylvab, line 155: The docstring no longer reflects the api, it needs to be updated.

openstef/pipeline/train_model.py
```diff
@@ -320,7 +320,7 @@ def train_pipeline_step_load_model(
 
     if not ignore_existing_models:
         try:
-            old_model, model_specs = serializer.load_model(experiment_name=str(pj.id))
+            old_model, model_specs = serializer.load_model(str(pj.id))
             old_model_age = old_model.age  # Age attribute is openstef specific
             return old_model, model_specs, old_model_age
         except (AttributeError, FileNotFoundError, LookupError):
```
ylvab, line 326: This will still take the latest existing model, not the specified model. This might be the intended behavior, but needs to be clear. 
Perhaps add a few lines docuementing ignore_existing_models in the docstring on lines 160-170?

ylvab (approved):
This is very useful as it makes it possible to train several models in a production environment but hold off deployment until the new model has been verified.

I have a few comments on clarifying documentation.

openstef/model/serializer.py
JanMaartenvanDoorn: Good catch, updated this.

openstef/pipeline/train_model.py
JanMaartenvanDoorn: Yes thought it would not be needed here to use a specific run id, but on second thought: it is more consistent to use the same logic as for creating forecasts. Now updated the code

openstef/model/serializer.py
lschilders: The docstring still does not specify the correct function args

openstef/pipeline/train_model.py
lschilders: Can you still add docs on `ignore_existing_models`?

JanMaartenvanDoorn: Yes added these

openstef/model/serializer.py
JanMaartenvanDoorn: Now it should

PR #615, opened 2025-05-14
fix(STEF-1465): Reset forecast column to P50 after sorting

MvLieshout:


test/unit/pipeline/test_pipeline_create_forecast.py
```diff
@@ -247,3 +249,60 @@ def side_effects(experiment_name):
             self.pj, forecast_data, "./test/unit/trained_models/mlruns"
         )
         self.assertTrue(create_forecast_pipeline_core_mock.called)
+
+
+class TestCreateForecastPipelineQuantileModel(BaseTestCase):
```
MvLieshout, line 254: This test was useful to find the source of the bug since there was no way to test the forecasting pipeline for a quantile model before. Do we want to keep it?

egordm: Good point. I think since the bug was part of sort_quantiles function and the fact that it violated the invariant (forecast column is always equal to P50), it might make sense to create a more minimal test with synthetic data?

egordm (changes requested):
Nice fix and find. I have left a comment regarding the test. If we make it more minimal, we don't have to add so many extra files to the repo.

test/unit/pipeline/test_pipeline_create_forecast.py
MvLieshout: I have actually already added to an existing test which makes sure the P50 column is equal to the forecast column. It can be found here: https://github.com/OpenSTEF/openstef/blob/8d733195ab1ebcc5923616496a64d3d86d17fd34/test/unit/postprocessing/test_postprocessing.py#L154. 

Since that functionality is already tested there, I think it might be kind of superfluous to keep this forecast pipeline model for the quantile model.

openstef/postprocessing/postprocessing.py
```diff
@@ -255,9 +255,12 @@ def sort_quantiles(
...
         return forecast
 
-    # sort the columns
+    # Sort the columns
     p_columns = np.sort(p_columns)
 
     forecast.loc[:, p_columns] = forecast[p_columns].apply(sorted, axis=1).to_list()
 
+    # Set the forecast columun equal to the median
+    forecast["forecast"] = forecast["quantile_P50"]
```
MvLieshout, line 264: I now see that this causes issues in the case that there is no P50. However, I am wondering how realistic that case is.

test/unit/pipeline/test_pipeline_create_forecast.py
egordm: Oh nice, looks like I missed it. Then in that case we already test the most important part. I think this test is optional then. As you said, it might be useful to find further issues in the future, so I have no strong opinion whether to remove it to keep it, actually. So long it doesn't significantly slow down the unit tests then it's okay to keep it.

egordm (approved):

test/unit/pipeline/test_pipeline_create_forecast.py
MvLieshout: It's quite fast, so I will leave it in there.

PR #642, opened 2025-08-15
Update CITATION.cff

FrankKr:
order citation according to DOI on zenodo


CITATION.cff
```diff
@@ -10,28 +10,36 @@ url: "https://github.com/OpenSTEF/openstef"
...
+    email: "Bart.Pleiter@Alliander.com"
+    affiliation: "Alliander"
   - family-names: "Stoel"
     given-names: "Willem Frederik"
     email: "wfstoel@outlook.com"
+  - family-names: "Bogaard"
+    name-particle: "van den"
+    given-names: "Jonas"
+    email: "Jonas.van.den.Bogaard@Alliander.com"
+    affiliation: "Alliander"
```
bartpleiter, line 38: Duplicate affiliation line

bartpleiter:
There seems to be a syntax error as well, but could be because of the duplicate affiliation line
 
 (CITATION.cff) Parser error: while parsing a block mapping at line 5, column 1 did not find expected key at line 21, column 2

Issue #628, opened 2025-07-29
[OpenSTEF 4.0] Move stef-beam into openstef-beam


egordm:
As STEF community
I want to have STEF BEAM
So that I have a way to evaluate and analyze my forecasts

❗Priority (What if we don't do this?/Are there any deadlines? etc.)

Definition of Done:

✅ Acceptance criteria

- [x] Removed STEF specific parts from STEF beam (predefined benchmarks)
- [x] Added STEF beam to the openstef 4.0 branch as a workspace project
- [x] Added packaging of STEF beam
- [x] Fixed possible lint / format / typing issues to pass the new CI/CD quality criteria
- [ ] Improve docstrings / examples and readme for STEF beam

📄 Documentation criteria:

Interfaces with other products and services

Update relevant confluence pages

🧪 Test criteria:

Unit tests written and passed

⌛ Dependencies:

N/A

🚀 Releasing:

N/A

Other information:

🌍 Background

PR #645, opened 2025-08-20
feature(#628): Added openstef core package for datasets and other shared code.

egordm:


packages/openstef-core/src/openstef_core/datasets/versioned_timeseries_dataset.py
```diff
@@ -0,0 +1,288 @@
...
+                Default is 'available_at'.
+
+        Raises:
+            MissingColumnsError: If the required timestamp_column or available_at_column are missing.
+        """
+        missing_columns = {timestamp_column, available_at_column} - set(data.columns)
+        if missing_columns:
+            raise MissingColumnsError(missing_columns=list(missing_columns))
+
+        if not data.attrs.get("is_sorted", False):
```
lschilders, line 102: This does allow for incorrectly setting `is_sorted` is True, leaving the index unsorted. Can we not just check whether it is sorted?

egordm: I figured checking if something is sorted is relatively expensive O(n). So setting an attribute makes it O(1) and much easier than sorting everytime. 

I think it is possible to circumvent it. But similarly you can just do dataset.data and circumvent all the checks.

So unless the normal workflow introduces inconsistencies / breaks invariants it should be fine?

PR #647, opened 2025-08-22
feature(#628): Addition of STEF BEAM

egordm:
Notable changes over 'STEF BEAM':
* Changed `ModelInterface` to be `BacktestForecasterMixin`
* Using OpenSTEF `VersionedTimeSeriesDataset` and `TimeSeriesDataset` over `BaseDataset`, `PandasDataset`, and `pd.DataFrame`
* Removed `jaxtyping` in favor of just annotating array shape in comments
* Renamed `analytics` module to `analysis`
* Moved some types like `LeadTime` and `AvailableAt` to openstef-core
* Added more descriptive docstrings with examples.
* For now not added openstef 3.0 implementations for model interface.
* Also didn't add any benchmarks. Later we will add our OpenSTEF 50 signals benchmark we have released on alliander website.

### DIFF
--- a/packages/openstef-beam/src/openstef_beam/backtesting/backtest_pipeline.py
+++ b/packages/openstef-beam/src/openstef_beam/backtesting/backtest_pipeline.py
@@ -93,7 +93,7 @@ class BacktestPipeline:
         """
         if config.prediction_sample_interval != forecaster.config.predict_sample_interval:
             raise ValueError(
-                "The prediction sample interval of the backtest config must match the model interface's predict sample "
+                "The prediction sample interval of the backtest config must match the forecaster's predict sample "
                 "interval.",
                 config.prediction_sample_interval,
                 forecaster.config.predict_sample_interval,
--- a/packages/openstef-beam/src/openstef_beam/benchmarking/__init__.py
+++ b/packages/openstef-beam/src/openstef_beam/benchmarking/__init__.py
@@ -2,12 +2,12 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
-"""Runs complete model comparison studies across multiple energy consumption targets.
+"""Runs complete model comparison studies across multiple forecasting targets.
 
-Comparing forecasting models properly requires testing them on many different energy
-consumption patterns (different buildings, regions, seasons). This module automates
-the entire process: training models, running backtests, calculating metrics, generating
-reports, and storing results for comparison.
+Comparing forecasting models properly requires testing them on many different forecasting
+scenarios (equipment types, consumption/prosumption, solar/wind parks, regions, seasons).
+This module automates the entire process: training models, running backtests, calculating
+metrics, generating reports, and storing results for comparison.
 
 The complete workflow:
     - Model training: Train different forecasting approaches on each target
--- a/packages/openstef-beam/src/openstef_beam/analysis/__init__.py
+++ b/packages/openstef-beam/src/openstef_beam/analysis/__init__.py
@@ -11,7 +11,7 @@ for energy forecasting operations.
 
 What you get:
     - Performance charts: See how models compare across different metrics
-    - Time series plots: Visualize predictions vs actual consumption over time
+    - Time series plots: Visualize predictions vs actual load over time
     - Error analysis: Understand when and why models make mistakes
     - Comparison reports: Side-by-side model performance analysis
 """
--- a/packages/openstef-beam/src/openstef_beam/benchmarking/benchmark_pipeline.py
+++ b/packages/openstef-beam/src/openstef_beam/benchmarking/benchmark_pipeline.py
@@ -164,10 +164,10 @@ class BenchmarkPipeline[T: BenchmarkTarget, F]:
         1. Gets all available targets from the target provider
         2. Optionally filters them based on provided criteria
         3. Processes each target sequentially or in parallel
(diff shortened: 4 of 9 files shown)


packages/openstef-beam/src/openstef_beam/backtesting/backtest_pipeline.py
```diff
@@ -0,0 +1,301 @@
...
+            config: Backtesting configuration including prediction and training intervals.
+            forecaster: Model implementing the required forecasting interface.
+
+        Raises:
+            ValueError: If the prediction sample intervals don't match between
+                config and forecaster.
+        """
+        if config.prediction_sample_interval != forecaster.config.predict_sample_interval:
+            raise ValueError(
+                "The prediction sample interval of the backtest config must match the model interface's predict sample "
```
lschilders: Bit of a nitpick, but do we want to use "model interface" here (and in some other places) when the naming has been changed? Could be confusing.

packages/openstef-beam/src/openstef_beam/benchmarking/__init__.py
```diff
@@ -0,0 +1,53 @@
+# SPDX-FileCopyrightText: 2025 Contributors to the OpenSTEF project <short.term.energy.forecasts@alliander.com>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+"""Runs complete model comparison studies across multiple energy consumption targets.
+
+Comparing forecasting models properly requires testing them on many different energy
+consumption patterns (different buildings, regions, seasons). This module automates
```
lschilders:
```suggestion
"""Runs complete model comparison studies across multiple energy consumption and production (prosumption) targets.

Comparing forecasting models properly requires testing them on many different energy
prosumption patterns (different buildings, regions, seasons). This module automates
```

packages/openstef-beam/src/openstef_beam/evaluation/metric_providers.py
```diff
@@ -0,0 +1,401 @@
...
+        metrics for all relevant quantiles.
+
+        Args:
+            subset: Evaluation subset containing predictions and ground truth data.
+
+        Returns:
+            QuantileMetricsDict mapping quantile keys to computed metric values.
+        """
+        quantiles = np.array([Quantile.parse(quantile) for quantile in subset.predictions.feature_names])
+        y_true: npt.NDArray[np.floating] = cast(pd.Series, subset.ground_truth.data.squeeze()).to_numpy()
```
lschilders, line 88: Can we not just use `np.ndarray`?

packages/openstef-core/src/openstef_core/datasets/timeseries_dataset.py
lschilders, line 1: I'd go for `from_` & `to_` or `read_` & `write_`, not a mix of both.

lschilders (commented):
Looking great! Good addition and changes compared to STEF BEAM. Just a few (mainly styling) comments.

packages/openstef-beam/src/openstef_beam/backtesting/backtest_pipeline.py
egordm: Good find. No we don't want to use it, I have missed this one during refactor. I will update this.

packages/openstef-beam/src/openstef_beam/benchmarking/__init__.py
egordm: Good one. I have changed the text to be more general. Refer to forecasting targets and forecasting scenarios.

packages/openstef-core/src/openstef_core/datasets/timeseries_dataset.py
egordm: I picked to_ and read_ on purpose to mirror pandas functions. Figured people familiar with it, would be searching for similarly named functions.

* https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_parquet.html
* https://pandas.pydata.org/docs/reference/api/pandas.read_parquet.html

Perhaps we should keep it like this?

packages/openstef-beam/src/openstef_beam/evaluation/metric_providers.py
egordm: `np.ndarray` doesn't support typehinting what dtype an array is.

My inital goal was to replace jaxtyping which does things like Float[Array, "size"]. Unfortunately there is not a 1:1 replacement from numpy. But we can at least hint that it's a float array.

packages/openstef-beam/src/openstef_beam/backtesting/backtest_pipeline.py
egordm: Fixed

packages/openstef-beam/src/openstef_beam/evaluation/metric_providers.py
lschilders: Ah ok nice, did not know that. Will change it on my branch as well then.

packages/openstef-core/src/openstef_core/datasets/timeseries_dataset.py
lschilders: Ah, that is indeed a good reason then.

lschilders (approved):
Great fixes, LGTM!

PR #648, opened 2025-08-26
feature(626): Flatliner, completeness checks, feature dropping & imputation

MvLieshout:


packages/openstef-core/tests/unit/feature_engineering/validation_transforms/test_missing_values_transform.py
```diff
@@ -0,0 +1,212 @@
...
+    # Act
+    transform.fit(sample_dataset)
+    result = transform.transform(sample_dataset)
+    
+    # Assert
+    assert not (result.data == -999.0).any().any()
+    assert result.data.loc[result.data.index[1], "load"] == 110.0 
+    assert result.data.loc[result.data.index[2], "temperature"] == 20.5 
+
+def test_all_null_dataset_with_trailing_removal(sample_dataset: TimeSeriesDataset):
```
MvLieshout, line 203: Not sure if this is the behaviour we want. What do you think?

Issue #626, opened 2025-07-29
[OpenSTEF 4.0] Implement feature engineering


bartpleiter:
As user of OpenSTEF
I want to have feature engineering pipelines implemented in OpenSTEF V4
So that I can use modular feature engineering with any model

❗Priority (What if we don't do this?/Are there any deadlines? etc.)

Definition of Done:

✅ Acceptance criteria

- [x] Create custom [TimeseriesDataset](https://lf-energy.atlassian.net/wiki/spaces/OS/pages/349044781/OpenSTEF+4.0+Architecture+Document#Interfaces) to wrap around input DataFrame and validate it. (columns, timeseries index, timezone)
    - [ ] ~Use [narwals](https://narwhals-dev.github.io/narwhals/basics/dataframe/#__tabbed_1_2) as a dataframe backend, still allow users to use any native dataframe type anywhere.~
- [x] [Feature Transform interface](https://lf-energy.atlassian.net/wiki/spaces/OS/pages/349044781/OpenSTEF+4.0+Architecture+Document#Interfaces) implemented as FeatureTransform implemented in `openstef-models/feature_engineering`
- [x] All currently implemented feature engineering from OpenSTEF v3 are implemented under the correct subfolder as FeatureTransform
   - [x] Holidays (not specific to NL, make configurable) (Lars)
   - [x] Weather features (windspeed, radiation, humidity) (Marnix)
   - [x] Lags (configurable see interfaces) (Egor)
   - [x] Energy domain transforms (pv/wind generation) (Marnix)
   - [x] Normalization (Scaling & Clipping), trend & seasonality (Lars)
   - [x] Cyclic features & Daylight (Lars)
   - [x] Flatliner, completeness checks, feature dropping & imputation (Marnix)
- [x] Feature processing pipeline
- [x] Documented using docstring (with new example in docsting format)

📄 Documentation criteria:

Interfaces with other products and services

Update relevant confluence pages

🧪 Test criteria:

Unit tests written and passed

⌛ Dependencies:

N/A

🚀 Releasing:

N/A

Other information:

🌍 Background

https://lf-energy.atlassian.net/wiki/spaces/OS/pages/334037011/OpenSTEF+4.0+Proposal#Module-Architecture%3A

MvLieshout:
@egordm, what is meant by "Feature dropping"?

PR #649, opened 2025-08-26
feature(#626): Missing value imputation

MvLieshout:


packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,175 @@
...
+
+        Returns:
+            pd.DataFrame: A DataFrame with trailing null rows removed for the specified
+                features. If no features are configured for checking or none of the
+                configured features exist in the data, returns the original data unchanged.
+        """
+        if not self.config.no_fill_future_values_features:
+            return data.data
+
+        features_to_check = [f for f in self.config.no_fill_future_values_features if f in data.data.columns]
```
MvLieshout: Do we want a warning/logging if features are given as input here that do not exist in the data?

Edit: Give back warning

packages/openstef-core/tests/unit/feature_engineering/validation_transforms/test_missing_values_transform.py
```diff
@@ -0,0 +1,216 @@
...
+    sample_dataset.data["load"] = [np.nan, np.nan, np.nan]
+    sample_dataset.data["temperature"] = [20.0, 21.0, 22.0]
+    config = MissingValuesTransformConfig(
+        imputation_strategy=ImputationStrategy.CONSTANT, fill_value=0.0, no_fill_future_values_features=["load"]
+    )
+    transform = MissingValuesTransform(config)
+
+    # Act & Assert
+    # Should raise ValueError when trying to fit on empty dataset after trailing removal
+    with pytest.raises(ValueError, match="Found array with 0 sample"):
```
MvLieshout: This is what it currently does, but is this also what we want?

packages/openstef-core/tests/unit/feature_engineering/validation_transforms/test_missing_values_transform.py
```diff
@@ -0,0 +1,216 @@
...
+@pytest.fixture
+def sample_dataset() -> TimeSeriesDataset:
+    """Create a sample dataset with missing values.
+
+    Returns:
+        TimeSeriesDataset: A sample dataset containing missing values for testing.
+    """
+    data = pd.DataFrame(
+        {
+            "load": [100.0, np.nan, 110.0, 120.0],
```
MvLieshout: Not sure how we handle this now. Do we also impute our target variable?

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,175 @@
...
+    def _remove_trailing_null_rows(self, data: TimeSeriesDataset) -> pd.DataFrame:
+        """Remove trailing rows that contain null values in specified features.
+
+        This method removes rows from the end of the dataset where specified features
+        (defined in config.no_fill_future_values_features) contain null values. This
+        is useful for cleaning datasets where future values should not be filled and
+        trailing nulls indicate incomplete data.
+
+        Args:
+            data (TimeSeriesDataset): The input time series dataset containing the data
```
MvLieshout: Remove type here

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,175 @@
...
+                to be processed.
+
+        Returns:
+            pd.DataFrame: A DataFrame with trailing null rows removed for the specified
```
MvLieshout: Type not needed here

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,175 @@
...
+        """Initialize the MissingValuesTransform with the given configuration."""
+        self.config = config
+
+        self.imputer_: SimpleImputer = SimpleImputer(
+            strategy=self.config.imputation_strategy.value,
+            fill_value=self.config.fill_value,
+            missing_values=self.config.missing_value,
+            keep_empty_features=False,
+        )
+        self.imputer_.set_output(transform="pandas")  # pyright: ignore[reportUnknownMemberType]
```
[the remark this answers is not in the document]
MvLieshout: We enforce `scikit-learn>=1.7.1`, so this is not necessary.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
MvLieshout: Also check other docs

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,175 @@
...
+
+        Args:
+            x: DataFrame to check for trailing null values
+
+        Returns:
+            Boolean Series where True indicates rows to keep (non-trailing null rows)
+        """
+        return ~x.bfill().isna().any(axis="columns")
+
+    def _remove_trailing_null_rows(self, data: TimeSeriesDataset) -> pd.DataFrame:
```
lschilders: I think it is nicer to have the same type as input parameter to the function as the output. I'd change TimeSeriesDataset type to pd.DataFrame and use it accordingly in the fit/transform.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,175 @@
...
+            A DataFrame with trailing null rows removed for the specified
+                features. If no features are configured for checking or none of the
+                configured features exist in the data, returns the original data unchanged.
+        """
+        if not self.config.no_fill_future_values_features:
+            return data.data
+
+        features_to_check = [f for f in self.config.no_fill_future_values_features if f in data.data.columns]
+        if not features_to_check:
+            return data.data
```
lschilders: Don't forget the warning :)

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,175 @@
...
+                The dataset's data attribute should be a pandas DataFrame or similar structure.
+
+        Returns:
+            The transformed TimeSeriesDataset with trailing null rows removed and
+                missing values imputed. The original dataset structure is preserved with
+                only the data attribute modified.
+        """
+        data.data = self._remove_trailing_null_rows(data)
+        data.data = self.imputer_.transform(data.data)  # pyright: ignore[reportAttributeAccessIssue,reportUnknownMemberType]
+        return data
```
lschilders: This modifies the original object. The transform should not be done inplace, so a `TimeSeriesDataset` should be created with new `data`.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,175 @@
...
+    True
+    >>> result.data.shape[1] == 3 # Empty feature column dropped
+    True
+    """
+
+    def __init__(self, config: MissingValuesTransformConfig):
+        """Initialize the MissingValuesTransform with the given configuration."""
+        self.config = config
+
+        self.imputer_: SimpleImputer = SimpleImputer(
```
lschilders: We don't use the underscore postfix anymore, `self.imputer` is preferred.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,175 @@
...
+    ...         "wind_speed": [5, 6, np.nan, np.nan],
+    ...         "empty_feature": [np.nan, np.nan, np.nan, np.nan]
+    ...     },
+    ...     index=pd.date_range("2025-01-01", periods=4, freq="1h"),
+    ... )
+    >>> dataset = TimeSeriesDataset(data, timedelta(hours=1))
+    >>> config = MissingValuesTransformConfig(
+    ...     imputation_strategy=ImputationStrategy.MEAN,
+    ...     missing_value=np.nan,
+    ...     no_fill_future_values_features=["load"]
```
lschilders: Since the `no_fill_future_values_features` should be applied to features only, it's better to also do that in the example, so not using `"load"`

packages/openstef-core/tests/unit/feature_engineering/validation_transforms/test_missing_values_transform.py
lschilders: In OpenSTEF 3.0 it removes columns that are all NaN before this transformation. I think this should be included here as well.

lschilders (changes requested):

packages/openstef-core/tests/unit/feature_engineering/validation_transforms/test_missing_values_transform.py
MvLieshout: The columns that are all NaN are still dropped by the imputer due to the argument `keep_empty_features=False`.

In the case that is tested here, the user actually explicitly states that the filling of trailing NaNs should not be done for this specific feature. Dropping it quietly could be confusing as the user did pick this feature for a reason I would say. 

We could do two things:
1. Drop the feature and raise a warning.
2. Actually keep it as is and raise an error.

I guess this is a question of how strict we want to be. Since we also raise a warning in the other case, I think that might be the most logical solution here too?

In other cases, where there are still rows left or where this feature is not asked for in `no_fill_future_values_features`, there is no issue.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,196 @@
...
+        Returns:
+            A TimeSeriesDataset with empty features removed.
+        """
+        if np.isnan(self.config.missing_value):
+            all_missing_mask = data.data.isna().all()
+        else:
+            all_missing_mask = (data.data == self.config.missing_value).all()
+
+        for col in data.data.columns[all_missing_mask]:
+            _logger.warning("Dropped column '%s' from dataset because it contains only missing values.", col)
```
MvLieshout, line 133: Logging warning to be in line with `versioned_timeseries_dataset.py.

lschilders:
```suggestion
            _logger.warning(f"Dropped column {col} from dataset because it contains only missing values.")
```

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -92,27 +94,37 @@ def __init__(self, config: MissingValuesTransformConfig):
...
             missing_values=self.config.missing_value,
             keep_empty_features=False,
         )
-        self.imputer_.set_output(transform="pandas")  # pyright: ignore[reportUnknownMemberType]
+        self.imputer.set_output(transform="pandas")  # pyright: ignore[reportUnknownMemberType]
 
-    @staticmethod
-    def _determine_trailing_null_rows(x: pd.DataFrame) -> pd.Series:
-        """Determine rows with trailing null values in a DataFrame.
+    def _drop_empty_features(self, data: TimeSeriesDataset) -> TimeSeriesDataset:
```
lschilders: Instead of creating a new TimeSeriesDataset with every operation, I think it is nicer to do all the intermediate operations on `pd.DataFrame` and only create a new `TimeSeriesDataset` at the end of the `transform`. Such that there is only 1 `TimeSeriesDataset` in and 1 out.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
MvLieshout: This is what I had at first, but the linter sadly did not agree 😢

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,197 @@
...
+    >>> result = transform.transform(dataset)
+    >>> int(result.data.isna().sum().sum()) == 0
+    True
+    >>> result.data.shape[0] == 3 # Last row dropped
+    True
+    >>> result.data.shape[1] == 3 # Empty feature column dropped
+    True
+    """
+
+    def __init__(self, config: MissingValuesTransformConfig):
```
egordm: For transforms it might be better to not have a separate config class, but make configuration params a part of the transform itself. If it helps with validation you could extend BaseConfig and TimeSeriesTransform.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,197 @@
...
+        Returns:
+            The transformed TimeSeriesDataset with trailing null rows removed and
+                missing values imputed. The original dataset structure is preserved with
+                only the data attribute modified.
+        """
+        data_cleaned = self._drop_empty_features(data.data)
+        data_cleaned = self._remove_trailing_null_rows(data_cleaned)
+        data_transformed = self.imputer.transform(data_cleaned)  # pyright: ignore[reportUnknownMemberType, reportUnknownVariableType]
+
+        return TimeSeriesDataset(data=data_transformed, sample_interval=data.sample_interval)  # pyright: ignore[reportArgumentType]
```
egordm: Most of the typing issues could be solves with proper / explicit typing. It it doesn't work let me know, I might be able to look with you on this.

egordm (changes requested):
A few nitpicks from me. Otherwise it looks nice.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,217 @@
...
+        description="Value to use when imputation_strategy is CONSTANT",
+    )
+    no_fill_future_values_features: list[str] = Field(
+        default_factory=list,
+        description="List of feature names for which trailing NaN values should not be filled",
+    )
+
+    imputer: SimpleImputer = Field(exclude=True, default_factory=SimpleImputer)
+
+    class Config:
```
MvLieshout: Needed to allow the imputer to be set in init.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
egordm: I believe %s is on purpose. Mainly to avoid formatting unless it will actually be printed. Warning level is on.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,209 @@
...
+                features. If no features are configured for checking or none of the
+                configured features exist in the data, returns the original data unchanged.
+        """
+        if not self.no_fill_future_values_features:
+            return data
+
+        features_to_check = [f for f in self.no_fill_future_values_features if f in data.columns]
+        for feature in self.no_fill_future_values_features:
+            if feature not in data.columns:
+                _logger.warning("Feature '%s' not found in dataset columns.", feature)
```
egordm: I would avoid logging multiple warnings and instead combine them into one.
```suggestion
        missing_columns = set(self.no_fill_future_values_features) - set(data.columns)
        if missing_columns:
            _logger.warning("Features '%s' not found in dataset columns.", missing_columns)
```

egordm: I wonder if we should make it strict instead. Instead of giving a warning throw an error, because the input is invalid. Also fewer edge cases to deal with in the transforms.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
```diff
@@ -0,0 +1,209 @@
...
+    fill_value: float | str | None = Field(
+        default=None,
+        description="Value to use when imputation_strategy is CONSTANT",
+    )
+    no_fill_future_values_features: list[str] = Field(
+        default_factory=list,
+        description="List of feature names for which trailing NaN values should not be filled",
+    )
+
+    imputer: SimpleImputer = Field(exclude=True, default_factory=SimpleImputer)
```
egordm: This seems to be a private member var. Do you think it makes sense to prefix it as `_imputer`?

egordm (approved):
Looks great. A few more nitpicks, but approved in advance. Since once those are fixed, it can be merged.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
MvLieshout: Yeah, I think we should make some general decisions in how strict we want to be in input validation.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/missing_values_transform.py
MvLieshout: I think that's reasonable. Creating a field with the prefix _ is not allowed, but when making it a private attribute this is possible. See new implementation.

PR #650, opened 2025-08-26
feature(#626): Add feature engineering

lschilders:


packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/feature_clipper.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Returns:
+            Transformed time series dataset with clipped features.
+        """
+        transformed_data = data.data.copy()
+
+        for col in self.feature_ranges:
+            if col in transformed_data.columns:
+                min_val, max_val = self.feature_ranges[col]
+                transformed_data[col] = transformed_data[col].clip(lower=min_val, upper=max_val)
```
egordm: I think this feature might be vectorized if you store the ranges in dataframes (also by doing df.max()). You could use something like this:

```
min_aligned = self.feature_mins.reindex(transformed_data.columns)
            max_aligned = self.feature_maxs.reindex(transformed_data.columns)
            transformed_data = transformed_data.clip(lower=min_aligned, upper=max_aligned, axis=1)
```

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/scaler.py
```diff
@@ -0,0 +1,121 @@
...
+class ScalingMethod(StrEnum):
+    """Scaling methods from sklearn.preprocessing."""
+
+    MinMax = "min-max"
+    MaxAbs = "max-abs"
+    Standard = "standard"
+    Robust = "robust"
+
+
+class Scaler(TimeSeriesTransform):
```
egordm: Suffixing the transforms would probably more clear what this component is. For example `ScalerTransform`

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/trend_features.py
```diff
@@ -0,0 +1,146 @@
...
+        default=timedelta(hours=24),
+        description="Rolling window size for the aggregation.",
+    )
+    aggregation_functions: list[AggregationFunction] = Field(
+        default_factory=lambda: [AggregationFunction.MEDIAN, AggregationFunction.MIN, AggregationFunction.MAX],
+        description="List of aggregation functions to compute over the rolling window. ",
+    )
+
+
+class RollingAggregateFeatures(TimeSeriesTransform):
```
egordm: The file of this feature should probably be called something like `rolling_aggregate_transform.py`? The trend feature is usually created by extrapolating with a simpler model.

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/trend_features.py
```diff
@@ -0,0 +1,146 @@
...
+class AggregationFunction(StrEnum):
+    """Enum of supported aggregation functions for rolling aggregates."""
+
+    MEAN = "mean"
+    MEDIAN = "median"
+    MAX = "max"
+    MIN = "min"
+
+
+class RollingAggregateFeaturesConfig(BaseConfig):
```
egordm: Would it make sense to move configuration inside the the actual transform?

This would simplify the api a lot, especially for new users. A separate config is usually used when we want to forward config object to somewhere else.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/cyclic_features.py
```diff
@@ -0,0 +1,228 @@
...
+        Returns:
+            DataFrame with month_sine and month_cosine columns.
+        """
+        return pd.DataFrame(
+            {
+                "month_sine": self._compute_sine(index.month, NUM_MONTHS_IN_YEAR),
+                "month_cosine": self._compute_cosine(index.month, NUM_MONTHS_IN_YEAR),
+            },
+            index=index,
+        )
```
egordm: Adding and documenting each of these functions might be a bit overkill. I would suggest creating _create_sine_and_cosine -> pd.DataFrame and then giving them the index in your transform function. Or keeping it since, and cosine, and building the list of pd.Series inside the transform instead.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/datetime_features.py
```diff
@@ -0,0 +1,153 @@
...
+    def _is_weekend_day(index: pd.DatetimeIndex) -> np.ndarray:
+        """Check if the day is a weekend day (Saturday or Sunday).
+
+        Args:
+            index: DatetimeIndex to check for weekend days.
+
+        Returns:
+            Numpy array where 1 indicates a weekend day and 0 otherwise.
+        """
+        return (index.weekday >= MIN_WEEKEND_IDX).astype(int)
```
egordm: Similar here. There functions might be overkill since they will probably never be overriden.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/holiday_features.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Set 1 for dates that match each specific holiday
+        for holiday_date, holiday_name in country_holidays.items():
+            feature_name = holiday_feature_mapping[holiday_name]
+            # Use .loc to handle cases where holiday_date might not be in index
+            mask = index.date == holiday_date
+            if mask.any():
+                holiday_features.loc[mask, feature_name] = 1
+
+        return holiday_features
```
egordm: I wonder if this loop might become a bottleneck in the future and if it can be made efficient. 

Currenltly if N is the length of the data and H is number of holidays, the complexity of this function is O(NH*2).
Perhaps it's possible to make this an O(N+H) by first putting holidays in a sparse dataframe and then merging on index?

Does this also work for holidays that span multiple days?

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/trend_features.py
lschilders: This is done in `feature/626-weather-transforms` (unfortunately...)

lschilders:
Update to continue during my holiday:
* All open conversations in this PR are not resolved yet.
* I made a start with implementing weather related transforms (`air_related_features`, `radiation_derived_features` and `wind_power_features`. The goal is to implement formulas that can be found on the internet (opposed to the ones in OpenSTEF v3) and document those well.
* School-holidays are not yet implemented. 
* [Marnix's comments](https://github.com/OpenSTEF/openstef/pull/650#issuecomment-3223741610) are still open questions to be answered.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/holiday_features.py
egordm: Fixed.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/datetime_features.py
egordm: Fixed.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/cyclic_features.py
egordm: Fixed

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/scaler.py
egordm: Fixed

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/trend_features.py
egordm: Fixed.

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-License-Identifier: MPL-2.0
+
+"""Forecasting feature transforms for time series data.
+
+This module provides transforms specifically designed for forecasting workflows,
+including feature scaling, trend analysis, and feature clipping utilities that
+prepare time series data for machine learning models.
+"""
+
+from .rolling_aggregate_transform import RollingAggregateTransform
```
MvLieshout, line 12: What is the reason for the relative import here?

I would prefer being consistent over all files and not use relative imports.

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/clipping_transform.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+
+    column_names: list[str] = Field(description="Columns to apply clipping to.")
+
+    _feature_mins: pd.Series = PrivateAttr(default_factory=pd.Series)
+    _feature_maxs: pd.Series = PrivateAttr(default_factory=pd.Series)
+    _is_fitted: bool = PrivateAttr(default=False)
+
+    @override
+    def fit(self, data: TimeSeriesDataset) -> None:
```
MvLieshout, line 69: For some functions the docstrings are missing.

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/rolling_aggregate_transform.py
```diff
@@ -0,0 +1,101 @@
...
+        >>> dataset = TimeSeriesDataset(data, timedelta(hours=1))
+        >>>
+        >>> # Initialize and apply transform
+        >>> transform = RollingAggregateTransform(
+        ...     columns=['load', 'temperature'],
+        ...     rolling_window_size=timedelta(hours=2),
+        ...     aggregation_functions=["mean", "max"]
+        ... )
+        >>> transformed_dataset = transform.transform(dataset)
+        >>> 'rolling_mean_load_PT2H' in transformed_dataset.data.columns
```
MvLieshout: I think we could improve this example by showing what these values would like (maybe just a selection of the columns if the example is too big otherwise?)

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+"""Temporal feature transforms for time series data.
+
+This module provides transforms that extract temporal features from datetime indices
+of time series datasets. These transforms add time-based features such as cyclic
+patterns, holiday indicators, and daylight information to enhance time series
+forecasting models.
+"""
+
+from .cyclic_features_transform import CyclicFeaturesTransform
```
MvLieshout, line 13: Same for this relative import

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/cyclic_features_transform.py
```diff
@@ -0,0 +1,164 @@
...
+        >>> # Create sample dataset
+        >>> data = pd.DataFrame({
+        ...     'load': [100, 120, 110]
+        ... }, index=pd.date_range('2025-01-01 12:00:00', periods=3, freq='1h'))
+        >>> dataset = TimeSeriesDataset(data, timedelta(hours=1))
+        >>>
+        >>> # Apply cyclic features with custom configuration
+        >>> transform = CyclicFeaturesTransform(included_features=["season", "timeOfDay"])
+        >>> transformed = transform.transform(dataset)
+        >>> 'season_sine' in transformed.data.columns
```
MvLieshout: Same for this example. I think showing the columns could be nice.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/cyclic_features_transform.py
```diff
@@ -0,0 +1,164 @@
...
+        return TimeSeriesDataset(
+            data=pd.concat(
+                [data.data, *features],
+                axis=1,
+            ),
+            sample_interval=data.sample_interval,
+        )
+
+
+__all__ = ["CyclicFeaturesTransform"]
```
MvLieshout, line 163: Is this still necessary if it also already in the `__init__.py`?

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/datetime_features_transform.py
```diff
@@ -0,0 +1,100 @@
...
+            quarter_dummies.index = data.index
+            features.extend([month_dummies, quarter_dummies])
+
+        return TimeSeriesDataset(
+            pd.concat([data.data, *features], axis=1),
+            sample_interval=data.sample_interval,
+        )
+
+
+__all__ = ["DatetimeFeaturesTransform"]
```
MvLieshout, line 100: I think this can be moved to `__init__.py` as well?

packages/openstef-core/src/openstef_core/feature_engineering/weather_transforms/radiation_derived_features.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if self.radiation_column not in data.feature_names:
+            raise MissingColumnsError([self.radiation_column])
+
+        try:
+            import pvlib  # noqa: PLC0415 - delayed import due to optional dependency
+        except ImportError as e:
+            raise MissingExtraError("feature_engineering") from e
+
+        # Convert radiation from J/m² to kWh/m² and rename to 'ghi'
```
MvLieshout, line 112: Do we somewhere document what type of input can be given and what its unit should be? I think that would be nice.

packages/openstef-core/tests/unit/feature_engineering/weather_transforms/test_radiation_derived_features.py
```diff
@@ -0,0 +1,371 @@
...
+    transform = RadiationDerivedFeaturesTransform(
+        coordinate=Coordinate(latitude=Latitude(52.0), longitude=Longitude(5.0)), radiation_column="solar_irradiance"
+    )
+
+    # Act & Assert
+    with pytest.raises(MissingColumnsError):
+        transform.transform(dataset)
+
+
+# Integration tests with real pvlib calls
```
MvLieshout: I am likely missing something. Are the tests above not calling pvlib?

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/rolling_aggregate_transform.py
MvLieshout: Don't know how difficult this is btw. If too much work, you can leave this for now imo.

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/clipping_transform.py
egordm: It's intentional. 

`@override` causes functions to inherit docs.

There is no need to document fit every time since the parent method is already well documented.

The idea is also that because user doesn't directly call the fit and transform usually, the most of the documentation should go into the class docstring.

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/rolling_aggregate_transform.py
egordm: Yeah, I don't know if an example docstring is the right place to do this. For that we should probably create an example script instead. Perhaps with a few plots.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/__init__.py
egordm: Let's fix these after the merge.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/cyclic_features_transform.py
egordm: See above.

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/datetime_features_transform.py
egordm: see above

packages/openstef-core/src/openstef_core/feature_engineering/weather_transforms/radiation_derived_features.py
egordm: Fixed.

packages/openstef-core/tests/unit/feature_engineering/weather_transforms/test_radiation_derived_features.py
egordm: Yeah it's probably a remnant from old code. Removed.

packages/openstef-core/src/openstef_core/feature_engineering/forecasting_transforms/rolling_aggregate_transform.py
egordm: Nevermind I see what you mean now. Fixing.

MvLieshout (approved):

packages/openstef-core/src/openstef_core/feature_engineering/temporal_transforms/cyclic_features_transform.py
egordm: Fixed.

PR #651, opened 2025-08-28
feature(626): Flatliner check transform

MvLieshout:


packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/flatliner_check.py
```diff
@@ -0,0 +1,138 @@
...
+        latest_measurements = data[start_time:latest_measurement_time].dropna()
+
+        flatliner_value = latest_measurements.median() if self.detect_non_zero_flatliner else 0
+
+        # check if all values are within a relative tolerance of each other
+        flatline_condition = np.isclose(
+            a=latest_measurements,
+            b=flatliner_value,
+            atol=0,
+            rtol=self.relative_tolerance,
```
MvLieshout, line 122: If `detect_non_zero_flatliner` is `False`, using the relative tolerance will have the effect that all values should be zero. This feels quite strict. Is this the best for our use case? Do we not also want an absolute tolerance to be configurable (this would then be different from what we have in OpenSTEF v3)? 

Edit: After discussing with Bart, we decided that it would be good to also add absolute tolerance as an argument.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/flatliner_check.py
```diff
@@ -0,0 +1,137 @@
...
+        Raises:
+            ValueError: If the input DataFrame does not contain a 'load' column.
+        """
+        if "load" not in data.data.columns:
+            raise ValueError("The DataFrame must contain a 'load' column.")
+        self._flatliner_indicator = self.detect_ongoing_flatliner(
+            data=data.data["load"],
+        )
+
+    def transform(self, data: TimeSeriesDataset) -> TimeSeriesDataset:  # noqa: PLR6301
```
MvLieshout: The noqa is there to get around the ruff error below (which exists because I give "self" as an argument): 

`packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/flatliner_check.py:128:9:` PLR6301 Method `transform` could be a function, class method, or static method`

 By ignoring it, we keep the same structure as in other transforms. If I make it static, I run into the following: `Method "transform" overrides class "TimeSeriesTransform" in an incompatible manner
`.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/flatliner_check.py
```diff
@@ -0,0 +1,142 @@
...
+    _flatliner_indicator: bool = PrivateAttr(default=False)
+
+    @property
+    def flatliner_indicator(self) -> bool:
+        """Indicates if a flatliner is currently detected after fitting."""
+        return self._flatliner_indicator
+
+    def __init__(self, **data: Any):
+        """Initializes the FlatlinerCheckTransform with the given configuration."""
+        super().__init__(**data)
```
egordm:
```suggestion
```


The init can be safely remove I believe.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/flatliner_check.py
```diff
@@ -0,0 +1,142 @@
...
+    relative_tolerance: float = Field(
+        default=1e-5,
+        description="The relative tolerance for considering values as equal when detecting flatliners.",
+    )
+    _flatliner_indicator: bool = PrivateAttr(default=False)
+
+    @property
+    def flatliner_indicator(self) -> bool:
+        """Indicates if a flatliner is currently detected after fitting."""
+        return self._flatliner_indicator
```
egordm: We might want to rename this to something more descriptive. Since it's a bool, something like `is_flatliner_detected` might be a good one.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/flatliner_check.py
egordm: I think most of such errors can be fixed by adding `@override` to state your intention to the linter.

Also, the transform method probably shouldn't be empty because we want to detect if it's a flatliner during forecasting as well right?

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/flatliner_check.py
```diff
@@ -0,0 +1,142 @@
...
+import numpy as np
+import pandas as pd
+from pydantic import Field, PrivateAttr
+
+from openstef_core.base_model import BaseConfig
+from openstef_core.datasets import TimeSeriesDataset
+from openstef_core.datasets.transforms import TimeSeriesTransform
+
+
+class FlatlinerCheckTransform(TimeSeriesTransform, BaseConfig):
```
egordm, line 25: The initial idea was also to throw a custom error when a flatliner is detected. `FlatlinerDetectedError` or similar. Perhaps we could make that behaviour a configurable option.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/flatliner_check.py
MvLieshout: Good point, I have added the _option_ to check it. I can imagine the case where one would not like to check this. To save compute when you trust the production data for example.

egordm (approved):
Looks good now. I think only a few changes are needed to let the CI pass?

PR #655, opened 2025-08-28
feature(626): Completeness check

MvLieshout:


packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/completeness_check.py
```diff
@@ -0,0 +1,169 @@
...
+            InsufficientlyCompleteError: If the dataset is not sufficiently complete.
+        """
+        self._completeness = self._calculate_completeness(
+            data=data.data,
+        )
+        self._is_sufficiently_complete = self._calculate_sufficiently_complete(
+            data=data.data,
+        )
+        if not self._is_sufficiently_complete and self.error_on_insufficient_completeness:
+            raise InsufficientlyCompleteError("The dataset is not sufficiently complete.")
```
egordm: For this transform it probably doesn't make sense to fit it since it's a test whether data is complete.

During transform a completely different dataset may be passed. Testing only needs to happen in transform.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/completeness_check.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @property
+    def completeness(self) -> float:
+        """Indicates how complete the data is."""
+        return self._completeness
+
+    @property
+    def is_sufficiently_complete(self) -> bool:
+        """Indicates whether the data is sufficiently complete."""
+        return self._is_sufficiently_complete
```
egordm: Similar to fit function the state for this transform can also be removed.

MvLieshout: Fixed. Now the transform function only returns the original dataframe if the data is sufficiently complete or an error if not sufficiently complete.

packages/openstef-core/src/openstef_core/feature_engineering/validation_transforms/completeness_check.py
```diff
@@ -0,0 +1,126 @@
...
+    weights: dict[str, float] | None = Field(
+        default=None,
+        description="Weights for each column to adjust their importance in the completeness calculation.",
+    )
+    completeness_threshold: float = Field(
+        default=0.5,
+        description="Threshold for completeness below which the data is considered insufficiently complete.",
+    )
+    _completeness: float = PrivateAttr(default=False)
+    _is_sufficiently_complete: bool = PrivateAttr(default=False)
```
egordm: We can omit state for transform as well. If data is incomplete an error is thrown. If user should be warned we can just log a warning.

PR #663, opened 2025-09-02
feature(#626): Air-related Features Transform

MvLieshout:


packages/openstef-models/src/openstef_models/feature_engineering/weather_transforms/air_related_features_transform.py
```diff
@@ -0,0 +1,196 @@
...
+    )
+    relative_humidity_column: str = Field(
+        default="relative_humidity",
+        description="Name of the relative humidity (%) column.",
+    )
+
+    def __init__(self, **kwargs: Any) -> None:
+        """Initialize the AirRelatedFeaturesTransform."""
+        super().__init__(**kwargs)
+        self._air_related_features: pd.DataFrame = pd.DataFrame()
```
egordm: This transform should probably be stateless. Just transform the data and return it.

If we store the data in the transform. High chance that there will be bugs when other data is used. Also this data is stored when transforms are persisted, which unnecessarily bloats the checkopoint.

packages/openstef-models/src/openstef_models/feature_engineering/weather_transforms/air_related_features_transform.py
```diff
@@ -0,0 +1,196 @@
...
+        ]
+
+        if missing_columns:
+            raise MissingColumnsError(missing_columns)
+
+        temperature = data.data[self.temperature_column]
+        pressure = data.data[self.pressure_column]
+        relative_humidity = data.data[self.relative_humidity_column]
+
+        self._air_related_features = pd.DataFrame(index=data.data.index)
```
egordm: See above about the stateless behaviour.

packages/openstef-models/src/openstef_models/feature_engineering/weather_transforms/air_related_features_transform.py
```diff
@@ -0,0 +1,196 @@
...
+
+from openstef_core.base_model import BaseConfig
+from openstef_core.datasets import TimeSeriesDataset
+from openstef_core.datasets.transforms import TimeSeriesTransform
+from openstef_core.exceptions import MissingColumnsError
+
+type AirRelatedFeatureName = Literal["saturation_vapour_pressure", "vapour_pressure", "dewpoint", "air_density"]
+
+
+class AirRelatedFeaturesTransform(BaseConfig, TimeSeriesTransform):
```
egordm: Air related sounds very vague.

A more suitable name might be AtmosphereDerivedFeaturesTransform. We could check what Ment thinks about this. But air density, dewpoint, pressure are all 'atmosphere' features I believe.

packages/openstef-models/src/openstef_models/feature_engineering/weather_transforms/air_related_features_transform.py
```diff
@@ -0,0 +1,196 @@
...
+    """Transform that calculates air-related meteorological features from basic weather data.
+
+    This transform calculates various air-related features including saturation vapour pressure,
+    vapour pressure, dewpoint, and air density using standard meteorological formulas. It requires
+    temperature, pressure, and relative humidity as input columns.
+    The calculated features can be used to enhance weather-based prediction models by providing
+    additional atmospheric state information that may correlate with energy consumption patterns.
```
egordm: This part is quite vague. I think we can write the actual reason. As far as I know air pressure and vapour affects PV generation. There were a few papers about this, citation needed. But here is what google says:


> Higher air pressure, often found at lower altitudes, tends to increase solar panel output (voltage, current, and intensity) because it creates a stronger downward pull on solar radiation. Conversely, lower air pressure at high altitudes has the opposite effect, potentially decreasing performance and voltage requirements.

packages/openstef-models/src/openstef_models/feature_engineering/weather_transforms/air_related_features_transform.py
```diff
@@ -0,0 +1,196 @@
...
+        >>> transform = AirRelatedFeaturesTransform(
+        ...     included_features=["dewpoint", "air_density"]
+        ... )
+        >>>
+        >>> # Apply transformation
+        >>> result = transform.transform(dataset)
+        >>> "dewpoint" in result.data.columns
+        True
+        >>> "air_density" in result.data.columns
+        True
```
egordm:
```suggestion
        >>> "dewpoint" in result.feature_names
        True
        >>> "air_density" in result.feature_names
        True
```

We can use a more direct attribute here. Or test for all the columns.

```
>>> result.feature_names
['temperature', 'pressure', 'relative_humidity', 'dewpoint', 'air_density'] 
```

packages/openstef-models/tests/unit/feature_engineering/weather_transforms/test_air_related_features_transform.py
```diff
@@ -0,0 +1,100 @@
...
+def test_air_density_calculation():
+    transform = AirRelatedFeaturesTransform()
+    temperature = pd.Series([20, 10, -10])
+    relative_humidity = pd.Series([80, 50, 50])
+    pressure = pd.Series([1000, 1000, 1020])
+    result = transform._calculate_air_density(temperature, relative_humidity, pressure)
+    assert isinstance(result, pd.Series)
+    assert np.all(result > 0)
+    # Air density increases as temperature decreases, relative humidity decreases and pressure increases
+    assert np.all(result.astype(dtype=float).diff().dropna() > 0)
```
egordm: These seem to be nice sanity checks.

Would it be an idea to also test on some known perhaps hardcoded outputs since these are scientific formulas?

egordm (changes requested):
Looks good. Nice that you have added the references for the formulas.

I have added a few comments and a few nitpicks.

egordm (approved):
LGTM! Nice work. Looks really clean!

PR #668, opened 2025-09-10
fix(STEF-2391): Improved feature clipping (to avoid extrapolation) by switching it from minmax to std clipping.

egordm:
* The solution is backwards compatible with models trained before this change. Newly trained models will used std clipping, older will continue using minmax until retrain.

### DIFF
--- a/openstef/model/metamodels/feature_clipper.py
+++ b/openstef/model/metamodels/feature_clipper.py
@@ -20,7 +20,7 @@ class FeatureClipper(BaseEstimator, TransformerMixin):
     extrapolating beyond these values during prediction.
     """
 
-    def __init__(self, columns: List[str], clip_std: float = 2.0):
+    def __init__(self, columns: List[str], clip_number_of_std: float = 2.0):
         """
         Initialize the FeatureClipper.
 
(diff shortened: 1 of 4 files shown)


openstef/model/metamodels/feature_clipper.py
```diff
@@ -24,7 +30,8 @@ def __init__(self, columns: List[str]):
             List of column names to be clipped.
         """
         self.columns: List[str] = columns
-        self.feature_ranges: Dict[str, Tuple[float, float]] = {}
+        self.feature_ranges: Dict[str, Tuple[float, float] | FeatureStats] = {}
+        self.clip_std = clip_std
```
MvLieshout: This naming is a bit vague to me as this is not the standard deviation itself but the number of std from the mean from which we clip. Maybe `clip_number_of_std` would be better?

MvLieshout (commented):
Implementation looks good. 1 comment about naming.

openstef/model/metamodels/feature_clipper.py
egordm: Yeah I think it would be. Fixed.

Issue #633, opened 2025-07-30
[OpenSTEF 4.0] Implement pipelines


egordm:
As openstef community
I want to pipelines
So that I can run common tasks easily and have openstef 3 feature parity

❗Priority (What if we don't do this?/Are there any deadlines? etc.)

Definition of Done:

✅ Acceptance criteria

- [x] Implemented pipelines in `openstef-models/pipelines`
    - [x] Train Pipeline
    - [x] Predict Pipeline
    - [x] Split Pipeline
- [x] Implement a Workflow interface [see architecture](https://lf-energy.atlassian.net/wiki/spaces/OS/pages/349044781/OpenSTEF+4.0+Architecture+Document#Deployment)
    - [x] Add a generic but flexible way to configure all the components
    - [x] Create and document examples for implementation
- [ ] Create opinionated Workflows similar to current openstef 3 pipelines
- [x] Document pipelines

📄 Documentation criteria:

Interfaces with other products and services

Update relevant confluence pages

🧪 Test criteria:

Unit tests written and passed

⌛ Dependencies:

N/A

🚀 Releasing:

N/A

Other information:

🌍 Background

PR #674, opened 2025-09-16
feature(#633): Added model pipeline, interfaces for postprocessing, storage, callbacks. Refactor by reorganizing transforms.

egordm:
This pull request refactors the OpenSTEF models package to improve clarity, modularity, and maintainability. The main changes include reorganizing feature engineering modules under a new `transforms` namespace, introducing a new base class for forecast postprocessing transforms, expanding exception handling, and updating documentation and examples to reflect these changes.

**Core package restructuring:**

* The `feature_engineering` subpackage has been replaced with a new `transforms` namespace, consolidating all feature, temporal, and forecasting transforms for better clarity and modularity. Old `feature_engineering` modules and their `__init__.py` files have been removed. [[1]](diffhunk://#diff-e759adbed3c551bf6129b16ddcff2f496852c6e1fe37fa0bb8c36af48da5dd19L64-L67) [[2]](diffhunk://#diff-fe7c7456f13b6d4b24b11880ced228a565bc4cfab01493ed11be89faf9fadb55L1-L32) [[3]](diffhunk://#diff-86fd2b68880c0537e9f53cd3889cdcec46b4534dca44e6cdef656b08e115cce9L1-L16) [[4]](diffhunk://#diff-6fc659532c2350d4f937e5aebb5994838d08aca2750414c63966a53e40519600L1-L19)

> We have different transforms like TimeSeriesTransforms (general purpose / preprocessing), ForecastTransforms (postprocessing), and more.
> Also a choice was 

**Transforms and pipeline improvements:**

* Introduced a new `ForecastTransform` base class in `openstef_core.datasets.transforms`, providing a standard interface for postprocessing/corrective operations on forecast outputs (e.g., calibration, quantile sorting). The base classes for time series transforms now include an `is_fitted` property for stateless transforms. [[1]](diffhunk://#diff-2632edfa6ce67dd845cf19c0ca6ed783c45842b213e8953b5a767633953d65cdR48-R57) [[2]](diffhunk://#diff-2632edfa6ce67dd845cf19c0ca6ed783c45842b213e8953b5a767633953d65cdR183-R192) [[3]](diffhunk://#diff-2632edfa6ce67dd845cf19c0ca6ed783c45842b213e8953b5a767633953d65cdR234-R306)

**Expanded exception handling:**

* Added new exception classes: `NotFittedError`, `ModelNotFoundError`, `UnreachableStateError`, and `ConfigurationError` to improve error reporting and debugging. The former `ModelNotFittedError` has been replaced by the more general `NotFittedError`.

**Documentation and code style updates:**

* Updated documentation and code style guides to use the new `transforms` import paths instead of `feature_engineering`. Example file naming conventions and logging instructions have also been updated accordingly. [[1]](diffhunk://#diff-522346a2505e21a14199528cc7fd25aa995145f6c52671b13ccb3413c7a6ea06L89-R89) [[2]](diffhunk://#diff-522346a2505e21a14199528cc7fd25aa995145f6c52671b13ccb3413c7a6ea06L114-R114) [[3]](diffhunk://#diff-522346a2505e21a14199528cc7fd25aa995145f6c52671b13ccb3413c7a6ea06L489-R489) [[4]](diffhunk://#diff-dba8a2ad9abf4bd68c030a55ed7cb853f25743eff7211a302d2d7d86cf811b9aL319-R319) [[5]](diffhunk://#diff-f8984869df65aed9f033c6787889a342d11bace29e91e4447ea6621c80f841a1L116-R116) [[6]](diffhunk://#diff-f8984869df65aed9f033c6787889a342d11bace29e91e4447ea6621c80f841a1L290-R290)

**Examples and integrations:**

* Added a new comprehensive example, `configuring_model_pipeline_example.py`, demonstrating how to set up a full forecasting pipeline with the new `transforms` structure. Also introduced an `integrations` optional dependency group in `pyproject.toml` and a new `integrations` package for external system hooks. [[1]](diffhunk://#diff-a86af6102d8e37cec4e22e70ecd7cad2814da09ec566bed79b1e24c7ce67f25dR1-R93) [[2]](diffhunk://#diff-7d06aeeff91cc46b8a0512805d691baca01de13ef989715595f6274eca593648R40-R43) [[3]](diffhunk://#diff-f7d05997b975e0e27e9f610ebbecaae685928045a7b84f671e4492b764b52b90R1-R10)

These changes make the codebase easier to use, extend, and maintain, and ensure that documentation and examples reflect the updated structure.

### DIFF
--- a/packages/openstef-core/src/openstef_core/datasets/transforms.py
+++ b/packages/openstef-core/src/openstef_core/datasets/transforms.py
@@ -35,7 +35,11 @@ class TimeSeriesTransform:
 
         >>> class ScaleTransform(TimeSeriesTransform):
         ...     def __init__(self):
-        ...         self.scale_factor = 1.0
+        ...         self.scale_factor = None
+        ...
+        ...     @property
+        ...     def is_fitted(self) -> bool:
+        ...         return self.scale_factor is not None
         ...
         ...     def fit(self, data):
         ...         self.scale_factor = data.data.max().max()
--- a/packages/openstef-models/pyproject.toml
+++ b/packages/openstef-models/pyproject.toml
@@ -31,14 +31,11 @@ dependencies = [
   "pycountry>=24.6.1",
 ]
 
-optional-dependencies.feature_engineering = [
+optional-dependencies.all = [
   "holidays>=0.79",
+  "joblib>=1",
   "pvlib>=0.13",
   "scikit-learn>=1.7.1",
-]
-
-optional-dependencies.integrations = [
-  "joblib>=1",
 ]
 
 urls.Documentation = "https://openstef.github.io/openstef/index.html"
--- a/packages/openstef-models/src/openstef_models/integrations/callbacks.py
+++ b/packages/openstef-models/src/openstef_models/integrations/callbacks.py
@@ -32,17 +32,17 @@ class ForecastingCallback:
         Creating a logging callback:
 
         >>> class LoggingCallback(ForecastingCallback):
-        ...     def on_before_fit(self, pipeline, dataset):
+        ...     def on_fit_start(self, pipeline, dataset):
         ...         print(f"Starting training with {len(dataset.data)} samples")
         ...
-        ...     def on_after_predict(self, pipeline, dataset, forecasts):
+        ...     def on_predict_end(self, pipeline, dataset, forecasts):
         ...         print(f"Generated {len(forecasts.data)} forecasts")
         >>>
         >>> callback = LoggingCallback()
         >>> workflow = ForecastingWorkflow(model, callbacks=callback) # doctest: +SKIP
     """
 
-    def on_before_fit(self, pipeline: "ForecastingWorkflow", dataset: VersionedTimeSeriesDataset | TimeSeriesDataset):
+    def on_fit_start(self, pipeline: "ForecastingWorkflow", dataset: VersionedTimeSeriesDataset | TimeSeriesDataset):
         """Called before model fitting begins.
 
         Use this hook for pre-training validation, data preprocessing,
@@ -53,7 +53,7 @@ class ForecastingCallback:
             dataset: Training dataset being used for fitting.
         """
 
-    def on_after_fit(self, pipeline: "ForecastingWorkflow", dataset: VersionedTimeSeriesDataset | TimeSeriesDataset):
+    def on_fit_end(self, pipeline: "ForecastingWorkflow", dataset: VersionedTimeSeriesDataset | TimeSeriesDataset):
         """Called after model fitting completes successfully.
 
         Use this hook for post-training validation, model evaluation,
@@ -64,7 +64,7 @@ class ForecastingCallback:
             dataset: Training dataset that was used for fitting.
         """
 
-    def on_before_predict(
+    def on_predict_start(
         self, pipeline: "ForecastingWorkflow", dataset: VersionedTimeSeriesDataset | TimeSeriesDataset
     ):
         """Called before prediction generation begins.
@@ -77,7 +77,7 @@ class ForecastingCallback:
             dataset: Input dataset being used for prediction.
         """
 
-    def on_after_predict(
+    def on_predict_end(
         self,
         pipeline: "ForecastingWorkflow",
         dataset: VersionedTimeSeriesDataset | TimeSeriesDataset,
--- a/packages/openstef-beam/pyproject.toml
+++ b/packages/openstef-beam/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
   "tqdm>=4.67.1",
 ]
 
-optional-dependencies.boto = [
+optional-dependencies.all = [
   "s3fs>=2025.5.1",
 ]
 urls.Documentation = "https://openstef.github.io/openstef/index.html"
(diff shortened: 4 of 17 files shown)


packages/openstef-models/pyproject.toml
lschilders, line 1: There is `optional-dependencies.feature_engineering` in this pyproject, should be renamed to `optional-dependencies.transforms`

packages/openstef-core/src/openstef_core/datasets/transforms.py
lschilders, line 1: Is this compatible with the `_is_fitted: bool = PrivateAttr(default=False)` that is now used in most transforms?

packages/openstef-models/src/openstef_models/integrations/callbacks.py
lschilders, line 1: Common naming patterns for these are normally `on_{verb}_start/end` I think. Might be better to rename it to this? `on_before/after` sounds a bit strange imo.

packages/openstef-models/src/openstef_models/models/forecasting_model.py
```diff
@@ -0,0 +1,183 @@
...
+    ) -> dict[LeadTime, ForecastInputDataset]:
+        input_data = self.preprocessing.transform(dataset=dataset)
+        return {
+            key: ForecastInputDataset.from_timeseries_dataset(
+                dataset=value,
+                target_column=self.target_column,
+                forecast_start=forecast_start,
+            )
+            for key, value in input_data.items()
+        }
```
lschilders:
```suggestion
        return {
            lead_time: ForecastInputDataset.from_timeseries_dataset(
                dataset=timeseries_dataset,
                target_column=self.target_column,
                forecast_start=forecast_start,
            )
            for lead_time, timeseries_dataset in input_data.items()
        }
```

packages/openstef-models/src/openstef_models/models/forecasting_model.py
lschilders, line 1: This nomenclature ForecastingModel <> Forecaster is analogous to v3 Pipeline <> Model right? Why choose a different one? imo, the v3 one is more clear. Then you would have FeatureTransformPipeline, ForecastingPipeline and ForecastTransformPipeline, which makes sense I'd say.

Edit: it might be to not confuse with ForecastingWorkflow?

packages/openstef-models/src/openstef_models/transforms/feature_pipeline.py
```diff
@@ -121,13 +122,13 @@ class FeaturePipeline(BaseModel):
         description="The lead times (horizons) for which the model will make predictions.",
     )
 
-    versioned_transforms: list[VersionedTimeSeriesTransform] = Field(
+    versioned_transforms: Sequence[VersionedTimeSeriesTransform] = Field(
```
lschilders, line 125: Why change this to Sequence? Do we expect/want tuples here as well?

lschilders (commented):
Looks great! Some questions/comments.

packages/openstef-models/src/openstef_models/transforms/feature_pipeline.py
egordm: Yes that is intentionally. Apparently when you type it aas list[BaseType] you can not assign a list[all subtypes] to it without doing casting. Therefore Sequence is needed which does make it possible.

It doesn't change much about the behaviour since it should be an immutable list / sequence anyways.

packages/openstef-models/src/openstef_models/models/forecasting_model.py
egordm: Yes, I ditched the pipeline and instead used ForecastingModel. The main reasoning:

* Pipeline does one thing, so we'd need a train and predict pipeline. Looking at the code, you can see they would do quite a similar thing.
* Forecaster does only forecasting, but preprocessing and postprocessing are also part of the model (state). Even if we want it separately configurable. So in this case ForecastingModel is a container for all of these preprocess, forecast, postprocess. Makes it very easy for saving stuff and passing it to other things.
* Workflow was already part of the idea, which is what takes over part of pipeline functionality where you can use callbacks and such. It does what we previously would call pipelines, but also making it easy to configure for a high level (to implement soon), and also combines multiple pipelines.

packages/openstef-models/src/openstef_models/models/forecasting_model.py
egordm: Added.

packages/openstef-models/src/openstef_models/integrations/callbacks.py
egordm: Makes sense. Fixed.

packages/openstef-core/src/openstef_core/datasets/transforms.py
egordm: It is now. I have updated it.

packages/openstef-models/pyproject.toml
egordm: I think feature engineering as optional dependency makes sense. But I got another idea to put all optional dependencies under all optional group. feature engineering dependency group is too broad to be useful.

Instead we should use all for now. And later see if there is any logical groups we can add.

Change the group to be called all.

packages/openstef-models/src/openstef_models/models/forecasting_model.py
lschilders: Thanks for the explanation, good reasoning, agreed.

lschilders (approved):
LGTM!

PR #675, opened 2025-09-19
add rIQD and relative pinball loss metrics

MentReeze:


packages/openstef-beam/src/openstef_beam/metrics/metrics_deterministic.py
```diff
@@ -345,3 +345,131 @@ def fbeta(
...
+    measurement_range_lower_q: float = 0.05,
+    measurement_range_upper_q: float = 0.95,
+) -> float:
+    """Calculate the relative Inter Quantile Distance (rIQD).
+
+    rIQD measures the average distance between two quantiles, normalized by the measurement range.
+
+    Args:
+        y_true: Ground truth values with shape (num_samples,). Should not contain
+            values close to zero to avoid division issues.
```
lschilders: Copy-paste from mape I think, not really relevant here.
```suggestion
        y_true: Ground truth values with shape (num_samples,).
```

packages/openstef-beam/src/openstef_beam/metrics/metrics_deterministic.py
```diff
@@ -345,3 +345,131 @@ def fbeta(
...
+        y_pred_lower_q: Predicted values of lower quantile with shape (num_samples,).
+        y_pred_upper_q: Predicted values of upper quantile with shape (num_samples,).
+        measurement_range_lower_q: Lower quantile for range calculation. Must be in [0, 1].
+        measurement_range_upper_q: Upper quantile for range calculation. Must be in [0, 1]
+            and greater than measurement_range_lower_q.
+
+    Returns:
+        The relative Inter Quantile Distance (rIQD) as a float. May return inf or
+        extremely large values if y_true is close to zero. Returns NaN if the measurement
+            range is zero.
```
lschilders:
```suggestion
        The relative Inter Quantile Distance (rIQD) as a float. Returns NaN if the measurement
            range is zero.
```

lschilders (commented):
Nice work. 2 nitpicks in docstrings. Also, since there is some custom logic in the `RIQDMetricProvider` I suggest adding a test in `packages/openstef-beam/tests/unit/evaluation/test_metric_provider.py` for the symmetric quantile index logic.

PR #682, opened 2025-09-23
add quantile calibration boxplot

MentReeze:
Added scripts to plot quantile calibration across multiple targets. I chose to plot the calibration error (observed - forecasted), enabling to plot multiple boxplots next to each other for each quantile while avoiding interpretation issues. This is an example result of this visualization:

image

### DIFF
--- a/packages/openstef-beam/src/openstef_beam/analysis/plots/quantile_calibration_box_plotter.py
+++ b/packages/openstef-beam/src/openstef_beam/analysis/plots/quantile_calibration_box_plotter.py
@@ -1,0 +1,185 @@
+# SPDX-FileCopyrightText: 2025 Contributors to the OpenSTEF project <short.term.energy.forecasts@alliander.com>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+"""Quantile calibration box plotting for forecast uncertainty validation across multiple targets.
+
+This module provides specialized boxplot visualization for evaluating probabilistic forecast
+calibration across multiple targets, showing calibration error distributions per quantile.
+"""
+
+import pandas as pd
+import plotly.express as px
+import plotly.graph_objects as go
+
+from openstef_core.types import Quantile
+
+
+class QuantileCalibrationBoxPlotter:
+    """Creates boxplots showing calibration error distributions across targets for each quantile.
+
+    This plotter visualizes how well probabilistic forecasts are calibrated by showing
+    the distribution of calibration errors (observed - forecasted probability) across
+    multiple targets for each quantile level. This helps identify:
+
+    - Which quantiles are systematically over/under-confident across targets
+    - The consistency of calibration across different targets
+    - Overall uncertainty estimation quality per quantile level
+
+    Well-calibrated forecasts will have boxplots centered around zero (no systematic bias)
+    with tight distributions (consistent calibration across targets).
+
+    Example:
+        Validating forecast calibration across multiple models and targets:
+
+        >>> from openstef_core.types import Quantile
+        >>> plotter = QuantileCalibrationBoxPlotter()
+        >>>
+        >>> # Define common quantiles for all models/targets
+        >>> quantiles = [Quantile(0.1), Quantile(0.5), Quantile(0.9)]
+        >>>
+        >>> # ModelA: Well-calibrated model across different sites
+        >>> _ = plotter.add_model("ModelA", quantiles, [Quantile(0.08), Quantile(0.52), Quantile(0.88)])  # Site1
+        >>> _ = plotter.add_model("ModelA", quantiles, [Quantile(0.12), Quantile(0.48), Quantile(0.92)])  # Site2
+        >>> _ = plotter.add_model("ModelA", quantiles, [Quantile(0.09), Quantile(0.51), Quantile(0.89)])  # Site3
+        >>>
+        >>> # ModelB: Overconfident model across different sites
+        >>> _ = plotter.add_model("ModelB", quantiles, [Quantile(0.15), Quantile(0.55), Quantile(0.85)])  # Site1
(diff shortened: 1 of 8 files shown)


packages/openstef-beam/src/openstef_beam/analysis/plots/quantile_calibration_box_plotter.py
```diff
@@ -0,0 +1,185 @@
...
+        model_name: str,
+        forecasted_probs: list[Quantile],
+        observed_probs: list[Quantile],
+    ) -> "QuantileCalibrationBoxPlotter":
+        """Add a model's forecasted and observed probabilities to the plot.
+
+        Args:
+            model_name (str): The name of the model.
+            forecasted_probs (list[float]): List of forecasted probabilities.
+            observed_probs (list[float]): List of observed probabilities.
```
lschilders: We don't use type hints in docstrings in v4. I see you copied it from the other quantile prob plotter. Can you remove it from there too?
```suggestion
            model_name: The name of the model.
            forecasted_probs: List of forecasted probabilities.
            observed_probs: List of observed probabilities.
```

packages/openstef-beam/src/openstef_beam/analysis/plots/quantile_calibration_box_plotter.py
```diff
@@ -0,0 +1,185 @@
...
+        model_df_list: list[pd.DataFrame] = []
+        for model_data in self.models_data:
+            # Calculate calibration errors (observed - forecasted)
+            calibration_errors = [
+                float(obs) - float(fore)
+                for obs, fore in zip(model_data["observed_prob"], model_data["forecasted_prob"], strict=True)
+            ]
+
+            # Create quantile labels
+            quantile_labels = [f"P{int(float(q) * 100):02d}" for q in model_data["forecasted_prob"]]
```
lschilders: The `Quantile` object has a method `format` that does approximately the same thing, but then formats as `quantile_Pxx`. I think it is better to re-use that to be consistent, if it fits on the x-axis.

packages/openstef-beam/src/openstef_beam/analysis/plots/quantile_calibration_box_plotter.py
```diff
@@ -0,0 +1,185 @@
...
+            y=0,
+            xref="paper",
+            yref="paper",
+            text="(Underestimation)",
+            showarrow=False,
+            font={"size": 10, "color": "gray"},
+            xanchor="right",
+            yanchor="bottom",
+        )
+
```
lschilders: I think it becomes more readable if you put this in a separate method `_add_over_under_estimation_region_labels` or something.

packages/openstef-beam/src/openstef_beam/analysis/visualizations/quantile_calibration_box_visualization.py
```diff
@@ -0,0 +1,81 @@
...
+
+from typing import override
+
+from openstef_beam.analysis.models import AnalysisAggregation, RunName, VisualizationOutput
+from openstef_beam.analysis.plots import QuantileCalibrationBoxPlotter
+from openstef_beam.analysis.visualizations.base import ReportTuple
+from openstef_beam.analysis.visualizations.quantile_probability_visualization import QuantileProbabilityVisualization
+
+
+class QuantileCalibrationBoxVisualization(QuantileProbabilityVisualization):
```
lschilders: I think inheriting from `QuantileProbabilityVisualization` is not the best here, since you will inherit all the methods, such as the methods for other aggregation than the ones supported here and `_create_plot_title`. You can duplicate the methods required in your own class, or maybe put them outside of classes as helper functions, since they don't depend on the objects itself.

packages/openstef-beam/src/openstef_beam/analysis/visualizations/quantile_calibration_box_visualization.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    @property
+    @override
+    def supported_aggregations(self) -> set[AnalysisAggregation]:
+        """Boxplot visualization requires multiple targets, so NONE aggregation is excluded."""
+        return {
+            AnalysisAggregation.RUN_AND_NONE,
+            AnalysisAggregation.RUN_AND_TARGET,
+        }
```
lschilders, line 68: The only supported aggregations are now comparing runs, it's also useful to have them for individual runs and targets, right? Also, note that every aggregation needs its method counterpart, so RUN_AND_TARGET needs `create_by_run_and_target`

lschilders (commented):
Plot is looking nice! Some comments on the implementation to align it more with the rest of the package.

packages/openstef-beam/src/openstef_beam/analysis/visualizations/quantile_calibration_box_visualization.py
```diff
@@ -53,16 +61,167 @@ class QuantileCalibrationBoxVisualization(QuantileProbabilityVisualization):
...
+
+        for quantile in quantile_columns:
+            quantile_metrics = global_metrics.metrics.get(quantile)
+            if quantile_metrics is not None:
+                observed_prob = quantile_metrics.get("observed_probability")
+                if observed_prob is not None:
+                    observed_probs.append(Quantile(observed_prob))
+                    forecasted_probs.append(quantile)
+
+        if len(observed_probs) != len(forecasted_probs):
```
lschilders: Can you call `_validate_probability_data` here? And preferable also change it in the `quantile_probability_visualization.py`, to remove duplicate code.

lschilders (approved):
Looks great! One small thing, approved in advance.

PR #701, opened 2025-09-29
add windowed metrics averaged over locations

MentReeze:
Extended the windowed metric visualization to be able to handle multiple targets for multiple runs (by taking the average over all targets within a group, per run) and groups (by taking the average over all targets in all groups, see example below). This allows for a more complete overview of the accuracy of a certain metric over time, without having to check each individual location. Windowed metrics for individual targets can still be plotted together using the "Target" aggregation level.
image

### DIFF
--- a/packages/openstef-beam/src/openstef_beam/analysis/visualizations/windowed_metric_visualization.py
+++ b/packages/openstef-beam/src/openstef_beam/analysis/visualizations/windowed_metric_visualization.py
@@ -9,10 +9,13 @@ how performance metrics evolve across different time windows.
 """
 
 import operator
+from collections import defaultdict
 from datetime import datetime
 from typing import Literal, override
 
-from openstef_beam.analysis.models import AnalysisAggregation, RunName, TargetMetadata, VisualizationOutput
+import numpy as np
+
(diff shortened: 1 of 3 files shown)


packages/openstef-beam/src/openstef_beam/analysis/visualizations/windowed_metric_visualization.py
```diff
@@ -228,5 +266,131 @@ def create_by_target(
...
+            # Average windowed metrics across all targets for this run
+            averaged_pairs = self._average_time_series_across_targets(
+                reports=target_reports,
+                metric_name=metric_name,
+                quantile_or_global=quantile_or_global,
+            )
+
+            # Skip if no averaged data points found for this run
+            if not averaged_pairs:
+                continue
```
lschilders: Instead of all the `continue` calls I would raise an error (as in the other methods in this class) like:
```suggestion
            if not averaged_pairs:
                    raise ValueError("No windowed metrics found for the specified window and metric across all groups.")
```
(If the `StrictExecutionCallback` is not added in a pipeline, this should not result in a fatal execution :))

lschilders (commented):
Looks great! Only a single suggestion to align the implementation across methods.

Issue #697, opened 2025-09-25
[OpenSTEF 4.0]  Create builtin benchmark for Liander Historical weather data


egordm:
As openstef user
I want to have a central benchmark
So that I can test various models agains a common measure

🌍 Background

> [!IMPORTANT]
> Changes should be done by forking [OpenSTEF 4.0.0 tracking branch](https://github.com/OpenSTEF/openstef/tree/release/v4.0.0) (`release/v4.0.0`) and then make a PR back onto `release/v4.0.0` branch.
> Please read the [contributing guide](https://openstef.github.io/openstef/v4/contribute/index.html) before you start.

❗Priority (What if we don't do this?/Are there any deadlines? etc.)

Definition of Done:

✅ Acceptance criteria

- [ ] The benchmark downloads measurements automatically (or manually)
- [ ] The benchmark downloads weather data automatically (or manually) (from openmeteo)
- [ ] The benchmark is openstef-beam compatible
- [ ] Integrations for basic openstef-models are created
- [ ] An example is created
- [ ] 

📄 Documentation criteria:

Interfaces with other products and services

Update relevant confluence pages

🧪 Test criteria:

Unit tests written and passed

⌛ Dependencies:

N/A

🚀 Releasing:

N/A

Other information:

egordm:
On hold

PR #717, opened 2025-10-17
feature(#697): added benchmark for public liander2024 dataset to openstef-beam

bartpleiter:
This pull request introduces a new built-in benchmark for the OpenSTEF BEAM project, specifically the Liander 2024 Short Term Energy Forecasting Benchmark. The changes add a new module to encapsulate the benchmark setup, including data loading, metric definitions, and analysis configuration. Additionally, the `huggingface-hub` dependency is added to enable dataset downloads from HuggingFace.

**New Liander 2024 Benchmark Integration:**

* Added the `liander2024.py` module, which provides the `Liander2024TargetProvider` class for loading benchmark targets, measurements, predictors, and metrics from HuggingFace datasets, and defines a comprehensive analysis configuration for visualizations and metrics.
* Implemented the `create_liander2024_benchmark_runner` function to set up a ready-to-use benchmark pipeline for Liander 2024, including backtesting, evaluation, and analysis configuration.
* Created a new `benchmarks` module with an `__init__.py` file to expose the Liander 2024 benchmark runner as a built-in benchmark entry point.

**Dependency Update:**

* Added `huggingface-hub` as a dependency in `pyproject.toml` to enable downloading benchmark datasets from HuggingFace.

### DIFF
--- a/packages/openstef-beam/src/openstef_beam/benchmarking/benchmarks/liander2024.py
+++ b/packages/openstef-beam/src/openstef_beam/benchmarking/benchmarks/liander2024.py
@@ -121,9 +121,9 @@ def create_liander2024_target_provider(
         weather_path_for_target=lambda target: Path("weather_forecasts_versioned")
         / target.group_name
         / f"weather_forecast_{target.name}.parquet",
-        profiles_path="profiles.parquet",
-        prices_path="EPEX.parquet",
-        targets_file="liander2024_targets.yaml",
+        profiles_path=lambda: Path("profiles.parquet"),
+        prices_path=lambda: Path("EPEX.parquet"),
+        targets_file_path=lambda: Path("liander2024_targets.yaml"),
         data_sample_interval=timedelta(minutes=15),
         metrics=lambda target: [
             RMAEProvider(quantiles=[Quantile(0.5)], lower_quantile=0.01, upper_quantile=0.99),
--- a/packages/openstef-beam/src/openstef_beam/benchmarking/target_provider.py
+++ b/packages/openstef-beam/src/openstef_beam/benchmarking/target_provider.py
@@ -199,11 +199,11 @@ class SimpleTargetProvider[T: BenchmarkTarget, F](TargetProvider[T, F]):
         >>> from openstef_beam.evaluation.metric_providers import RMAEProvider, RCRPSProvider
         >>> provider = SimpleTargetProvider(
         ...     data_dir=Path("./benchmark_data"),
-        ...     measurements_path_template="demand_{name}.parquet",
-        ...     weather_path_template="weather_{name}.parquet",
-        ...     profiles_path="standard_profiles.parquet",
-        ...     prices_path="energy_prices.parquet",
-        ...     targets_file="energy_targets.yaml",
+        ...     measurements_path_for_target=lambda t: Path(t.group_name) / f"load_{t.name}.parquet",
+        ...     weather_path_for_target=lambda t: Path(t.group_name) / f"weather_{t.name}.parquet",
+        ...     profiles_path=lambda: Path("standard_profiles.parquet"),
+        ...     prices_path=lambda: Path("energy_prices.parquet"),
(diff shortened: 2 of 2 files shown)


packages/openstef-beam/src/openstef_beam/benchmarking/benchmarks/liander2024.py
```diff
@@ -0,0 +1,276 @@
...
+
+    targets: list[dict[str, str | float]]
+
+    def __init__(self) -> None:
+        """Initialize the benchmark target provider by loading target definitions from HuggingFace."""
+        path: str = hf_hub_download(
+            repo_id=self.DATASET_REPO_ID,
+            filename="liander2024_targets.yaml",
+            repo_type="dataset",
+        )
```
egordm: It would probably nice to make path the input argument, and if it's none, then the dataset will be downloaded from huggingface.

packages/openstef-beam/src/openstef_beam/benchmarking/benchmarks/liander2024.py
```diff
@@ -0,0 +1,276 @@
...
+                group_name=cast(str, target["group_name"]),
+                latitude=cast(float, target["latitude"]),
+                longitude=cast(float, target["longitude"]),
+                limit=float(target["limit"])
+                if target.get("limit") is not None
+                else 0,  # Currently there is no limit defined yet in the dataset
+                benchmark_start=datetime(2024, 3, 1, tzinfo=UTC),
+                benchmark_end=datetime(2024, 12, 31, tzinfo=UTC),
+                train_start=datetime(2024, 1, 1, tzinfo=UTC),
+            )
```
egordm: Can we put benchmark_start, benchmark_end and train_start into the dataset itself? I think it would be nice for everyone to know what they are supposed to train for.

Once that is done can we just use SimpleTargetProvider? Because the dataset is basically formatted for that. I wonder if we have to reimplement various methods.

pyproject.toml
```diff
@@ -27,6 +27,7 @@ classifiers = [
 ]
 
 dependencies = [
+  "huggingface-hub>=0.35.3",
```
egordm: We shoud probably move this as an optional dependency of openstef-beam.

egordm (changes requested):

packages/openstef-beam/src/openstef_beam/benchmarking/benchmarks/liander2024.py
bartpleiter: That is indeed something to consider, but it also might be a good idea to separate the dataset from the benchmark. As in: to have the dataset on HuggingFace, but leave the use case of the dataset to the user. In our case that would be a benchmark where we train on three months and try to score high on certain metrics (OpenSTEF specific), but for others this might be somewhat different. I would say if we include these in the dataset, we should also include the evaluation criteria

packages/openstef-beam/src/openstef_beam/benchmarking/benchmarks/liander2024.py
bartpleiter: good one!

packages/openstef-beam/src/openstef_beam/benchmarking/benchmarks/liander2024.py
bartpleiter: After reconsideration, the dataset litterally has benchmark in the name. Lets add these settings in the yaml. While at it I'll also add artificial limits for "positive" and "negative" (as there is no sign convention I think)

bartpleiter: After reconsideration, the dataset literally has benchmark in the name. Lets add these settings in the yaml. While at it I'll also add artificial limits for "positive" and "negative" (as there is no sign convention I think)

egordm (approved):
LGTM!

PR #737, opened 2025-10-31
added iterative imputer

### LINKED ISSUE #654: [OpenSTEF 4.0] Add support for IterativeImputer for feature imputation
https://scikit-learn.org/stable/modules/generated/sklearn.impute.IterativeImputer.html
Idea is that when data is missing imputer can impute the data with a smaller model.

> [!IMPORTANT]

Definition of Done:

✅ Acceptance criteria

- [ ] Used existing imputer (https://scikit-learn.org/stable/modules/generated/sklearn.impute.IterativeImputer.html)
- [ ] Implemented as TimeSeriesTransform or as en extra case for existing ImputerTransform
- [ ] User can pick which columns / features to impute and the relevant covariants.
- [ ] Edge cases are documented
- [ ] Unit tests are created
- [ ] Docstrings and lint checks pass.

📄 Documentation criteria:

🧪 Test criteria:

Unit tests written and passed

⌛ Dependencies:

N/A

🚀 Releasing:

N/A

Other information:

🌍 Background

fleur-petit:
closes #654 

- Added strategy "iterative" to Imputer transform.
  - The attribute "impute_estimator" can be used to pass desired estimator (for example BayesianRidge (default), or RandomForestRegressor)
- Tests added:
  - parametrise basic test to test different strategies
  - test if iterative imputer can find linear relationships between features
  - test should work better than mean if missing value is not coincidentally the mean
  - test `initial_strategy` is used when only one feature is given with iterative strategy 

✅ Acceptance criteria

- [x] Used existing imputer (https://scikit-learn.org/stable/modules/generated/sklearn.impute.IterativeImputer.html)
- [x] Implemented as extra case for existing Imputer
- [x] User can pick which columns / features to impute and the relevant covariants.
- [x] Edge cases are documented
- [x] Unit tests are created
- [x] Docstrings and lint checks pass.

### DIFF
--- a/packages/openstef-models/src/openstef_models/transforms/general/imputer.py
+++ b/packages/openstef-models/src/openstef_models/transforms/general/imputer.py
@@ -8,7 +8,7 @@ This module provides functionality for handling missing values in time series da
 through various imputation strategies.
 """
 
-import warnings
+import logging
 from typing import Any, Literal, cast, override
 
 import numpy as np
@@ -124,7 +124,7 @@ class Imputer(BaseConfig, TimeSeriesTransform):
     ...         random_state=0,
     ...     ),
     ...     max_iterations=20,
-    ...     tolerance=1e-2
+    ...     tolerance=0.1
     ... )
     >>> transform_iterative.fit(dataset)
     >>> result_iterative = transform_iterative.transform(dataset)
@@ -134,7 +134,7 @@ class Imputer(BaseConfig, TimeSeriesTransform):
     np.False_
     >>> np.isnan(result_iterative.data["radiation"][3]) # Check if trailing NaN is preserved
     np.True_
-    >>> result_iterative.data["wind_speed"].isna().sum() == 2  # Check if trailing NaNs are preserved
+    >>> result_iterative.data["wind_speed"].isna().sum() == 2  # Wind speed NaNs preserved
     np.True_
     """
 
@@ -160,7 +160,7 @@ class Imputer(BaseConfig, TimeSeriesTransform):
     )
     tolerance: float = Field(
         default=1e-3,
-        description="Tolerance for IterativeImputer convergence",
+        description="Threshold for differences between consecutive iterations for IterativeImputer convergence",
     )
     max_iterations: int = Field(
         default=40,
@@ -169,13 +169,13 @@ class Imputer(BaseConfig, TimeSeriesTransform):
     selection: FeatureSelection = Field(
         default=FeatureSelection.ALL,
         description=(
-            "If strategy is 'iterative' both columns to use as predictors for imputation and columns to impute. "
-            "Else only columns to impute."
+            "Features to impute. If strategy is 'iterative', these features are also used as predictors for imputation."
         ),
     )
 
     _imputer: SimpleImputer | IterativeImputer = PrivateAttr()
     _is_fitted: bool = PrivateAttr(default=False)
+    _logger: logging.Logger = PrivateAttr(default=logging.getLogger(__name__))
 
     @model_validator(mode="after")
     def validate_fill_value_with_strategy(self) -> "Imputer":
@@ -242,11 +242,10 @@ class Imputer(BaseConfig, TimeSeriesTransform):
 
         # Warn if using iterative imputation with only one feature
         if self.imputation_strategy == "iterative" and len(features) < 2:  # noqa: PLR2004
-            warnings.warn(
+            self._logger.warning(
                 "Iterative imputer with only one feature will fall back to initial_strategy. "
-                f"Using '{self.initial_strategy}' for imputation.",
-                UserWarning,
-                stacklevel=2,
+                "Using '%s' for imputation.",
+                self.initial_strategy,
             )
 
         data_subset = data.data[features]
--- a/packages/openstef-models/tests/unit/transforms/general/test_imputer.py
+++ b/packages/openstef-models/tests/unit/transforms/general/test_imputer.py
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 
+import logging
+import logging.handlers
 from datetime import datetime, timedelta
 from typing import cast
 
@@ -52,7 +54,7 @@ def correlated_dataset() -> TimeSeriesDataset:
     data = pd.DataFrame(
         {
             "radiation": [10.0, np.nan, 30.0, 40.0, 50.0],
-            "temperature": [20.0, 40.0, 60.0, np.nan, 100.0],  # Missing: should be ~80
+            "temperature": [20.0, 40.0, 60.0, np.nan, 100.0],
         },
         index=pd.date_range(datetime.fromisoformat("2025-01-01T00:00:00"), periods=5, freq="1h"),
     )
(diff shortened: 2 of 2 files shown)


packages/openstef-models/src/openstef_models/transforms/general/imputer.py
```diff
@@ -112,12 +150,31 @@ class Imputer(BaseConfig, TimeSeriesTransform):
...
+        description="Tolerance for IterativeImputer convergence",
+    )
+    max_iterations: int = Field(
+        default=40,
+        description="Maximum iterations for IterativeImputer",
+    )
     selection: FeatureSelection = Field(
         default=FeatureSelection.ALL,
-        description="Features to impute.",
+        description=(
```
fleur-petit, line 171: @bartpleiter is the description of the `selection` for "iterative" strategy clear enough? Is it not too confusing that except for columns for imputation, a user can also select predictor columns here? The the distinction between columns to impute and column to use as predictor is not so clear in IterativeImputer: all selected columns are both used as predictors as well as imputed when they have missing values.

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
```diff
@@ -34,23 +41,88 @@ def sample_dataset() -> TimeSeriesDataset:
...
     # Arrange & Act & Assert
     with pytest.raises(ValidationError, match="fill_value must be provided when imputation_strategy is CONSTANT"):
         Imputer(imputation_strategy="constant")
 
 
-def test_basic_imputation_works(sample_dataset: TimeSeriesDataset):
-    """Test that basic imputation removes NaN values."""
-    # Arrange
-    transform = Imputer(imputation_strategy="mean")
+@pytest.mark.parametrize(
```
fleur-petit, line 71: added parametrize to test if basic imputation works for all strategies.

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
```diff
@@ -225,3 +297,99 @@ def test_no_missing_values_data_preservation():
...
+
+    mean_error_temp = abs(mean_temperature - true_temperature)
+    iter_error_temp = abs(iter_temperature - true_temperature)
+
+    # Iterative should be closer to true values
+    assert iter_error_rad < mean_error_rad
+    assert iter_error_temp < mean_error_temp
+
+
+def test_iterative_imputer_only_one_feature():
```
fleur-petit, line 349: @bartpleiter is it clear to you what this test does?

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
```diff
@@ -225,3 +297,99 @@ def test_no_missing_values_data_preservation():
...
+        initial_strategy="median",
+        max_iterations=5,
+        tolerance=1e-3,
+    )
+
+    # Act
+    transform.fit(dataset)
+
+    # Assert
+    with pytest.warns(UserWarning, match="Iterative imputer with only one feature"):
```
fleur-petit: @bartpleiter do you think this warning is useful?

packages/openstef-models/src/openstef_models/transforms/general/imputer.py
```diff
@@ -98,6 +110,32 @@ class Imputer(BaseConfig, TimeSeriesTransform):
...
+    ... )
+    >>> transform_iterative.fit(dataset)
+    >>> result_iterative = transform_iterative.transform(dataset)
+    >>> result_iterative.data["temperature"].isna().sum() == 0  # Temperature NaNs filled
+    np.True_
+    >>> np.isnan(result_iterative.data["radiation"][1])  # Radiation first NaN replaced
+    np.False_
+    >>> np.isnan(result_iterative.data["radiation"][3]) # Check if trailing NaN is preserved
+    np.True_
+    >>> result_iterative.data["wind_speed"].isna().sum() == 2  # Check if trailing NaNs are preserved
```
bartpleiter:
```suggestion
    >>> result_iterative.data["wind_speed"].isna().sum() == 2  # Wind speed NaNs preserved
```

packages/openstef-models/src/openstef_models/transforms/general/imputer.py
bartpleiter: I think this is a good place to indicate that there is no way to differentiate between the selected columns for imputation and the predictor columns for imputation. Explanation is clear, I would maybe rephrase it a bit into: 

"Features to impute. If strategy is 'iterative', then these features are also used as predictors for the imputation."

packages/openstef-models/src/openstef_models/transforms/general/imputer.py
```diff
@@ -174,6 +240,15 @@ def transform(self, data: TimeSeriesDataset) -> TimeSeriesDataset:
...
 
+        # Warn if using iterative imputation with only one feature
+        if self.imputation_strategy == "iterative" and len(features) < 2:  # noqa: PLR2004
+            warnings.warn(
+                "Iterative imputer with only one feature will fall back to initial_strategy. "
+                f"Using '{self.initial_strategy}' for imputation.",
+                UserWarning,
+                stacklevel=2,
+            )
+
```
bartpleiter, line 250: Nice to warn the user on this case! Let's use the logger for this though instead of warnings. See https://github.com/OpenSTEF/openstef/blob/b6c13fa3994bb86f6114236a14c70fe224468e0d/packages/openstef-models/src/openstef_models/models/forecasting_model.py#L291-L295 for an example

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
```diff
@@ -34,23 +41,88 @@ def sample_dataset() -> TimeSeriesDataset:
...
+    """Create sample dataset with simple linear relation between features where temperature = 2 * radiation.
+
+    Returns:
+        A TimeSeriesDataset with 2 features: radiation and temperature (with some NaN),
+        spanning 5 hours.
+    """
+    data = pd.DataFrame(
+        {
+            "radiation": [10.0, np.nan, 30.0, 40.0, 50.0],
+            "temperature": [20.0, 40.0, 60.0, np.nan, 100.0],  # Missing: should be ~80
```
bartpleiter: Let's remove this comment as I think it is not needed (and tells something about what will be done with the data), and radiation also does not have it
```suggestion
            "temperature": [20.0, 40.0, 60.0, np.nan, 100.0],
```

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
bartpleiter: Great to use parametrize! However, now the test becomes quite complex to read/maintain due to both parameters and the nested if/elif structure within the test. Perhaps it might be better to split the test up into two, where one test tests the non-iterative strategies, and the other one tests the iterative ones.

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
```diff
@@ -225,3 +297,99 @@ def test_no_missing_values_data_preservation():
...
+
+    # Act
+    transform.fit(correlated_dataset)
+    result = transform.transform(correlated_dataset)
+
+    # Assert
+    imputed_value = cast(float, result.data.loc[result.data.index[3], "temperature"])
+    expected_value = cast(float, correlated_dataset.data.loc[correlated_dataset.data.index[3], "radiation"]) * 2
+    # Allow some tolerance since it's ML-based
+    assert abs(imputed_value - expected_value) < 10.0, f"Expected ~{expected_value}, got {imputed_value}"
```
bartpleiter: I think it would be nice to also add a check for the imputed value of radiation[1] (as it should also be getting imputed, right?). Might be good to slightly reduce the tolerance though. Given the size of the test dataset a few percent instead of ~10 percent should be fast enough.

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
```diff
@@ -225,3 +297,99 @@ def test_no_missing_values_data_preservation():
...
+
+
+def test_iterative_imputer_better_than_mean(correlated_dataset: TimeSeriesDataset):
+    """Test that iterative imputation outperforms mean imputation for linearly related features.
+
+    When there is a simple linear relationship, and the missing values are not coincidentally the mean, iterative
+    imputer should do better than mean imputer.
+    """
```
bartpleiter: I do not fully understand this test 😅. Given the previous test with the expected values, does this test still add value? (e.g. could it fail if the previous test passes?)

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
bartpleiter: I think I do: it tests whether the warning is raised when only one feature is used, and makes sure there is output based on the initial strategy.

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
bartpleiter: Yes I do (as a logging warning). You probably have to adjust this assert to check if a warning has been thrown by the logger

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
fleur-petit: good point, will remove it.

packages/openstef-models/tests/unit/transforms/general/test_imputer.py
fleur-petit: good point, will add it to assert.

bartpleiter (approved):
Looks good to me!

PR #740, opened 2025-11-07
fix: Retraining data bug Liander 2024 benchmark

MvLieshout:


examples/benchmarks/liander_2024_benchmark_xgboost_gblinear.py
```diff
@@ -44,7 +44,6 @@
 # Model configuration
 FORECAST_HORIZONS = [LeadTime.from_string("P3D")]  # Forecast horizon(s)
 PREDICTION_QUANTILES = [Q(0.1), Q(0.3), Q(0.5), Q(0.7), Q(0.9)]  # Quantiles for probabilistic forecasts
-LAG_FEATURES = [timedelta(days=-7)]  # Lag features to include
```
MvLieshout, line 47: Was not used, so deleted to avoid confusion.

packages/openstef-beam/src/openstef_beam/backtesting/backtest_event_generator.py
```diff
@@ -128,7 +128,9 @@ def _train_iterator(self) -> Iterator[BacktestEvent]:
             if horizon_end > end_time:
                 break
 
-            training_start = current_time - self.forecaster_config.training_context_length
+            training_start = max(
+                current_time - self.forecaster_config.training_context_length, min(self.index).to_pydatetime()
+            )
```
egordm, line 134:
```suggestion
            training_start = max(
                current_time - self.forecaster_config.training_context_length,
                cast("pd.Series[pd.Timestamp]", self.index.min()).to_pydatetime()
            )
```

Using python min iterates over the whole series using python. Using pandas min, offloads it to internal C code which is faster. The difference will probably not be noticed, but it's probably a better practise.

The cast solves some type quirks, but might not be needed soon.

packages/openstef-beam/src/openstef_beam/backtesting/backtest_event_generator.py
```diff
@@ -149,6 +152,9 @@ def _calculate_coverage(self, start: datetime, end: datetime) -> float:
         Returns:
             The ratio of available data points to expected data points.
         """
+        if start >= end:
```
egordm: Not sure if this case should be possible. Probably some validation error needs to be thrown if EventGenerator is created with such configuration.

packages/openstef-beam/src/openstef_beam/benchmarking/benchmarks/liander2024.py
```diff
@@ -47,7 +47,7 @@ class Liander2024TargetProvider(SimpleTargetProvider[BenchmarkTarget, list[Liand
     targets_file_path: str = Field(default="liander2024_targets.yaml", init=False)
 
     data_start: datetime = Field(
-        default=datetime.fromisoformat("2024-02-01T00:00:00Z"),
+        default=datetime.fromisoformat("2024-01-01T00:00:00Z"),
```
egordm: Weather data is incomplete before februari for a lot of predictors due to OpenMeteo starting to collect versioned data at that time. We should keep it at 2 or some weird results / forecasts will happen.

I probably should have added this as comment.

packages/openstef-beam/src/openstef_beam/backtesting/backtest_event_generator.py
MvLieshout: Nice, thanks. The cast did not seem to solve the type issue for me. I have ignored it for now.

packages/openstef-beam/src/openstef_beam/backtesting/backtest_event_generator.py
MvLieshout: The clipping of the start date to the start of the index causes the start training date to be equal to the end training date for the first time. I agree that bigger than should not be possible, but equal to should be right? Since training on zero data does not make sense, I set the coverage to zero so there would not be a train event.

It would be cleaner to then just not calculate the coverage at all I think. So I altered the while-loop.

egordm (approved):
LGTM!

PR #741, opened 2025-11-07
Added Lightgbm, LightGBM Linear Trees and Hybrid Stacking Forecasters

Lars800:
Expanded openstef-models with additional forecasters including:

- LightGBM 
- LightGBM with Linear Trees
- Hybrid (Stacking Regressor with LightGBM and GBLinear)

The implementation closely follows the structure of existing forecasters and is fully compatible with openstef-beam etc. 
LightGBM and sklearn's StackingRegressor do no support multiquantile regression out of the box. To make them compatible two custom SKlearn-like estimators were implemented: LGBMQuantileRegressor and HybridQuantileRegressor. These classes wrap multiple single-quantile models together to enable mulit-quantile prediction. 

All new classes have testing comprable to existing classes. 
The preset forecasting_model was updated to run the benchmark for the new models.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,4 +129,10 @@ optimization_results/
 dev_sandbox/
 
 # Benchmark outputs
-benchmark_results*/
+benchmark_results/
+optimization_results/
+dev_sandbox/
+
+# Benchmark outputs
+benchmark_results*/optimization_results/
+dev_sandbox/
--- a/examples/benchmarks/liander_2024_benchmark_xgboost_gblinear.py
+++ b/examples/benchmarks/liander_2024_benchmark_xgboost_gblinear.py
@@ -49,27 +49,14 @@ N_PROCESSES = 12  # Amount of parallel processes to use for the benchmark
 
 # Model configuration
 FORECAST_HORIZONS = [LeadTime.from_string("P3D")]  # Forecast horizon(s)
-PREDICTION_QUANTILES = [
-    Q(0.05),
-    Q(0.1),
-    Q(0.3),
-    Q(0.5),
-    Q(0.7),
-    Q(0.9),
-    Q(0.95),
-]  # Quantiles for probabilistic forecasts
+PREDICTION_QUANTILES = [Q(0.1), Q(0.3), Q(0.5), Q(0.7), Q(0.9)]  # Quantiles for probabilistic forecasts
 
 BENCHMARK_FILTER: list[Liander2024Category] | None = None
 
-USE_MLFLOW_STORAGE = False
-
-if USE_MLFLOW_STORAGE:
-    storage = MLFlowStorage(
-        tracking_uri=str(OUTPUT_PATH / "mlflow_artifacts"),
-        local_artifacts_path=OUTPUT_PATH / "mlflow_tracking_artifacts",
-    )
-else:
-    storage = None
+storage = MLFlowStorage(
+    tracking_uri=str(OUTPUT_PATH / "mlflow_artifacts"),
+    local_artifacts_path=OUTPUT_PATH / "mlflow_tracking_artifacts",
+)
 
 common_config = ForecastingWorkflowConfig(
     model_id="common_model_",
@@ -85,6 +72,7 @@ common_config = ForecastingWorkflowConfig(
     temperature_column="temperature_2m",
     relative_humidity_column="relative_humidity_2m",
     energy_price_column="EPEX_NL",
+    verbosity=0,
 )
 
 xgboost_config = common_config.model_copy(update={"model": "xgboost"})
@@ -130,6 +118,13 @@ def _target_forecaster_factory(
                 }
             )
         )
+
+    return OpenSTEF4BacktestForecaster(
+        config=backtest_config,
+        workflow_factory=_create_workflow,
+        debug=False,
+        cache_dir=OUTPUT_PATH / "cache" / f"{context.run_name}_{target.name}",
+    )
 
     return OpenSTEF4BacktestForecaster(
         config=backtest_config,


packages/openstef-models/src/openstef_models/estimators/lgbm.py
MvLieshout, line 1: @egordm, do you have an opinion on this code structure? This introduces a folder to separate the LGBMRegressor from the LGBMQuantileRegressor.

packages/openstef-models/src/openstef_models/estimators/hybrid.py
```diff
@@ -0,0 +1,188 @@
+# SPDX-FileCopyrightText: 2025 Contributors to the OpenSTEF project <short.term.energy.forecasts@alliander.com>
+#
+# SPDX-License-Identifier: MPL-2.0
+"""Hybrid quantile regression estimators for multi-quantile forecasting.
+
+This module provides the HybridQuantileRegressor class, which combines LightGBM and linear models
+using stacking for robust multi-quantile regression, including serialization utilities.
```
MvLieshout, line 7: In theory, this could be any two multi-quantile regressors right? I think it would be nice to generalize this to accept any two regressors. However, I can imagine that is quite some work and not top prio right now. So might be nice to do that later on if there is time (or we can do it ourselves and just have this as a proof of concept). If we keep this is as an PoC, might be nice to move this change to a separate branch where we can keep it for reference.

packages/openstef-models/src/openstef_models/models/forecasting/lgblinear_forecaster.py
MvLieshout, line 1: Nitpick, but I think we should be consistent between `lightgbm` and `lgbm`. The last one has my preference.

I would then name the regressor LGBMLinearForecaster (note the caps) for consitency also with other forecasters.

packages/openstef-models/src/openstef_models/models/forecasting/lgblinear_forecaster.py
```diff
@@ -0,0 +1,329 @@
...
+
+    Note:
+        These parameters are optimized for probabilistic forecasting with
+        quantile regression. The default objective function is specialized
+        for magnitude-weighted pinball loss.
+    """
+
+    # Core Tree Boosting Parameters
+
+    n_estimators: int = Field(
```
MvLieshout, line 54: How did you pick the default values? In general we try to stick to the defaults that are already there.

packages/openstef-models/src/openstef_models/models/forecasting/lgblinear_forecaster.py
```diff
@@ -0,0 +1,329 @@
...
+    def hyperparams(self) -> LgbLinearHyperParams:
+        return self._config.hyperparams
+
+    @property
+    @override
+    def is_fitted(self) -> bool:
+        return self._lgblinear_model.__sklearn_is_fitted__()
+
+    @override
+    def fit(self, data: ForecastInputDataset, data_val: ForecastInputDataset | None = None) -> None:
```
MvLieshout, line 278: Could you take a look at the XGB forecaster, I see some differences in implemenation. The use of the `_prepare_fit_input()` function for example and the `np.asarray()` for the target. This might also take away the need to use the type ignores.

packages/openstef-models/src/openstef_models/models/forecasting/hybrid_forecaster.py
```diff
@@ -0,0 +1,142 @@
...
+    @override
+    def fit(self, data: ForecastInputDataset, data_val: ForecastInputDataset | None = None) -> None:
+        """Fit the Hybrid model to the training data.
+
+        Args:
+            data: Training data in the expected ForecastInputDataset format.
+            data_val: Validation data for tuning the model (optional, not used in this implementation).
+
+        """
+        input_data: pd.DataFrame = data.input_data()
```
MvLieshout: (Same as for lgblinear): Could you take a look at the XGB forecaster, I see some differences in implemenation. The use of the _prepare_fit_input() function for example and the np.asarray() for the target. This might also take away the need to use the type ignores.

packages/openstef-models/src/openstef_models/models/forecasting/hybrid_forecaster.py
MvLieshout, line 1: Just curious, do you have any ideas/seen any examples on how to make this forecaster explainable?

packages/openstef-models/tests/unit/models/forecasting/test_lightgbm_forecaster.py
```diff
@@ -0,0 +1,162 @@
...
+    # Forecast data quality
+    assert not result.data.isna().any().any(), "Forecast should not contain NaN or None values"
+
+    # Since forecast is deterministic with fixed random seed, check value spread (vectorized)
+    # All quantiles should have some variation (not all identical values)
+    stds = result.data.std()
+    assert (stds > 0).all(), f"All columns should have variation, got stds: {dict(stds)}"
+
+
+def test_lightgbm_forecaster__not_fitted_error(
```
MvLieshout: This test tests the same functionality as the one below right?

packages/openstef-models/tests/unit/models/forecasting/test_lgbm_forecaster.py
```diff
@@ -0,0 +1,162 @@
...
+    assert len(feature_importances.index) > 0
+
+    # Columns should match expected quantile formats
+    expected_columns = pd.Index([q.format() for q in base_config.quantiles], name="quantiles")
+    pd.testing.assert_index_equal(feature_importances.columns, expected_columns)
+
+    # Values should be normalized (sum to 1.0 per quantile column) and non-negative
+    col_sums = feature_importances.sum(axis=0)
+    pd.testing.assert_series_equal(col_sums, pd.Series(1.0, index=expected_columns), atol=1e-10)
+    assert (feature_importances >= 0).all().all()
```
MvLieshout, line 146: Might also be nice to add the tests with the different objectives. I am also okay with you leaving this as an TODO for us. Could you in that case leave a TODO in the code?

packages/openstef-models/tests/unit/models/forecasting/test_lgblinear_forecaster.py
```diff
@@ -0,0 +1,162 @@
...
+    # Forecast data quality
+    assert not result.data.isna().any().any(), "Forecast should not contain NaN or None values"
+
+    # Since forecast is deterministic with fixed random seed, check value spread (vectorized)
+    # All quantiles should have some variation (not all identical values)
+    stds = result.data.std()
+    assert (stds > 0).all(), f"All columns should have variation, got stds: {dict(stds)}"
+
+
+def test_lgblinear_forecaster__not_fitted_error(
```
MvLieshout: Same as for lightgbm, this test is duplicated?

packages/openstef-models/tests/unit/models/forecasting/test_lgbmlinear_forecaster.py
```diff
@@ -0,0 +1,162 @@
...
+    assert len(feature_importances.index) > 0
+
+    # Columns should match expected quantile formats
+    expected_columns = pd.Index([q.format() for q in base_config.quantiles], name="quantiles")
+    pd.testing.assert_index_equal(feature_importances.columns, expected_columns)
+
+    # Values should be normalized (sum to 1.0 per quantile column) and non-negative
+    col_sums = feature_importances.sum(axis=0)
+    pd.testing.assert_series_equal(col_sums, pd.Series(1.0, index=expected_columns), atol=1e-10)
+    assert (feature_importances >= 0).all().all()
```
MvLieshout, line 146: Same as for lightgbm: add tests for different loss functions or leave as TODO for us in the code.

packages/openstef-models/src/openstef_models/estimators/lgbm.py
MvLieshout: Also, I was wondering about if we could not use any inheritance from classes we have defined?

packages/openstef-models/src/openstef_models/models/forecasting/lgblinear_forecaster.py
```diff
@@ -0,0 +1,329 @@
...
+            min_data_in_leaf=config.hyperparams.min_data_in_leaf,
+            min_data_in_bin=config.hyperparams.min_data_in_bin,
+            reg_alpha=config.hyperparams.reg_alpha,
+            reg_lambda=config.hyperparams.reg_lambda,
+            num_leaves=config.hyperparams.num_leaves,
+            max_bin=config.hyperparams.max_bin,
+            colsample_bytree=config.hyperparams.colsample_bytree,
+            random_state=config.hyperparams.random_state,
+            early_stopping_rounds=config.hyperparams.early_stopping_rounds,
+            verbosity=config.verbosity,
```
egordm: Would it make sense to pass the hyperparams object directly instead of reassigning each hyperparam again?

packages/openstef-models/src/openstef_models/models/forecasting/lightgbm_forecaster.py
```diff
@@ -0,0 +1,329 @@
...
+            min_child_weight=config.hyperparams.min_child_weight,
+            min_data_in_leaf=config.hyperparams.min_data_in_leaf,
+            min_data_in_bin=config.hyperparams.min_data_in_bin,
+            reg_alpha=config.hyperparams.reg_alpha,
+            reg_lambda=config.hyperparams.reg_lambda,
+            num_leaves=config.hyperparams.num_leaves,
+            max_bin=config.hyperparams.max_bin,
+            colsample_bytree=config.hyperparams.colsample_bytree,
+            random_state=config.hyperparams.random_state,
+            early_stopping_rounds=config.hyperparams.early_stopping_rounds,
```
egordm: Same here. Maybe we should pass the hyperparams object directly to avoid reassignment.

packages/openstef-models/src/openstef_models/estimators/hybrid.py
```diff
@@ -0,0 +1,188 @@
...
+            final_estimator = QuantileRegressor(quantile=q)
+
+            self._models.append(
+                StackingRegressor(
+                    estimators=[("lightgbm", lightgbm_model), ("gblinear", linear)],  # type: ignore
+                    final_estimator=final_estimator,
+                    verbose=3,
+                    passthrough=False,
+                    n_jobs=None,
+                    cv=2,
```
egordm: Should cv and verbose be configurable (hyper)parameters?

packages/openstef-models/src/openstef_models/models/forecasting/hybrid_forecaster.py
```diff
@@ -0,0 +1,142 @@
...
+
+MODEL_CODE_VERSION = 2
+
+
+class HybridForecasterState(BaseConfig):
+    """Serializable state for Hybrid forecaster persistence."""
+
+    version: int = Field(default=MODEL_CODE_VERSION, description="Version of the model code.")
+    config: HybridForecasterConfig = Field(..., description="Forecaster configuration.")
+    model: str = Field(..., description="Base64-encoded serialized Hybrid model.")
```
egordm: State for serialization is no longer necessary. Using something like pickle directly is now preferred.

packages/openstef-models/src/openstef_models/models/forecasting/lgblinear_forecaster.py
```diff
@@ -0,0 +1,329 @@
...
+    """Serializable state for LgbLinear forecaster persistence.
+
+    Contains all information needed to restore a trained LgbLinear model,
+    including configuration and the serialized model weights. Used for
+    model saving, loading, and version management in production systems.
+    """
+
+    version: int = Field(default=MODEL_CODE_VERSION, description="Version of the model code.")
+    config: LgbLinearForecasterConfig = Field(..., description="Forecaster configuration.")
+    model: str = Field(..., description="Base64-encoded serialized LgbLinear model.")
```
egordm: State for serialization is no longer necessary. Using something like pickle directly is now preferred.

packages/openstef-models/src/openstef_models/models/forecasting/lightgbm_forecaster.py
```diff
@@ -0,0 +1,329 @@
...
+    """Serializable state for LightGBM forecaster persistence.
+
+    Contains all information needed to restore a trained LightGBM model,
+    including configuration and the serialized model weights. Used for
+    model saving, loading, and version management in production systems.
+    """
+
+    version: int = Field(default=MODEL_CODE_VERSION, description="Version of the model code.")
+    config: LightGBMForecasterConfig = Field(..., description="Forecaster configuration.")
+    model: str = Field(..., description="Base64-encoded serialized LightGBM model.")
```
egordm: State for serialization is no longer necessary. Using something like pickle directly is now preferred.

Lars800: Removed State

packages/openstef-models/src/openstef_models/estimators/lgbm.py
Lars800: There are 2 reasons having the estimator file separate from the Forecaster class. 

The primary reason is to mimic other Forecaster classes as closely as possible. XGBoost and GBLinear call the external xgboost library for their estimator. LightGBM and SKLearn unfortunately do not natively support Multi Quantile Regression, hence the need to manually implement this feature. To avoid clutter in the Forecaster classes it seems logical to implement this in a separate Estimator class that mimics an SKLearn Style estimator like LGBMRegressor and XGBoostRegressor. 

The reason to put the Estimator classes in a separate folder is to avoid the duplication of code. LGBMLinear and LGBM both use the LGBMQuantileRegressor as their estimator. In this case we can choose to 1. duplicate the code, 2. let one file import from the other or 3. put the estimator somewhere else. 3 seemed most logical to me, but Im open to implement any other option

packages/openstef-models/src/openstef_models/models/forecasting/hybrid_forecaster.py
Lars800: Yes! We could do a weighted sum of the feature importances of the base learners, based on the parameters of the final estimators. This is experimental, but something we can put on the _Nice to Have List_

packages/openstef-models/src/openstef_models/models/forecasting/lgblinear_forecaster.py
Lars800: Optimal values are a WIP. I will set these to the Docs default and store optimal values in a separate place.

Lars800: Exception is max_depth, this is unbounded by default but that makes a big mess in our usecase

packages/openstef-models/tests/unit/models/forecasting/test_lgbm_forecaster.py
Lars800: Which different objectives would you like to include? LightGBM does not support any relevant objectives other than Quantile for out use-case. We could look into defining custom objective functions, but this would be quite cumbersome. 
https://lightgbm.readthedocs.io/en/stable/Parameters.html#parameters-format
```
custom objective function (gradients and hessians not computed directly by LightGBM)
custom
must be passed through parameters explicitly in the C API
Note: cannot be used in CLI version
```

Lars800: Added TODOs for LGBM and LGBMLinear

packages/openstef-models/src/openstef_models/models/forecasting/lgblinear_forecaster.py
Lars800: This does make sense from a readability perspective, but this makes the Estimator class less SKLearn like. Two potential solutions: 

1. Pass **hyperparams, no change needed to estimator class but comes at a small penalty of transparancy.
```
        self._lgbmlinear_model = LGBMQuantileRegressor(
            quantiles=[float(q) for q in config.quantiles],
            linear_tree=True,
            random_state=config.random_state,
            early_stopping_rounds=config.early_stopping_rounds,
            verbosity=config.verbosity,
            **config.hyperparams.model_dump(),
        )
```

2. create a .from_hyperparams() class method on the estimator classes

I have implemented 1 for now. Can implement 2 if needed

PR #751, opened 2025-11-17
Kalman filter

Guilly-Kolkman:
Implemented Kalman filter as both a pre and post processing option with tests.

- Basekalman
- KalmanPreprocessor
- KalmanPostprocessor

The implementation follows the sktime implementation of KalmanFilterTransformerFP build on filterpy. The kalman filter reduces noise and improves results. However, the method is computationally more expensive.

For the preprocessor you can add it in the preprocessing list in workflows of models with, for example:
`            preprocessing = [
            *checks,
            *feature_adders,
            HolidayFeatureAdder(country_code=config.location.country_code),
            DatetimeFeaturesAdder(onehot_encode=False),
            KalmanPreprocessor(
                selection=config.kalman_features,
            ),
            *feature_standardizers,
        ],`
The kalman_features are currently based only on weather features.

For postprocessing a flag is added which can be set to True or False depending on the user's need.


packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
```diff
@@ -145,6 +156,31 @@ class ForecastingWorkflowConfig(BaseConfig):  # PredictionJob
         default=[],
         description="If not None, rolling aggregate(s) of load will be used as features in the model.",
     )
+    kalman_features: FeatureSelection = Field(
+        default=FeatureSelection(
```
MvLieshout, line 158: How did you specifically choose these? If you want to just select all of them, you could also use `FeatureSelection.ALL`.

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
```diff
@@ -280,7 +322,7 @@ def create_forecasting_workflow(config: ForecastingWorkflowConfig) -> CustomFore
                 hyperparams=config.xgboost_hyperparams,
             )
         )
-        postprocessing = [QuantileSorter()]
+        postprocessing: list[Transform[ForecastDataset, ForecastDataset]] = [QuantileSorter()]
```
MvLieshout: I do not think the typehint is necessary here?

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
```diff
@@ -271,6 +310,9 @@ def create_forecasting_workflow(config: ForecastingWorkflowConfig) -> CustomFore
             *feature_adders,
             HolidayFeatureAdder(country_code=config.location.country_code),
             DatetimeFeaturesAdder(onehot_encode=False),
+            KalmanPreprocessor(
```
MvLieshout: Is there a reason to only add it to xgboost? Did it not perform well in gblinear?

Also whether we want to add it to the preset, depends on the results also. I would probably leave it out (for now).

packages/openstef-models/src/openstef_models/transforms/general/kalman_filter.py
```diff
@@ -0,0 +1,150 @@
...
+        if not features_list:
+            return df
+
+        kf = KalmanFilterTransformerFP(state_dim=len(features_list))
+        out = df.copy(deep=True)
+        out[features_list] = kf.fit_transform(X=df[features_list])  # type: ignore[assignment]
+        return out
+
+
+# Preprocessor: operates on TimeSeriesDataset
```
MvLieshout: This comment can be removed.

packages/openstef-models/src/openstef_models/transforms/general/kalman_filter.py
```diff
@@ -0,0 +1,150 @@
...
+class KalmanPreprocessor(BaseKalman, TimeSeriesTransform):
+    """Apply Kalman Smoothing to time series data to reduce noise and improve temporal consistency.
```
MvLieshout, line 49: Is there a specific source you used for the implementation? If so, I think it is nice to add it to the docs.

packages/openstef-models/src/openstef_models/transforms/general/kalman_filter.py
```diff
@@ -0,0 +1,150 @@
...
+This class provides implementations of Kalman Smoothing as both a preprocessor
+and postprocessor for time series datasets. The Kalman Smoother helps reduce noise
+"""
+
+from collections.abc import Iterable
+from typing import override
+
+import pandas as pd
+from pydantic import Field
+from sktime.transformations.series.kalman_filter import (
```
MvLieshout, line 16: Can you add this as a dependency? I don't think we have it already.

packages/openstef-models/src/openstef_models/transforms/general/kalman_filter.py
```diff
@@ -0,0 +1,150 @@
...
+        df_sm = self._run_kalman_smoother(df, features)
+        return data.copy_with(data=df_sm, is_sorted=True)
+
+    @override
+    def features_added(self) -> list[str]:
+        # Preprocessor doesn't add columns
+        return []
+
+
+# Postprocessor: operates on ForecastDataset quantiles
```
MvLieshout: Can also be removed.

packages/openstef-models/src/openstef_models/transforms/general/kalman_filter.py
```diff
@@ -0,0 +1,150 @@
...
+    """Base class for Kalman Smoothing transforms."""
+
+    selection: FeatureSelection = Field(default=FeatureSelection.ALL, description="Columns to smooth")
+    state_dim: int = Field(
+        default=1,
+        description="Kalman filter state dimension (1 = per-column independent)",
+    )
+
+    @staticmethod
+    def _run_kalman_smoother(df: pd.DataFrame, features: Iterable[str]) -> pd.DataFrame:
```
MvLieshout: I think the terms smoother and filter are not actually the same thing right? If I understand correctly the term filter is used for only looking backwards and smoother is used if both past and future data is used. I think it is best to stick to the filter term as we use only past data to avoid confusion.

packages/openstef-models/tests/unit/transforms/general/test_kalman.py
```diff
@@ -0,0 +1,160 @@
...
+    return ForecastDataset(forecast_data, timedelta(hours=1))
+
+
+def test_kalman_preprocessor(sample_dataset: TimeSeriesDataset):
+    """Test KalmanPreprocessor on sample dataset."""
+    dataset = sample_dataset
+    transform = KalmanPreprocessor()
+    result = transform.fit_transform(dataset)
+
+    expected_values = pd.DataFrame(
```
MvLieshout, line 55: It would also be nice to add a tests for the multiple column case where all columns are transformed.

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
floriangoethals: For the preprocessing, we primarily focused on weather data. We can add a default to the feature selection weather (FeatureSelection.WEATHER).

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
floriangoethals: We added this for convenience, we will take it out. Good you noticed, thank you.

packages/openstef-models/src/openstef_models/transforms/general/kalman_filter.py
floriangoethals: added as dependency of openstef models

packages/openstef-models/src/openstef_models/transforms/general/kalman_filter.py
Lars800: The implementation was done to mimic other transforms. No additional sources were used

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
MvLieshout: Ah okay, I think it is fine as it is now.

PR #765, opened 2025-11-20
Research/v4.1.0 additional forecasters

Lars800:
V2 of the PR to introduce HybridForecaster, LGBMForecaster and LGBMLinearForecaster

These three new forecasters expand on existing OpenSTEF forecasters with new algorithms. 
Comments on initial PR were resolved, all forecaster specific logic is contained within the forecaster.py file. 

To enable multi quantile regression for SKlearn-like estimators, the MultiQuantileRegressor was introduced as a utility class. 
This provides a wrapper arround SKLearn with multi-quantile capabilities

### DIFF
--- a/packages/openstef-models/pyproject.toml
+++ b/packages/openstef-models/pyproject.toml
@@ -28,9 +28,8 @@ classifiers = [
 ]
 
 dependencies = [
-  "lightgbm>=4.6",
-  "skops>=0.13",
   "holidays>=0.79",
+  "lightgbm>=4.6",
   "mlflow-skinny>=3,<4",
   "openstef-beam>=4.0.0.dev0,<5",
   "openstef-core>=4.0.0.dev0,<5",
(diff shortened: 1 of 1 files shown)


packages/openstef-models/src/openstef_models/models/forecasting/lgbmlinear_forecaster.py
```diff
@@ -0,0 +1,336 @@
...
+            "n_jobs": config.n_jobs,
+            **config.hyperparams.model_dump(),
+        }
+
+        self._lgbmlinear_model: MultiQuantileRegressor = MultiQuantileRegressor(
+            base_learner=LGBMRegressor,  # type: ignore
+            quantile_param="alpha",
+            hyperparams=lgbmlinear_params,
+            quantiles=[float(q) for q in config.quantiles],
+        )
```
egordm, line 252: Not really a code related question. But LightGBM model doesn't require target variable normalization? We found that it makes a big difference in linear models.

Lars800: Good remark! Perhaps it will help with the stability of the linear trees!

PR #771, opened 2025-11-28
OpenSTEF Meta V0.1

Lars800:
Introducting OpenSTEF-Meta, a one-stop-shop for all things meta learning. 

This sub-package introduces four common meta learning algorithms:

**Residual Forecaster**
- 2 stage Forecasting Model
- Primary model is fitted as usual 
- Secondary model fitted on residuals

**Stacking Forecaster**
- Multiple Base Forecasters
- Single Regressor is fitted on Base Predictions

**Learned Weights Forecaster**
- Multiple Base Forecasters
- Classification model learns optimal model weights

**Rules Forecaster**
- Multiple Base Forecasters
- Pre-defined rules on how to combine base methods.
 ---
**This is an initial implementation**
- The methods have been tested on the Liander 2024 Huggingface dataset. 
- The results are accurate and improve over existing LGBM, XGBoost and GBLinear models
- The code can still be further optimized for efficiency


examples/benchmarks/liander_2024_ensemble.py
```diff
@@ -0,0 +1,158 @@
+"""Liander 2024 Benchmark Example.
+
+====================================
+
+This example demonstrates how to set up and run the Liander 2024 STEF benchmark using OpenSTEF BEAM.
+The benchmark will evaluate XGBoost and GBLinear models on the dataset from HuggingFace.
+"""
```
egordm, line 7: Examples in individual packages are a bit hard to discover. Can we move it in examples/benchmarks? And if other more functional examples are needed they can go in examples/examples/metalearning or something.

packages/openstef-meta/src/openstef_meta/framework/final_learner.py
```diff
@@ -0,0 +1,149 @@
...
+    "shortwave_radiation",
+    "direct_radiation",
+    "diffuse_radiation",
+    "direct_normal_irradiance",
+    "load",
+}
+
+SELECTOR = Selector(
+    selection=FeatureSelection(include=WEATHER_FEATURES),
+)
```
egordm: We should probably avoid hardcoding these defaults since they may change from benchmark to benchmark. Instead we should make them configurable.

This way, we can set them automatically in presets and allow users to redefine them.

packages/openstef-meta/src/openstef_meta/transforms/selector.py
```diff
@@ -0,0 +1,50 @@
...
+from pydantic import Field
+
+from openstef_core.base_model import BaseConfig
+from openstef_core.datasets import TimeSeriesDataset
+from openstef_core.datasets.validated_datasets import ForecastInputDataset
+from openstef_core.transforms import TimeSeriesTransform
+from openstef_models.utils.feature_selection import FeatureSelection
+
+
+class Selector(BaseConfig, TimeSeriesTransform):
```
egordm: Possibly a quick win to add this one already in openstef 4 release. (and put it in openstef-models transforms or openstef-core transforms). Makes this pr smalled.

packages/openstef-meta/src/openstef_meta/framework/final_learner.py
Lars800: Will Remobe hard coded

packages/openstef-beam/src/openstef_beam/backtesting/backtest_forecaster/openstef4_backtest_forecaster.py
```diff
@@ -40,6 +42,10 @@ class OpenSTEF4BacktestForecaster(BaseModel, BacktestForecasterMixin):
         default=False,
         description="When True, saves intermediate input data for debugging",
     )
+    contributions: bool = Field(
+        default=False,
+        description="When True, saves intermediate input data for explainability",
```
MvLieshout: I feel like this description is a bit vague.
```suggestion
        description="When True, saves intermediate prediction contributions of input data for explainability",
```

packages/openstef-models/src/openstef_models/integrations/mlflow/mlflow_storage_callback.py
```diff
@@ -97,11 +98,16 @@ def on_fit_end(
...
+            raise NotImplementedError(
+                "MLFlowStorageCallback does not yet support EnsembleForecastingWorkflow model storage."
+            )
+
         # Create a new run
         run = self.storage.create_run(
             model_id=context.workflow.model_id,
             tags=context.workflow.model.tags,
-            hyperparams=context.workflow.model.forecaster.hyperparams,
+            hyperparams=context.workflow.model.forecaster.hyperparams,  # type: ignore TODO Make MLFlow compatible with OpenSTEF Meta
```
egordm: We should probably address this before merging. 

It's mostly in hyperparams if I understand it correctly? Since the rest is pickling.

packages/openstef-models/src/openstef_models/explainability/mixins.py
```diff
@@ -44,6 +45,19 @@ def feature_importances(self) -> pd.DataFrame:
...
+
+        Args:
+            data: Input dataset for which to compute feature contributions.
+            scale: Whether to scale contributions to sum to the prediction value.
+
+        Returns:
+            DataFrame with contributions per base learner.
+        """
+        raise NotImplementedError
+
```
egordm: In openstef-meta there is a ContributionsMixin. Should we remove it in favor of this method?

packages/openstef-beam/src/openstef_beam/backtesting/backtest_forecaster/openstef4_backtest_forecaster.py
```diff
@@ -50,6 +56,8 @@ class OpenSTEF4BacktestForecaster(BaseModel, BacktestForecasterMixin):
     def model_post_init(self, context: Any) -> None:
         if self.debug:
```
MvLieshout: Nitpick: if statements could be combined
```suggestion
        if self.debug or self.contributions:
             self.cache_dir.mkdir(parents=True, exist_ok=True)
```

packages/openstef-models/src/openstef_models/models/forecasting/flatliner_forecaster.py
```diff
@@ -102,3 +102,15 @@ def feature_importances(self) -> pd.DataFrame:
             index=["load"],
             columns=[quantile.format() for quantile in self.config.quantiles],
         )
+
+    @override
+    def predict_contributions(self, data: ForecastInputDataset, *, scale: bool = True) -> pd.DataFrame:
+
+        if scale:
+            pass
```
egordm: The 'if scale' can probably be removed. `@override` already helps us avoid linter complaning about an unused argument.

packages/openstef-models/src/openstef_models/models/forecasting/gblinear_forecaster.py
```diff
@@ -326,6 +326,47 @@ def predict(self, data: ForecastInputDataset) -> ForecastDataset:
...
+        Returns:
+            DataFrame with contributions per base learner.
+        """
+        # Get input features for prediction
+        input_data: pd.DataFrame = data.input_data(start=data.forecast_start)
+        xgb_input: xgb.DMatrix = xgb.DMatrix(data=input_data)
+
+        # Generate predictions
+        booster = self._gblinear_model.get_booster()
+        predictions_array: np.ndarray = booster.predict(xgb_input, pred_contribs=True, strict_shape=True)[:, :, :-1]
```
egordm: Nice! Didn't know xgboost had this feature. TIL.

packages/openstef-beam/src/openstef_beam/backtesting/backtest_forecaster/openstef4_backtest_forecaster.py
```diff
@@ -58,6 +66,11 @@ def quantiles(self) -> list[Q]:
         if self._workflow is None:
             self._workflow = self.workflow_factory()
         # Extract quantiles from the workflow's model
+
+        if isinstance(self._workflow.model, EnsembleForecastingModel):
+            # Assuming all ensemble members have the same quantiles
```
MvLieshout: Do we also enforce this?

Then we do not have to assume I guess

packages/openstef-models/src/openstef_models/workflows/custom_forecasting_workflow.py
```diff
@@ -153,6 +154,10 @@ def fit(
 
             result = self.model.fit(data=data, data_val=data_val, data_test=data_test)
 
+            if isinstance(result, EnsembleModelFitResult):
+                self._logger.info("Discarding EnsembleModelFitResult for compatibility.")
+                result = result.combiner_fit_result
```
egordm: Nitpick. But I think using log level info might be too verbose. debug would be more appropriate, or no logging at all, since it's not really actionable.

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
```diff
@@ -100,9 +100,9 @@ class ForecastingWorkflowConfig(BaseConfig):  # PredictionJob
     model_id: ModelIdentifier = Field(description="Unique identifier for the forecasting model.")
 
     # Model configuration
-    model: Literal["xgboost", "gblinear", "flatliner", "hybrid", "lgbm", "lgbmlinear"] = Field(
-        description="Type of forecasting model to use."
-    )  # TODO(#652): Implement median forecaster
+    model: Literal[
+        "xgboost", "gblinear", "flatliner", "stacking", "residual", "learned_weights", "lgbm", "lgbmlinear"
+    ] = Field(description="Type of forecasting model to use.")  # TODO(#652): Implement median forecaster
```
egordm: We already have a preset for metamodels in openstef-meta package. Would it make sense to keep them separate so we don't have to sink two presets on changes to one ensemble model?

packages/openstef-models/src/openstef_models/explainability/mixins.py
egordm: Or would it make sense to add ContributionsMixin here and have explainable forecaster extend it. (or have some kind of inbetween protocal approach). So we have a standard interface.

packages/openstef-meta/src/openstef_meta/transforms/selector.py
egordm: It would be very valuable to move this to openstef-models since it's pretty general purpose.

examples/benchmarks/liander_2024_ensemble.py
egordm: I think this one is fixed. Thanks.

packages/openstef-meta/src/openstef_meta/models/ensemble_forecasting_model.py
```diff
@@ -0,0 +1,736 @@
...
+            data: Raw time series dataset to prepare for forecasting.
+            forecast_start: Optional start time for forecasts. If provided and earlier
+                than the cutoff time, overrides the cutoff for data filtering.
+            forecaster_name: Name of the forecaster for which to prepare input data.
+
+        Returns:
+            Processed forecast input dataset ready for model prediction.
+        """
+        logger.debug("Preparing input data for forecaster '%s'.", forecaster_name)
+        input_data = restore_target(dataset=data, original_dataset=data, target_column=self.target_column)
```
MvLieshout: Not sure if this does anything here since the preprocessing is done after, but I might be missing something.

packages/openstef-meta/src/openstef_meta/presets/forecasting_workflow.py
```diff
@@ -0,0 +1,517 @@
...
+            forecaster_preprocessing[model_type] = [
+                SampleWeighter(
+                    target_column=config.target_column,
+                    weight_exponent=config.forecaster_sample_weight_exponent[model_type],
+                    weight_floor=config.sample_weight_floor,
+                    weight_scale_percentile=config.sample_weight_scale_percentile,
+                ),
+                Selector(
+                    selection=FeatureSelection(
+                        exclude={  # Fix hardcoded lag features should be replaced by a LagsAdder classmethod
```
MvLieshout: Can you add a TODO: to comments such as these? This makes it easier to find what we need to fix
```suggestion
                        exclude={  # TODO: Fix hardcoded lag features should be replaced by a LagsAdder classmethod
```

packages/openstef-meta/src/openstef_meta/presets/forecasting_workflow.py
```diff
@@ -0,0 +1,517 @@
...
+                    weight_floor=config.sample_weight_floor,
+                    weight_scale_percentile=config.sample_weight_scale_percentile,
+                ),
+            ]
+        else:
+            msg = f"Unsupported base model type: {model_type}"
+            raise ValueError(msg)
+
+    # Build combiner
+    if config.ensemble_type == "learned_weights":
```
MvLieshout: Possible improvement for readability: using a match case here. Can also be left as TODO (but leaving a comment would be nice then).

packages/openstef-meta/src/openstef_meta/utils/datasets.py
```diff
@@ -0,0 +1,282 @@
...
+    @property
+    def target_series(self) -> pd.Series | None:
+        """Return the target series if available."""
+        if self.target_column in self.data.columns:
+            return self.data[self.target_column]
+        return None
+
+    @staticmethod
+    def get_learner_and_quantile(feature_names: pd.Index) -> tuple[list[str], list[Quantile]]:
+        """Extract base learner names and quantiles from feature names.
```
MvLieshout: To stay in line with the rest, base learner -> forecaster.

Also saw this in some other docstrings.

```suggestion
        """Extract forecaster names and quantiles from feature names.
```

packages/openstef-meta/src/openstef_meta/presets/forecasting_workflow.py
Lars800: Nice I had never used this before, good addition. A lot more readable, although a few more lines replicated

packages/openstef-meta/src/openstef_meta/models/ensemble_forecasting_model.py
Lars800: Good catch, this line is supposed to be after the preprocessing (based on openstef Models)

packages/openstef-beam/src/openstef_beam/backtesting/backtest_forecaster/openstef4_backtest_forecaster.py
Lars800: We do not enforce it explicitly, but if the Forecasting Workflow (Config) is used, it is always the case. I have removed the inline comment. If statement still applies, as ensemble forecasting workflow does not have a .forecaster property. Alternatively we can define a .forecaster property on Ensemble Forecasting Model that returns self.forecasters[0] (To ensure compatibility without If statements with beam etc.

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
Lars800: Yes! I missed this in the migration, cleaning up now

packages/openstef-models/src/openstef_models/workflows/custom_forecasting_workflow.py
Lars800: I will change the log level. I do think it would be nice to keep the full EnsembleModelFit result in the future, to really evaluate the setup. 

current setup:
```
EnsembleModelFitResults:
    forecaster_results: ModelFitResult[] // Performance of base forecasters
    combiner_results: ModelFitResult // Performance of forecasters + combiner
```
Right now the combiner fit result gives the total performance after applying the combiner models. Ideally we would separate this. We would get something like this:

```
EnsembleModelFitResults:
    forecaster_results: ModelFitResult[] // Performance of base forecasters
    combiner_results: ModelFitResult // Added performance of combiner 
    full_results: ModelFitResult // Performance of forecasters + combiner
```

This off course implies extra work for the callbacks etc to make everyting compatible, so for now we discard larger structure and keep only the total result.

packages/openstef-models/src/openstef_models/explainability/mixins.py
Lars800: The ContributionsMixIn was an approach I tried, the current setup feels more natural. Ill cleanup the unused code

packages/openstef-models/src/openstef_models/integrations/mlflow/mlflow_storage_callback.py
Lars800: Yes, this is required for use in production as I understand it. I left this integration un resolved as I do not fully understand this part of the package. 

The primary issue here is EnsembleForecastingModel does not have a Forecaster attribute. So forecaster.hyperparams is unavailable. Depending on what they are used for downstream, we can either:
- Pass combiner.hyperparams
- Pass a dictionary of hyperparams, something like
```
{
'forecaster_model_1' : Hyperparams()
'forecaster_model_2' : Hyperparams()
'combiner' : Hyperparams()
}
```

This way everything is neatly saved, but it does require additional changes

packages/openstef-models/src/openstef_models/models/forecasting/lgbm_forecaster.py
MvLieshout, line 1: `colsample_bytree` not used in this file?

packages/openstef-models/pyproject.toml
```diff
@@ -29,6 +29,7 @@ classifiers = [
 
 dependencies = [
   "holidays>=0.79",
+  "lightgbm>=4.6",
```
egordm: Just like xgboost, we should probably make lightgbm an optional dependency. (at least for openstef-models)

packages/openstef-core/src/openstef_core/datasets/validated_datasets.py
```diff
@@ -429,8 +431,217 @@ def __init__(
...
+            EnsembleForecastDataset combining all input datasets.
+        """
+        ds1 = next(iter(datasets.values()))
+        additional_columns: dict[str, pd.Series] = {}
+        if isinstance(ds1.target_series, pd.Series):
+            additional_columns[ds1.target_column] = ds1.target_series
+        elif target_series is not None:
+            additional_columns[ds1.target_column] = target_series
+
+        sample_weight_column = "sample_weight"
```
egordm, line 541: Not sure if we want to hardcode this.

packages/openstef-core/src/openstef_core/datasets/validated_datasets.py
```diff
@@ -429,8 +431,217 @@ def __init__(
...
+            quantile=quantile,
+        )
+        prediction_data[self.target_column] = target
+        return ForecastInputDataset(
+            data=prediction_data,
+            sample_interval=self.sample_interval,
+            target_column=self.target_column,
+            forecast_start=self.forecast_start,
+        )
+
```
egordm: Not 100% sure if we want this logic to be in a dataset. An argument to be made is that this is business logic of metamodels. Putting it in a dataset makes assumptions and locks that dataset to always function this way and influences forecaster's logic.

packages/openstef-beam/src/openstef_beam/benchmarking/baselines/openstef4.py
```diff
@@ -71,8 +89,15 @@ def quantiles(self) -> list[Q]:
...
+
         # Extract quantiles from the workflow's model
-        return self._workflow.model.forecaster.config.quantiles
+        if isinstance(self._workflow.model, EnsembleForecastingModel):
+            name = self._workflow.model.forecaster_names[0]
+            return self._workflow.model.forecasters[name].config.quantiles
+        if self._workflow.model.forecaster is not None:
+            return self._workflow.model.forecaster.config.quantiles
+        msg = f"Cannot determine quantiles from model type {type(self._workflow.model)}"
+        raise TypeError(msg)
```
egordm: This should be easier. Perhaps we should add an overridable property on the Forecaster

packages/openstef-beam/src/openstef_beam/benchmarking/baselines/openstef4.py
```diff
@@ -136,6 +163,12 @@ def predict(self, data: RestrictedHorizonVersionedTimeSeries) -> TimeSeriesDatas
             predict_data.to_parquet(path=self.cache_dir / f"debug_{id_str}_predict.parquet")
             forecast.to_parquet(path=self.cache_dir / f"debug_{id_str}_forecast.parquet")
 
+        if self.contributions and isinstance(self._workflow.model, EnsembleForecastingModel):
+            id_str = data.horizon.strftime("%Y%m%d%H%M%S")
+            contributions = self._workflow.model.predict_contributions(predict_data, forecast_start=data.horizon)
+            df = pd.concat([contributions, forecast.data.drop(columns=["load"])], axis=1)
+
+            df.to_parquet(path=self.cache_dir / f"contrib_{id_str}_predict.parquet")
```
egordm: Here as well. This could be generic since base models can also have contributions.

packages/openstef-meta/src/openstef_meta/models/forecast_combiners/forecast_combiner.py
```diff
@@ -0,0 +1,148 @@
...
+
+class ForecastCombinerConfig(BaseConfig):
+    """Hyperparameters for the Final Learner."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    hyperparams: HyperParams = Field(
+        description="Hyperparameters for the final learner.",
+    )
+
```
egordm: We are locking hyperparams down here to a generic types. That means pydantic won't be able to validate the subtypes for custom combiners.

packages/openstef-meta/src/openstef_meta/models/forecast_combiners/learned_weights_combiner.py
```diff
@@ -0,0 +1,453 @@
...
+from openstef_meta.models.forecast_combiners.forecast_combiner import (
+    ForecastCombiner,
+    ForecastCombinerConfig,
+)
+from openstef_meta.utils.datasets import combine_forecast_input_datasets
+
+logger = logging.getLogger(__name__)
+
+
+Classifier = LGBMClassifier | XGBClassifier | LogisticRegression | DummyClassifier
```
egordm: Using closed types like this makes it hard to add a new type of classifier. We should prefer some base type approach.

packages/openstef-meta/src/openstef_meta/models/forecast_combiners/stacking_combiner.py
```diff
@@ -0,0 +1,260 @@
...
+    GBLinearHyperParams,
+)
+from openstef_models.models.forecasting.lgbm_forecaster import LGBMHyperParams
+
+if TYPE_CHECKING:
+    from openstef_models.models.forecasting.forecaster import Forecaster
+
+logger = logging.getLogger(__name__)
+
+ForecasterHyperParams = GBLinearHyperParams | LGBMHyperParams
```
egordm: Same here. We are constraining the possible types to only two.

egordm: Fixed in followup

PR #773, opened 2025-12-02
feature: added use_median option to flatliner forecaster so it predic…

bartpleiter:
…ts the median of the training data.


lschilders (approved):
LGTM! I'd say `predict_median` or similar is a better name for the attribute in the configuration, but I leave that up to you.

PR #777, opened 2025-12-05
feature: nonzero flatliner preset

lschilders:


packages/openstef-models/src/openstef_models/models/component_splitting/linear_component_splitter.py
lschilders, line 1: Already present [here](https://github.com/OpenSTEF/openstef/blob/052d4c670f984408ddfa6b5dc9a70425208c31d2/packages/openstef-models/src/openstef_models/models/component_splitting_model.py#L99)

PR #781, opened 2025-12-09
feature: check for model config change and skip model selection

lschilders:


egordm (changes requested):
Equality check between objects may be unstable since it's possible that same data but different instance may not pass the equality check.

I would suggest adding something like versioning. Setting version tag on mlflow models. Before loading the model you can check it against the mlflow run metadata.

As an extra we can also add try catch around model loading so if there are some breaking changes in openstef it doesn't jam up training new model. Shouldn't happen if migration to models is added, but still a good to have.

packages/openstef-models/src/openstef_models/integrations/mlflow/mlflow_storage_callback.py
```diff
@@ -174,19 +180,34 @@ def _run_model_selection(self, workflow: CustomForecastingWorkflow, result: Mode
...
         new_metrics = result.metrics_full
 
-        # Restore the old model and evaluate
-        old_model = self.storage.load_run_model(run_id=cast(str, run.info.run_id))
-        if not isinstance(old_model, ForecastingModel):
-            self._logger.warning(
-                "Loaded old model from run %s is not a ForecastingModel, skipping model selection",
-                cast(str, run.info.run_id),
-            )
+        old_model = self._restore_model(
```
egordm: Nitpick: I think general naming pattern for functions is _try_load_model, _try_evaluate_model. It implies that it catches errors and returns a result, even on failure.

PR #786, opened 2025-12-16
feature: Selector transform

lschilders:
This pull request introduces a new feature selection transform called `Selector` to the `openstef_models` package. The `Selector` transform enables users to select or exclude specific features from a time series dataset based on a flexible selection strategy. Comprehensive unit tests are also added to ensure correct behavior.

**New Feature: Feature Selection Transform**

* Added the `Selector` class in `selector.py`, which allows selecting a subset of features from a `TimeSeriesDataset` using a `FeatureSelection` strategy. This is useful for preprocessing data before model training or inference. The transform ensures that the target column is always included when using a `ForecastInputDataset`.

**API and Module Updates**

* Registered the new `Selector` transform in the `__init__.py` of the `general` transforms submodule, making it available for import and use. [[1]](diffhunk://#diff-ad8cd6918e1ba7200416fb155497644c30fbf46a1a1d773f83197594ec2cb18bR20) [[2]](diffhunk://#diff-ad8cd6918e1ba7200416fb155497644c30fbf46a1a1d773f83197594ec2cb18bR30)

**Testing**

* Added a new unit test file `test_selector.py` to verify that the `Selector` correctly selects or excludes features and handles target columns appropriately for forecast datasets.

### DIFF
--- a/packages/openstef-models/src/openstef_models/transforms/general/selector.py
+++ b/packages/openstef-models/src/openstef_models/transforms/general/selector.py
@@ -21,7 +21,32 @@ from openstef_models.utils.feature_selection import FeatureSelection
 
 
 class Selector(BaseConfig, TimeSeriesTransform):
-    """Selects features based on FeatureSelection."""
+    """Selects features based on FeatureSelection.
+
+    Example:
+        >>> import pandas as pd
+        >>> from datetime import timedelta
+        >>> from openstef_core.datasets import TimeSeriesDataset
+        >>> from openstef_models.transforms.general import Selector
+        >>> from openstef_models.utils.feature_selection import FeatureSelection
+        >>>
+        >>> # Create sample dataset
+        >>> data = pd.DataFrame(
+        ...     {
+        ...         "load": [100.0, 110.0, 120.0],
+        ...         "temperature": [20.0, 22.0, 23.0],
+        ...         "humidity": [60.0, 65.0, 70.0],
+        ...     },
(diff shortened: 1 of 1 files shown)


packages/openstef-models/tests/unit/transforms/general/test_selector.py
```diff
@@ -0,0 +1,84 @@
...
+            {"load", "temperature"},
+            id="exclude_subset",
+        ),
+        pytest.param(
+            FeatureSelection.ALL,
+            {"load", "temperature", "humidity"},
+            id="all_features",
+        ),
+        pytest.param(
+            FeatureSelection.NONE,
```
MvLieshout, line 39: Don't know if I find dropping the load column intuitive here, since it is added in the other cases. Maybe we should raise a warning in this case that also the load is dropped?

packages/openstef-models/tests/unit/transforms/general/test_selector.py
```diff
@@ -0,0 +1,84 @@
...
+    # Act
+    transformed = selector.fit_transform(dataset)
+
+    # Assert
+    assert set(transformed.feature_names) == expected_features
+    for feature in expected_features:
+        pd.testing.assert_series_equal(transformed.data[feature], dataset.data[feature])
+
+
+def test_selector__fit_adds_target_to_include() -> None:
```
MvLieshout: I think this test is just another case that can be added to the parametrized test.

MvLieshout (commented):
Two comments about the tests. Logic looks good.

Issue #652, opened 2025-08-28
[OpenSTEF 4.0] Implement Median Model


egordm:
> [!IMPORTANT]
> Changes should be done by forking [OpenSTEF 4.0.0 tracking branch](https://github.com/OpenSTEF/openstef/tree/release/v4.0.0) (`release/v4.0.0`) and then make a PR back onto `release/v4.0.0` branch.
> Please read the [contributing guide](https://openstef.github.io/openstef/v4/contribute/index.html) before you start.

Definition of Done:

✅ Acceptance criteria

- [ ] Implemented using existing model interface
- [ ] Created tests
- [ ] Created doc page and example script

📄 Documentation criteria:

Interfaces with other products and services

Update relevant confluence pages

🧪 Test criteria:

Unit tests written and passed

⌛ Dependencies:

#627 

🌍 Background

See implementation from openstef 3.0

https://github.com/OpenSTEF/openstef/blob/main/openstef/model/regressors/median.py

Note: Not all the tests for the current implementation really make sense. For example, in [the test to see how a large gap is handled](https://github.com/OpenSTEF/openstef/blob/027d9a4a125673d25c64b22e2225fe45b8b143e2/test/unit/model/regressors/test_median.py#L146), the lagged values should actually not be known for all time stamps. The bottom left corner should contain nan's. The lag values are the measured values. So since we do not know the measured value yet of the 2023-01-01 00:01:00 timestep yet, we do not know the T-1min lag for the 2023-01-01 00:02:00 timestep. 

<img width="360" height="90" alt="Image" src="https://github.com/user-attachments/assets/7316cb61-8c9e-491f-af4b-77c9e323633b" />

The data frame definition should then be like:
<img width="542" height="149" alt="Image" src="https://github.com/user-attachments/assets/5f89f004-266c-4208-b968-f6c5aa2f2fbc" />

PR #793, opened 2025-12-19
Implement-median-model-v4

JanMaartenvanDoorn:
### Changes proposed in this PR include:
- Added Median model to openSTEF V4.0
- Checked behavior against openSTEF v3.x and matches
- Implements #652

### DIFF
--- a/packages/openstef-models/src/openstef_models/transforms/time_domain/rolling_aggregates_adder.py
+++ b/packages/openstef-models/src/openstef_models/transforms/time_domain/rolling_aggregates_adder.py
@@ -79,6 +79,11 @@ class RollingAggregatesAdder(BaseConfig, TimeSeriesTransform):
     _last_valid_aggregates: dict[str, float] = PrivateAttr(default_factory=dict[str, float])
     _is_fitted: bool = PrivateAttr(default=False)
 
+    @property
+    @override
+    def is_fitted(self) -> bool:
+        return self._is_fitted
+
     def _make_column_name(self, func: AggregationFunction) -> str:
         suffix = timedelta_to_isoformat(td=self.rolling_window_size)
         return f"rolling_{func}_{self.feature}_{suffix}"


packages/openstef-models/tests/unit/models/forecasting/test_median_forecaster.py
```diff
@@ -0,0 +1,463 @@
...
+def test_median_returns_median():
+    # Arrange
+    index = pd.date_range("2020-01-01T00:00", periods=3, freq="h")
+    training_data = create_timeseries_dataset(
+        index=index,
+        load=[1.0, 4.0, 7.0],
+        load_lag_PT1H=[1.0, np.nan, np.nan],
+        load_lag_PT2H=[4.0, 1.0, np.nan],
+        load_lag_PT3H=[7.0, 4.0, 1.0],
+        available_at=index,
```
egordm: Usually forecast input dataset no longer has an available at. So this could be removed for simplicity. If input dataset as an extra timestamp then it's horizon but in this case only one is supported.

packages/openstef-models/src/openstef_models/models/forecasting/median_forecaster.py
```diff
@@ -0,0 +1,374 @@
...
+        """Check if the frequency of the input data matches the model frequency.
+
+        Args:
+            index (pd.DatetimeIndex): The input data to check.
+
+        Returns:
+            bool: True if the frequencies match, False otherwise.
+        """
+        input_frequency = self._infer_frequency(index) if index.freq is None else index.freq
+        return input_frequency == self.frequency
```
egordm: Maybe it would be nice to move this to TimeSeriesDataset. To have something like one function called `validate_sample_interval` that checks the data against the set sample interval. If user wants to be sure they can call it. 

It would make it easier to test and median model code would be a lot simpler.

packages/openstef-models/tests/unit/models/forecasting/test_median_forecaster.py
JanMaartenvanDoorn: Great, removed this.

packages/openstef-models/src/openstef_models/models/forecasting/median_forecaster.py
JanMaartenvanDoorn: Problem here is that self.frequency is part of the model and not of the input data. So we are having two frequencies here: 1. the frequency of the input data and 2. the frequency of the model itself which corresponds to the frequence of the lags. The method here required both frequencies to be the same and as this is specific for the median model I would propose leaving this check in the median model.

packages/openstef-core/src/openstef_core/datasets/timeseries_dataset.py
```diff
@@ -122,10 +122,27 @@ def __init__(
...
+            ValueError: If data frequency does not match sample_interval.
         """
         if not isinstance(data.index, pd.DatetimeIndex):
             raise TypeError("Data index must be a pandas DatetimeIndex.")
 
+        if sample_interval is None:
+            inferred_freq = pd.Timedelta(
+                self._infer_frequency(data.index) if data.index.freq is None else data.index.freq  # type: ignore
+            )
+            sample_interval = inferred_freq.to_pytimedelta()
```
egordm: Oh maybe we shouldn't infer frequency from the input data? Forcing user to specify it would be the best. The data may contain holes or similar throwing off the inference.

We should probably only have a validate function that optionally validates if the data uses the right sample interval for functions that are sensitive to this. Like the median model.

It probably also solves issues you have been having with failing doctests.

packages/openstef-models/src/openstef_models/models/forecasting/median_forecaster.py
```diff
@@ -0,0 +1,325 @@
...
+
+        # Check if lags are evenly spaced
+        lag_deltas = sorted(self.lags_to_time_deltas_.values())
+        lag_intervals = [(lag_deltas[i] - lag_deltas[i - 1]).total_seconds() for i in range(1, len(lag_deltas))]
+        if not all(interval == lag_intervals[0] for interval in lag_intervals):
+            msg = (
+                "Lag features are not evenly spaced. "
+                "Please ensure lag features are evenly spaced and match the data frequency."
+            )
+            raise ValueError(msg)
```
egordm, line 326: Usually I like splitting chunks like this off into generic standalone functions cause they are easy to test in isolation but harder to test in model.

For example here we could have a function check_timedeltas_evenly_spaced which is easy to test and doesn't require extra test for the model.

This is more of a personal preference. If it's not worth it we can skip it here.

packages/openstef-models/src/openstef_models/models/forecasting/median_forecaster.py
```diff
@@ -0,0 +1,325 @@
...
+
+        # Check that lag frequency matches data frequency
+        expected_lag_interval = lag_intervals[0]
+        if expected_lag_interval != self.frequency_.total_seconds():
```
egordm: Most of the operations can be done on timedelta structure the total seconds would be unnecessary. In this case it would be a simple comparison.

egordm (changes requested):
Looks great. The only change I would suggest is to avoid actually inferring timedelta, it's something user can better give manually given how many edge cases there are for this. We can only do verification. This way we also don't need any changes in stef beam and such.

packages/openstef-models/src/openstef_models/models/forecasting/median_forecaster.py
JanMaartenvanDoorn: good catch, changed this

packages/openstef-core/src/openstef_core/datasets/timeseries_dataset.py
JanMaartenvanDoorn: now changed such that inference is only used for checking th einput data. And this check is also optional.

packages/openstef-models/src/openstef_models/models/forecasting/median_forecaster.py
JanMaartenvanDoorn: These checks are strongly coupled to this model. Lets keep them here for now, if we need similar check somewhere else we can split them off.

egordm: Alright. Lets keep them here.

egordm (approved):
Great work. Looks good to me and is ready to merge.

PR #795, opened 2025-12-19
isotonic calibration

fleur-petit:
## Summary

This PR implements a postprocessing calibration method for quantile estimation using sk-learn's IsotonicRegression.

## Changes

- Added `IsotonicCalibrator` to postprocessing transforms.
- Added tests to check `IsotonicCalibrator` functionality on different data with different settings.
- Added example with visualisation to show the effect of `IsotonicCalibrator` on observed vs expected quantiles

## Testing

- [x] Existing tests pass
- [x] New tests test new calibration method in different scenario's 
- [x] Example script to visually inspect effect of isotonic calibration on observed vs expected quantiles.

### DIFF
--- a/packages/openstef-models/src/openstef_models/integrations/mlflow/mlflow_storage_callback.py
+++ b/packages/openstef-models/src/openstef_models/integrations/mlflow/mlflow_storage_callback.py
@@ -75,13 +75,14 @@ class MLFlowStorageCallback(BaseConfig, ForecastingCallback):
         if not self.model_reuse_enable:
             return
 
-        # Find the latest successful run for this model
+        # If run_name is provided, load that specific run
         if context.workflow.run_name is not None:
             run = self.storage.search_run(
                 model_id=context.workflow.model_id,
                 run_name=context.workflow.run_name,
             )
         else:
+            # Find the latest successful run for this model
             runs = self.storage.search_latest_runs(model_id=context.workflow.model_id)
             run = next(iter(runs), None)
 
@@ -156,9 +157,17 @@ class MLFlowStorageCallback(BaseConfig, ForecastingCallback):
         if context.workflow.model.is_fitted:
             return
 
-        # Find the latest successful run for this model
-        runs = self.storage.search_latest_runs(model_id=context.workflow.model_id)
-        run = next(iter(runs), None)
+        # If run_name is provided, load that specific run
+        if context.workflow.run_name is not None:
+            run = self.storage.search_run(
+                model_id=context.workflow.model_id,
+                run_name=context.workflow.run_name,
+            )
+        else:
+            # Find the latest successful run for this model
+            runs = self.storage.search_latest_runs(model_id=context.workflow.model_id)
+            run = next(iter(runs), None)
+
         if run is None:
             raise ModelNotFoundError(model_id=context.workflow.model_id)
 
--- a/packages/openstef-models/src/openstef_models/models/forecasting_model.py
+++ b/packages/openstef-models/src/openstef_models/models/forecasting_model.py
@@ -26,7 +26,7 @@ from openstef_core.datasets import (
     TimeSeriesDataset,
 )
 from openstef_core.datasets.timeseries_dataset import validate_horizons_present
-from openstef_core.exceptions import NotFittedError
+from openstef_core.exceptions import InsufficientlyCompleteError, NotFittedError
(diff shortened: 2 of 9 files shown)


packages/openstef-models/src/openstef_models/transforms/postprocessing/isotonic_quantile_calibrator.py
```diff
@@ -0,0 +1,199 @@
...
+    def is_fitted(self) -> bool:
+        return self._is_fitted
+
+    @override
+    def fit(self, data: ForecastDataset) -> None:
+        if data.target_series is None:
+            raise ValueError("Input data must contain target series for calibration.")
+
+        # Determine which quantiles to calibrate
+        quantiles_to_fit = self.quantiles if self.quantiles is not None else data.quantiles
```
Lars800, line 122: ~The quantiles to be calibrated are typically known at init time.  To simplify the fit method slightly, you could specify the _calibrators during init like so:~
```
@override
def model_post_init(self):
    self._calibrators = [IsotonicRegression() for _ in self.quantiles]
```

~This removes the need for creating IsotonicRegressors and checking the quantiles_to_fit in the fit method.~ 

Update: I just realized the constructor requires y_min and y_max, hence moving initialization outside of the fit function does not make sense.

packages/openstef-models/src/openstef_models/transforms/postprocessing/isotonic_quantile_calibrator.py
```diff
@@ -0,0 +1,199 @@
...
+from openstef_core.datasets import ForecastDataset
+from openstef_core.exceptions import NotFittedError
+from openstef_core.mixins import Transform
+from openstef_core.types import Quantile
+
+
+class IsotonicQuantileCalibrator(BaseModel, Transform[ForecastDataset, ForecastDataset]):
+    """Calibrate quantile predictions using isotonic regression.
+
+    This transform learns a monotonic mapping from predicted quantile values to
```
Lars800, line 31: The HOW is really clear from the instructions, what could be clarified more is an explaination of WHY. 
For example for QuantileSorter: Sort predictions based on quantile, because it does not make sense that P51 < P50. 
As Isotonic regression is a bit niche, explaining why it should be used can help in understanding and adoption

packages/openstef-models/src/openstef_models/transforms/postprocessing/isotonic_quantile_calibrator.py
```diff
@@ -0,0 +1,199 @@
...
+                raise ValueError(missing_column_message)
+
+            # Get predictions and actuals
+            predictions = data.data[column_name].to_numpy()  # type: ignore
+            actuals = data.target_series.to_numpy()  # type: ignore
+
+            # Remove NaN values (required for IsotonicRegression)
+            mask = ~(np.isnan(predictions) | np.isnan(actuals))
+            predictions_clean = predictions[mask]
+            actuals_clean = actuals[mask]
```
Lars800, line 144: Calibriation can be quite tricky in case we have a very small data. To limit this risk we could introduce a warning or condition on the number of datapoints. If the fraction or absolute number of datapoints is below this threshold, we can warn the user or even skip the calibration entirely.

packages/openstef-models/src/openstef_models/transforms/postprocessing/isotonic_quantile_calibrator.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def _estimate_local_quantiles(
+        self, predictions: np.ndarray, actuals: np.ndarray, quantile_level: float
+    ) -> np.ndarray:
+        """Estimate local quantiles using adaptive windowing.
+
+        Returns:
+            Array of locally estimated quantile values matching the input order.
+        """
+        # Sort by predicted values
```
Lars800, line 190: Should we set a lower limit on the length of the window? If we use the `np.quantile` we need at least 100 observations. Otherwise we risk `Q50 == Q51`

packages/openstef-models/src/openstef_models/transforms/postprocessing/isotonic_quantile_calibrator.py
fleur-petit: The .05 is a bit arbitrary, what do you think?

PR #798, opened 2025-12-31
fix: handling no valid data in flatliner check

lschilders:
Makes sure that checks from `presets` are failing on `CompletenessChecker` rather than on `FlatlinerChecker` with cryptic message: `TypeError: unsupported operand type(s) for -: 'NoneType' and 'datetime.timedelta'`.


MvLieshout (approved):
LGTM!

PR #807, opened 2026-02-03
add inverse frequency sample weighting method

MentReeze:
Is there a nicer way to indicate that some parameters are only used for one or the other method? Or is it fine like this (clarified in the parameter description).

### DIFF
--- a/examples/benchmarks/liander_2024_benchmark_xgboost_gblinear.py
+++ b/examples/benchmarks/liander_2024_benchmark_xgboost_gblinear.py
@@ -34,7 +34,7 @@ from openstef_models.presets import (
 
 logging.basicConfig(level=logging.INFO, format="[%(asctime)s][%(levelname)s] %(message)s")
 
-OUTPUT_PATH = Path("./benchmark_results_test_convenience")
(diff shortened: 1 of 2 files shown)


packages/openstef-models/src/openstef_models/transforms/general/sample_weighter.py
```diff
@@ -132,13 +143,23 @@ def transform(self, data: TimeSeriesDataset) -> TimeSeriesDataset:
...
+                exponent=self.weight_exponent,
+                floor=self.weight_floor,
+            )
+        else:  # inverse_frequency
+            weights = inverse_frequency_sample_weight(
+                x=target,
+                n_bins=self.n_bins,
+                dampening_exponent=self.dampening_exponent,
+                floor=self.weight_floor,
+            )
```
lschilders: I'd suggest to change this to a match-case statement:
```
match self.method:
      case "exponential":
           ...
      case "inverse_frequency":
           ...
      case _:
           raise ValueError()      
```
Might be nice to move it to a separate method `_calculate_weights` or something.

PR #818, opened 2026-03-02
refactor(STEF-2702): Meta rework v2 — flatten configs, abstract hparams, clean up presets

egordm:
## Summary

Architectural refactoring of the openstef-meta and openstef-models packages, building on top of the HybridForecaster2.0 work.

### Changes

**Phase 1: Extract base classes**
- Extracted `BaseForecastingModel` and `ContributionsMixin` from `ForecastingModel`
- `EnsembleForecastingModel` now extends `BaseForecastingModel` directly (sibling, not child)

**Phase 2: Flatten forecaster configs ("object IS its own config")**
- Removed all `ForecasterConfig` wrapper classes — forecasters now hold their hyperparams directly
- Added abstract `hparams` property on `Forecaster` base, concrete `hyperparams` field on subclasses
- Eliminated pyright ignores via this pattern

**Phase 3: Flatten combiner configs**
- Same pattern applied to `ForecastCombiner` — abstract `hparams`, concrete `hyperparams` on `WeightsCombiner`/`StackingCombiner`
- Column separator changed to `__`
- LightGBM made optional
- Classification moved into combiner

**Phase 4: Polish**
- Renamed preset to `EnsembleForecastingWorkflowConfig` / `create_ensemble_forecasting_workflow`
- Extracted `_build_forecasters()` and `_build_combiner()` helper functions from monolithic factory
- Prefixed internal helpers with `_`
- Added inline documentation across key source files
- Fixed all lint and type errors to zero

### Validation

| Check | Result |
|---|---|
| Lint (ruff) | 0 errors |
| Type (pyright) | 0 errors |
| Tests | 755 passed |
| Doctests | 104 passed |
| Coverage | 87.74% |


packages/openstef-meta/src/openstef_meta/models/ensemble_forecasting_model.py
```diff
@@ -2,11 +2,11 @@
...
 
-"""High-level forecasting model that orchestrates the complete prediction pipeline.
+"""Ensemble forecasting model combining multiple base forecasters.
 
-Combines feature engineering, forecasting, and postprocessing into a unified interface.
-Handles both single-horizon and multi-horizon forecasters while providing consistent
-data transformation and validation.
+Orchestrates parallel base forecasters whose predictions are aggregated by a
+``ForecastCombiner``.  Extends ``BaseForecastingModel`` as a sibling of
+``ForecastingModel`` (D1).
```
MvLieshout: What does the "D1" stand for here?

packages/openstef-meta/src/openstef_meta/models/ensemble_forecasting_model.py
```diff
@@ -43,80 +44,67 @@ class EnsembleModelFitResult(ModelFitResult):
...
-        ...     config=ConstantMedianForecasterConfig(horizons=[LeadTime.from_string("PT36H")])
+        ...     horizons=[LeadTime.from_string("PT36H")]
         ... )
         >>> forecaster_2 = ConstantMedianForecaster(
-        ...     config=ConstantMedianForecasterConfig(horizons=[LeadTime.from_string("PT36H")])
+        ...     horizons=[LeadTime.from_string("PT36H")]
         ... )
-        >>> combiner_config = WeightsCombiner.Config(
+        >>> combiner = WeightsCombiner(
         ...     horizons=[LeadTime.from_string("PT36H")],
```
MvLieshout, line 82: What happens if the horizons for the combiner is different from the forecasters? Do we check for that anywhere? I cannot find it.

Edit: It does not happen when using `EnsembleForecastingWorkflowConfig` because only a single config.horizons is passed. So low impact I guess...

packages/openstef-models/tests/unit/models/forecasting/test_constant_median_forecaster.py
```diff
@@ -38,22 +37,22 @@ def sample_forecast_input_dataset() -> ForecastInputDataset:
 
 
 @pytest.fixture
-def sample_forecaster_config() -> ConstantMedianForecasterConfig:
+def sample_forecaster_config() -> ConstantMedianForecaster:
```
MvLieshout: I guess the naming for these fixtures can also be changed:
```suggestion
def sample_forecaster() -> ConstantMedianForecaster:
```

MvLieshout (commented):
New setup looks very clean! Some questions, let's discuss this afternoon.

packages/openstef-meta/src/openstef_meta/models/ensemble_forecasting_model.py
egordm: Leftover reference to a section in the design doc. Removed — thanks for catching it.

packages/openstef-meta/src/openstef_meta/models/ensemble_forecasting_model.py
egordm: Good point. Added a `model_validator` on `EnsembleForecastingModel` that rejects mismatched horizons at construction time — all forecasters and the combiner must share the same horizons list. This catches the issue immediately rather than waiting for a confusing runtime error during fit.

packages/openstef-models/tests/unit/models/forecasting/test_constant_median_forecaster.py
egordm: Agreed, `_config` suffix was stale after the flatten refactor. Renamed to `sample_forecaster`.

PR #824, opened 2026-03-05
Fix: Correct timedelta conversion in train-test split.

### LINKED ISSUE #823: Train test split fails when time series data frequency is daily or lower
In `openstef.model_selection.model_selection.split_data_train_validation_test`, the time delta of the data points is converted from a pandas to a Python `timedelta` like this (line 170):
```
delta = timedelta(seconds=delta.seconds)
```
This fails when the value of `delta` is greater equal than one day, yielding incorrect train test splits. A resolution is to use
```
delta = timedelta(seconds=delta.total_seconds())
```
instead.

don-alejandrino:
This implementation also works when the delta in the data timestamps is >= 1 day.

Fixes #823.


don-alejandrino:
Messed up commit signing.

PR #825, opened 2026-03-05
Fix: Correct timedelta conversion in train-test split.

don-alejandrino:
This implementation also works when the delta in the data timestamps is >= 1 day.

Fixes https://github.com/OpenSTEF/openstef/issues/823.


egordm:
Release is created at 3.4.93.

don-alejandrino:
That was fast, thank you!

PR #832, opened 2026-03-10
fix: timezone handling for offset in AvailableAt apply() (matching apply_index())

lschilders:
This pull request updates the semantics and implementation of the `AvailableAt.apply` method to ensure that day offsets are interpreted in the correct timezone, matching the behavior of the vectorized `apply_index` method. The changes also update and expand the test suite to cover edge cases involving timezone boundaries and daylight saving transitions.

Key changes:

**Core logic and semantics:**

* The docstring and implementation of `AvailableAt.apply` in `types.py` have been updated so that the day offset is always interpreted in `self.tzinfo` (or falls back to `date.tzinfo`), ensuring consistency with `apply_index`. Now, the reference date is converted to the source timezone before applying the day offset, so "D-1" always means "the previous calendar day in that timezone". [[1]](diffhunk://#diff-a5fdd3bb0fdb7e7fda4a88fd6ed02992f205ca1d03765b5be5ebf87867b04239L229-R236) [[2]](diffhunk://#diff-a5fdd3bb0fdb7e7fda4a88fd6ed02992f205ca1d03765b5be5ebf87867b04239L240-L243)

**Test suite improvements:**

* Existing tests in `test_types.py` have been updated to reflect the corrected timezone logic, particularly for cases where the reference and target timezones differ. For example, the expected result for a UTC offset with an Amsterdam reference has been changed to match the new semantics.
* New test cases have been added to explicitly verify behavior when crossing day boundaries between UTC and Amsterdam timezones, both for D-1 and D0 offsets, ensuring correctness around midnight and DST transitions.
* Additional test cases have been added to the vectorized `apply_index` tests to verify correct shifting across DST changes and to ensure that element-wise `apply` matches the vectorized results. [[1]](diffhunk://#diff-760276c5b93cc79ab4961be4ec33d2302f37c2e109d3ef990201de188d8b836bR311-R322) [[2]](diffhunk://#diff-760276c5b93cc79ab4961be4ec33d2302f37c2e109d3ef990201de188d8b836bR333-R334)

### DIFF
--- a/packages/openstef-core/src/openstef_core/types.py
+++ b/packages/openstef-core/src/openstef_core/types.py
@@ -226,9 +226,14 @@ class AvailableAt(PydanticStringPrimitive):
     def apply(self, date: datetime) -> datetime:
         """Apply this availability offset to a reference date.
 
-        The time-of-day is interpreted in ``self.tzinfo`` (falls back to
-        ``date.tzinfo``).  The result is returned in the reference date's
-        timezone, or naive when the reference date is naive.
+        The day offset is interpreted in ``self.tzinfo`` (falls back to
+        ``date.tzinfo``), so "D-1" means "the previous calendar day in
+        that timezone".  The time-of-day is also placed in that timezone.
+        The result is returned in the reference date's timezone, or naive
+        when the reference date is naive.
+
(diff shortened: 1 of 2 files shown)


egordm (approved):
LGTM!

PR #834, opened 2026-03-12
fix: Fix length mismatch bug in combiner fit

MvLieshout:
This pull request addresses a regression in the `WeightsCombiner` model by ensuring that label indices are properly aligned with the combined dataset when additional features have fewer rows than the ensemble dataset. It also adds a targeted regression test to verify this behavior.

Regression fix for label alignment:

* Updated the `fit` method in `learned_weights_combiner.py` to reindex `labels` based on the index of the combined dataset, preventing shape mismatches during sample weight computation.

Testing improvements:

* Added a regression test in `test_learned_weights_combiner.py` to confirm that fitting works correctly when `additional_features` has a shorter index than the ensemble dataset, and that no shape mismatch errors are raised.
* Imported `ForecastInputDataset`, `numpy`, and `pandas` in the test file to support the new test case.


egordm (approved):
LGTM!

PR #837, opened 2026-03-13
fix(STEF-2854): handle backtest robustness issues

egordm:
## Summary

Robustness improvements discovered during STEF-2854 model comparison backtest (ensemble benchmarking on STEF50 dataset).

## Fixes

1. **`InsufficientlyCompleteError` raised during `WindowedMetricVisualization.format_html()`** — `metric_line` can be shorter than `reference_line` when the metric time series is incomplete.
   - File: `packages/openstef-beam/src/openstef_beam/evaluation/visualization/windowed_metric_visualization.py`

2. **Too strict data validation in `InsufficientlyCompleteError`** — the error message included the number of unique hours, but the check should only verify that the data is not empty.
   - File: `packages/openstef-core/src/openstef_core/exceptions.py`

3. **Pydantic serialization warning for `Quantile` type** — `Quantile.__get_pydantic_core_schema__` lacked a serializer, causing `UserWarning: Pydantic serializer warnings` when serializing models containing `Quantile` fields.
   - File: `packages/openstef-core/src/openstef_core/types.py`

4. **Broadcast shape mismatch in `LearnedWeightsCombiner.fit()`** — `combine_forecast_input_datasets` uses an inner join that can drop rows, but `labels` was computed before the join, causing a shape mismatch.
   - File: `packages/openstef-meta/src/openstef_meta/models/forecast_combiners/learned_weights_combiner.py`

5. **`IndexError` in `chronological_train_test_split` with near-empty datasets** — when a dataset has fewer than 2 unique timestamps, the split function crashes with `index 1 is out of bounds`.
   - File: `packages/openstef-models/src/openstef_models/utils/data_split.py`

6. **Empty data after inner join in combiner crashes downstream** — when `_prepare_input_data` inner join produces empty data (additional features have different datetime index), the combiner's predict/score path crashes with `ValueError: Input data must be 2 dimensional and non empty` or `ValueError: Input contains NaN`.
   - Fix: `_prepare_input_data` now validates the result isn't empty after the inner join and raises `InsufficientlyCompleteError` — which naturally propagates up and is caught by the backtest harness, retaining the previous model.
   - File: `packages/openstef-meta/src/openstef_meta/models/forecast_combiners/learned_weights_combiner.py`

## Features

7. **`nan_aware_weighted_mean` helper** — extracted reusable NaN-aware weighted mean with weight renormalization. Used by ensemble combiner.
   - File: `packages/openstef-core/src/openstef_core/utils/math.py`
   - Tests: `packages/openstef-core/tests/test_math.py`

8. **`skip_analysis` parameter for `BenchmarkPipeline.run()`** — allows skipping the analysis phase during backtesting when only predictions are needed.
   - File: `packages/openstef-beam/src/openstef_beam/benchmarking/benchmark_pipeline.py`

9. **`filterings` override in `AnalysisConfig`** — allows manual selection of which filterings (LeadTime, AvailableAt) are included in comparison analysis. Defaults to `None` (all subsets included, backward compatible).
   - File: `packages/openstef-beam/src/openstef_beam/analysis/analysis_pipeline.py`

## Testing

- All existing tests pass
- Added `test_quantile_serialization_no_warnings` for fix 3
- Added tests for `nan_aware_weighted_mean` helper

### DIFF
--- a/packages/openstef-beam/src/openstef_beam/analysis/analysis_pipeline.py
+++ b/packages/openstef-beam/src/openstef_beam/analysis/analysis_pipeline.py
@@ -103,10 +103,7 @@ class AnalysisPipeline:
             no providers support the requested aggregation level.
         """
         return [
-            provider.create(
-                reports=reports,
-                aggregation=aggregation,
-            )
+            provider.create(reports=reports, aggregation=aggregation)
             for provider in self.config.visualization_providers
             if aggregation in provider.supported_aggregations
         ]
--- a/packages/openstef-beam/src/openstef_beam/analysis/analysis_pipeline.py
+++ b/packages/openstef-beam/src/openstef_beam/analysis/analysis_pipeline.py
@@ -35,6 +35,11 @@ class AnalysisConfig(BaseConfig):
     visualization_providers: list[VisualizationProvider] = Field(
         default=[], description="List of visualization providers to use for generating analysis outputs"
     )
+    filterings: list[Filtering] | None = Field(
+        default=None,
+        description="When set, only include these filterings (e.g. LeadTime, AvailableAt) in analysis. "
+        "None means use all filterings found in the evaluation data.",
+    )
 
 
 class AnalysisPipeline:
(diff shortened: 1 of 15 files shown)


lschilders (approved):
LGTM! (mind the Sonarcloud error in one of the files)

PR #839, opened 2026-03-18
add shifter transform and tests

MentReeze:


packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
lschilders, line 1: I think a better setup would be to have something like `shift_features` or `align_features` with type `FeatureSelection` in the config with default `FeatureSelection.NONE`, since these can be overwritten more easily (see `clip_features`)

packages/openstef-models/src/openstef_models/transforms/general/shifter.py
```diff
@@ -0,0 +1,135 @@
...
+    @override
+    def transform(self, data: TimeSeriesDataset) -> TimeSeriesDataset:
+        if self._shift == timedelta(0):
+            return data
+
+        features = self.selection.resolve(data.feature_names)
+        transformed_data = data.data.copy()
+
+        original_index = cast(pd.DatetimeIndex, data.data.index)
+        shifted_index = original_index - self._shift
```
lschilders: `TimeSeriesDataset already has index built-in (which is already a `DatetimeIndex`), `cast` in 109 might then also not be needed?
```suggestion
        shifted_index = data.index - self._shift
```

packages/openstef-models/src/openstef_models/transforms/general/shifter.py
```diff
@@ -0,0 +1,135 @@
...
+        nan_mask_shifted = feature_data.isna().set_axis(shifted_index)  # pyright: ignore[reportUnknownMemberType]
+        realigned[nan_mask_shifted.reindex(original_index, method="nearest")] = np.nan
+
+        # Handle timestamps outside the range covered by the shifted data
+        outside_mask = (original_index < shifted_index.min()) | (original_index > shifted_index.max())  # pyright: ignore[reportUnknownMemberType]
+        if self.fill_edges:
+            edge = feature_data.iloc[[-1]] if self._shift > timedelta(0) else feature_data.iloc[[0]]
+            realigned.loc[outside_mask] = edge.to_numpy()
+        else:
+            realigned.loc[outside_mask] = np.nan
```
lschilders: I think something simpler can be done with `ffill` or `bfill` (maybe before restoring nans and then after filling restore the nans). Setting to nan is redudant I think, since it is already nan?

packages/openstef-models/src/openstef_models/transforms/general/shifter.py
lschilders, line 1: I think aggregation_period is a more common term than averaging_period, docs heavily mention averaging

lschilders (changes requested):
Nice implementation, some comments

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
MentReeze: I'm not sure about this. The default parameters of the clipper are quite widely applicable I think, so these can be hardcoded in create_forecasting_workflow as is done now. But for the Shifter, it really depends on which aggregations you are using. This may vary more for different variables and for different openstef users. So allowing to specify these other parameters (except for the FeatureSelection) in the config might be more flexible?

packages/openstef-models/src/openstef_models/transforms/general/shifter.py
MentReeze: I had this before, but I think I found an even cleaner solution now by using the limit_area and limit_direction parameters of DataFrame.interpolate. See the commit. fyi, the default interpolate method fills forward nans automatically with the lastest value. But not backwards. That's why I use limit_direction="both" now.

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
egordm: Currently we assume that there is only one shift is needed. Is that true for most of the cases? Say there are mutliple radiation features which require different shifting or an additional wind variable. Wouldn't it be better to make the setting a list of shifts instead? Otherwise just putting a whole Shift as a setting is okay, but would maybe make it a None by default to avoid confusion?

MentReeze: Good point, requiring multiple shifts would in theory be possible indeed (but we do not need it right now yet). I implemented the list of shifts, is this what you mean?

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
```diff
@@ -311,6 +317,7 @@ def create_forecasting_workflow(
         ),
         CompletenessChecker(completeness_threshold=config.completeness_threshold),
     ]
+    feature_aligners = list(config.shifters)
```
lschilders: This is not needed as config.shifters will always be a list?

lschilders (approved):

PR #853, opened 2026-04-07
feat: Expand clipper functionality with replacement by NaN 

majidkhoshrou:
## Summary
This PR extends the current clipping transform with optional NaN handling for out-of-range values.

## Changes
- Add `outlier_action` parameter:
  - `clip` (default): existing behavior
  - `nan`: replace out-of-range values with NaN
- Rename `Clipper` to `OutlierHandler`
- Update unit tests for both clip and NaN behavior

## Motivation
For some applications, clipping many outliers may distort the input distribution by concentrating values at the learned bounds. In these cases, it is preferable to replace out-of-range values with `NaN` so downstream models can ignore or specially handle them.

## Base
The branch was originally created from `v4.0.0.a30` and rebased onto `release/v4.0.0`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,9 +129,3 @@ benchmark_results*/
 # Mlflow
 /mlflow
 /mlflow_artifacts_local
-
-# ----------------------------------------
-# Local developer files (not part of repo)
-# Personal experiments, debug scripts, plots
-# ----------------------------------------
-local/
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,9 @@ benchmark_results*/
 # Mlflow
 /mlflow
 /mlflow_artifacts_local
+
+# ----------------------------------------
+# Local developer files (not part of repo)
+# Personal experiments, debug scripts, plots
+# ----------------------------------------
+local/
--- a/packages/openstef-models/src/openstef_models/transforms/general/outlier_handler.py
+++ b/packages/openstef-models/src/openstef_models/transforms/general/outlier_handler.py
@@ -42,7 +42,7 @@ class OutlierHandler(BaseConfig, TimeSeriesTransform):
         >>> import pandas as pd
         >>> from datetime import timedelta
         >>> from openstef_core.datasets import TimeSeriesDataset
-        >>> from openstef_models.transforms.general.clipper import OutlierHandler
+        >>> from openstef_models.transforms.general.outlier_handler import OutlierHandler
         >>> from openstef_models.utils.feature_selection import FeatureSelection
         >>>
         >>> training_data = pd.DataFrame({
@@ -164,7 +164,3 @@ class OutlierHandler(BaseConfig, TimeSeriesTransform):
     @override
     def features_added(self) -> list[str]:
         return []
-
-
-# Backward compatibility alias
-Clipper = OutlierHandler
--- a/packages/openstef-models/src/openstef_models/transforms/general/outlier_handler.py
+++ b/packages/openstef-models/src/openstef_models/transforms/general/outlier_handler.py
@@ -11,6 +11,7 @@ learned bounds or replacing them with NaN.
 
 from typing import Literal, override
 
+import numpy as np
 import pandas as pd
 from pydantic import Field, PrivateAttr
 
@@ -153,8 +154,8 @@ class OutlierHandler(BaseConfig, TimeSeriesTransform):
         else:  # outlier_action == "nan"
             feature_data = data.data[features].copy()
 
-            below_lower = feature_data < lower_bound.values
-            above_upper = feature_data > upper_bound.values
+            below_lower = feature_data < np.asarray(lower_bound)
+            above_upper = feature_data > np.asarray(upper_bound)
             out_of_range_mask = below_lower | above_upper
 
             transformed_data[features] = feature_data.where(~out_of_range_mask)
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -129,3 +129,9 @@ benchmark_results*/
 # Mlflow
 /mlflow
 /mlflow_artifacts_local
+
+# ----------------------------------------
+# Local developer files (not part of repo)
```
MvLieshout: Not sure why we need this in the gitignore?

I think we can probably remove it

packages/openstef-models/src/openstef_models/transforms/general/outlier_handler.py
MvLieshout, line 1: I think we should also rename the file to `outlier_handler.py`. Note that all references to this file then also need to be fixed. The corresponding test file will then also need to be renamed.

MvLieshout, line 1: Some quality checks are failing. You can run `uv run poe all` to see locally why they are failing.

packages/openstef-models/src/openstef_models/transforms/general/outlier_handler.py
```diff
@@ -21,56 +21,81 @@
 from openstef_models.utils.feature_selection import FeatureSelection
 
 type ClipMode = Literal["minmax", "standard"]
+type OutlierAction = Literal["clip", "nan"]
 
 
-class Clipper(BaseConfig, TimeSeriesTransform):
-    """Transform that clips specified features to their observed min and max values.
+class OutlierHandler(BaseConfig, TimeSeriesTransform):
```
MvLieshout, line 28: The Clipper itself is also mentioned in some places in the code. We need to rename there as well.

packages/openstef-models/tests/unit/transforms/general/test_outlier_handler.py
```diff
@@ -37,80 +38,123 @@ def test_dataset() -> TimeSeriesDataset:
     )
 
 
+# ---------------------------
+# CLIP BEHAVIOR (DEFAULT)
+# ---------------------------
```
MvLieshout: These comments are not necessary. Create bloated test files.

packages/openstef-models/tests/unit/transforms/general/test_clipper.py
```diff
@@ -4,12 +4,13 @@
...
 from datetime import timedelta
 
+import numpy as np
 import pandas as pd
 import pytest
 
 from openstef_core.datasets import TimeSeriesDataset
 from openstef_core.exceptions import NotFittedError
-from openstef_models.transforms.general import Clipper
+from openstef_models.transforms.general import Clipper  # alias still works
```
MvLieshout: Since we are in alpha, I do not think we need this kind of backwards compatibility. Cleaner to just rename.

packages/openstef-models/src/openstef_models/transforms/general/outlier_handler.py
```diff
@@ -87,35 +112,59 @@ def is_fitted(self) -> bool:
...
+        return lower_bound, upper_bound
+
     @override
     def transform(self, data: TimeSeriesDataset) -> TimeSeriesDataset:
         if not self._is_fitted:
             raise NotFittedError(self.__class__.__name__)
 
         features = self.selection.resolve(data.feature_names)
-        transformed_data = data.data.copy(deep=False)
+        transformed_data = data.data.copy()
```
MvLieshout: Why remove the `deep=False` here in the copy?

packages/openstef-models/tests/unit/transforms/general/test_clipper.py
```diff
@@ -37,80 +38,123 @@ def test_dataset() -> TimeSeriesDataset:
...
-    assert transformed_dataset.sample_interval == test_dataset.sample_interval
 
+    pd.testing.assert_frame_equal(transformed.data, expected_df)
+
+
+# ---------------------------
+# MISSING FEATURES
+# ---------------------------
 
 def test_clipper__handles_missing_features():
```
MvLieshout: You do not need to redefine the dataframes in this test since they are [fixtures](https://docs.pytest.org/en/stable/explanation/fixtures.html). You can use them in you test by adding them as args to the test
```suggestion
def test_clipper__handles_missing_features(train_dataset, test_dataset):
```

You can also check the other test for how to do this.

packages/openstef-models/src/openstef_models/transforms/general/outlier_handler.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+
+    selection: FeatureSelection = Field(default=FeatureSelection.ALL, description="Features to handle.")
+    mode: ClipMode = Field(
+        default="minmax",
+        description="Boundary mode: 'minmax' uses observed min/max, 'standard' uses mean ± n_std * std.",
+    )
+    outlier_action: OutlierAction = Field(
+        default="clip",
+        description="How to handle out-of-range values: 'clip' keeps current behavior, 'nan' replaces them with NaN.",
```
MvLieshout: I do not think we should assume people know what the current behaviour is.
```suggestion
        description="How to handle out-of-range values: 'clip' clips the values to the bounds, 'nan' replaces them with NaN.",
```

MvLieshout (approved):

PR #858, opened 2026-04-15
feat(STEF-2869): add nan_features to workflow configs

MvLieshout:
Add nan_features field to ForecastingWorkflowConfig and EnsembleForecastingWorkflowConfig. When set, a second OutlierHandler with outlier_action='nan' replaces out-of-range values with NaN instead of clipping them to bounds. Default is NONE (no features).

### DIFF
--- a/packages/openstef-meta/src/openstef_meta/presets/forecasting_workflow.py
+++ b/packages/openstef-meta/src/openstef_meta/presets/forecasting_workflow.py
@@ -224,6 +224,10 @@ class EnsembleForecastingWorkflowConfig(BaseConfig):
         default=FeatureSelection(include=None, exclude=None),
         description="Feature selection for which features to clip.",
     )
+    nan_features: FeatureSelection = Field(
+        default_factory=lambda: FeatureSelection.NONE,
+        description="Feature selection for which out-of-range values are replaced with NaN instead of clipped.",
+    )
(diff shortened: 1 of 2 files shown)


packages/openstef-meta/src/openstef_meta/presets/forecasting_workflow.py
```diff
@@ -221,8 +221,13 @@ class EnsembleForecastingWorkflowConfig(BaseConfig):
         description="If not None, rolling aggregate(s) of load will be used as features in the model.",
     )
     clip_features: FeatureSelection = Field(
-        default=FeatureSelection(include=None, exclude=None),
-        description="Feature selection for which features to clip.",
+        default=FeatureSelection.ALL,
+        description="Feature selection for which features to clip to their learned range.",
+    )
+    nan_features: FeatureSelection = Field(
```
lschilders: `nan_features` is a bit of a confusing name imo. Maybe something similar to `drop_out_of_range_features`?

MvLieshout: Good point, will change the name. I feel like `drop_out_of_range_features` is also not really fitting though. Since we do not always drop these features right? For example in the future we could impute these values. What do you think about `nan_on_outlier_features`?

lschilders (approved):

PR #859, opened 2026-04-21
feat: enable flatliner fallback

MvLieshout:
## Changes

- Add `predict_nonzero_flatliner` field to `EnsembleForecastingWorkflowConfig` (was missing, already present in single-model config)
- Enable `error_on_flatliner=True` in both single-model and ensemble presets so `FlatlinerDetectedError` is actually raised and the flatliner fallback path is reachable

## Context

Previously `error_on_flatliner=False` meant the `FlatlinerDetectedError` was never raised, making the flatliner fallback dead code. This activates it for all models.


JanMaartenvanDoorn (dismissed):
Looks good!

MvLieshout:
Thanks for the review @JanMaartenvanDoorn! I have added some tests as well now and made a small change to the openstef-models preset to make it in line with the openstef-meta variant. Could you take a look again?

JanMaartenvanDoorn (dismissed):
Yes, approved.

bartpleiter (approved):

PR #860, opened 2026-04-28
feat(STEF-3001): added lag load fallback.

bartpleiter:


packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
```diff
@@ -70,6 +70,18 @@ class LagsAdder(BaseConfig, TimeSeriesTransform):
...
+        description="Enable single-step fallback for missing lag values. "
+        "When True, NaN values in a lag column are filled with values from "
+        "an older lag shifted by lag_fallback_offset (default: 7 days). "
+        "Useful for robustness against measurement delays.",
+    )
+    lag_fallback_offset: timedelta = Field(
+        default=timedelta(days=7),
+        description="Offset added to a lag to source the fallback value. "
+        "For example, with offset=7d, a NaN in load_lag_P2D is filled from load at T-9D.",
+    )
```
lschilders: I think the naming we use for enabling features in models is with `_enabled`, so this should be `lag_fallback_enabled`. Or we could make this a single nullable setting `lag_fallback_offset: timedelta | None = None`, where None means disabled. What do you think?

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
lschilders, line 1: Don't we need these settings in openstef-meta presets as well?

packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
```diff
@@ -172,8 +188,47 @@ def transform(self, data: TimeSeriesDataset) -> TimeSeriesDataset:
...
+
+            feature_name = self._lag_feature(lag)
+            fallback_values = target.shift(freq=fallback_lag)
+
+            if mask is not None:
+                nan_cells = mask & df[feature_name].isna()
+                df.loc[nan_cells, feature_name] = fallback_values
+            else:
+                nan_cells = df[feature_name].isna()
+                df.loc[nan_cells, feature_name] = fallback_values
```
lschilders: Can we not do some smart logic here with [`pd.combine_first`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.combine_first.html)?

lschilders (commented):
Nice implementation, some comments. Note the Sonarcloud issues as well ;)

packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
bartpleiter: I do not see _enabled anywhere in the repo, but it does make more sense for a boolean flag. In this case I like having a nullable lag_fallback_offset more as it keeps the "feature" compact to just one config line, so I'll do that one!

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
bartpleiter: Yes, but I did add them already in packages/openstef-meta/src/openstef_meta/presets/forecasting_workflow.py right?

packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
bartpleiter: Nice suggestion! Did not know this existed, and applied it now

packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
lschilders: Ah I was confused with data-layer indeed, but nullable also has my preference :)

packages/openstef-models/src/openstef_models/presets/forecasting_workflow.py
lschilders: Ah yes excuse me...

packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
```diff
@@ -207,6 +202,7 @@ def _apply_fallback(
         For each lag L, fills NaN cells with the target shifted by L + lag_fallback_offset,
         but only if L + lag_fallback_offset <= history_available (window guard).
         """
+        assert self.lag_fallback_offset is not None  # noqa: S101
```
lschilders: `assert`s should not be used in production code... guards should be in place when using this function

packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
lschilders, line 150: Maybe a split on versioned and non-versioned will decrease the complexity enough?

packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
bartpleiter: Oof that one slipped through. nice catch though

packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
bartpleiter: That resolved the issue, and honestly is cleaner anyway

lschilders, line 1: Missed this before, but generally I think we are not doing inplace transformations on DataFrames in OpenSTEF. Returning the DataFrame in the methods is a better pattern I think.

lschilders (approved):
LGTM!

PR #861, opened 2026-04-28
Fix/remove many lags meta lgbm

bartpleiter:


bartpleiter:
Converted to draft until backtest validated that this does not reduce model performance somehow

bartpleiter:
Backtest showed decreased performance, keeping it in

PR #862, opened 2026-04-29
fix(STEF-3036): flatliner feature_importances shape mismatch

MvLieshout:
## Summary

`FlatlinerForecaster.feature_importances` built a DataFrame with a single scalar value but N quantile column labels, raising:

```
ValueError: Shape of passed values is (1, 1), indices imply (1, N)
```

whenever `len(quantiles) > 1`.

This crashes the MLflow callback's feature-importance plotting during flatliner-fallback training, which means the flatliner model is **never persisted**. Downstream the prediction service then falls back to serving the previous (non-flatliner) model, so the flatliner fallback is silently broken for any model with more than one quantile.

## Reproduction (TDD)

Added regression test `test_feature_importances_shape_matches_quantiles` that:
- Constructs a `FlatlinerForecaster` with 5 quantiles
- Asserts `.feature_importances` has columns matching the configured quantiles

The test fails on `release/v4.0.0` and passes after the fix.

## Fix

Build the row as a 2D list of zeros sized to the quantile count.

## Verification

- `uv run poe lint` ✅
- `uv run poe format` ✅
- `uv run poe type` ✅ (0 errors, 0 warnings)
- `uv run --package openstef-models pytest packages/openstef-models/tests/unit/models/forecasting/test_flatliner_forecaster.py` ✅ (4 passed)

## Ticket
STEF-3036

### DIFF
--- a/packages/openstef-models/src/openstef_models/models/forecasting/flatliner_forecaster.py
+++ b/packages/openstef-models/src/openstef_models/models/forecasting/flatliner_forecaster.py
@@ -95,10 +95,11 @@ class FlatlinerForecaster(Forecaster, ExplainableForecaster, ContributionsMixin)
     @property
     @override
     def feature_importances(self) -> pd.DataFrame:
+        columns = [quantile.format() for quantile in self.quantiles]
         return pd.DataFrame(
-            data=[0.0],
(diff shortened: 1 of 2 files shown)


lschilders (approved):
LGTM!

PR #869, opened 2026-05-06
fix(STEF-3054): preserve preprocessing NaN in restore_target

MvLieshout:
## Problem

`restore_target` unconditionally restored all target values from the original dataset, undoing any NaN introduced by the `OutlierHandler`. This meant that even when `nan_on_outlier_features` included the target column, outlier rows were never excluded from training.

## Solution: Sentinel Columns

The `OutlierHandler` now adds boolean sentinel columns (`__outlier_nan_<feature>__`) when using `outlier_action="nan"`, explicitly marking which values were NaN'd as outliers.

`restore_target` checks for the sentinel column of the target feature:
- If `__outlier_nan_load__` exists and is `True` → preserve the NaN (don't restore original value)
- After restoring, all sentinel columns are removed from the dataset (they are not model features)

This is explicit — no inference about which NaN values to preserve. Only NaN values explicitly flagged by the OutlierHandler are kept.

## Testing

- 4 new unit tests (OutlierHandler sentinel column behavior + restore_target sentinel handling)
- All 369 openstef-models tests pass

### DIFF
--- a/packages/openstef-models/src/openstef_models/models/forecasting_model.py
+++ b/packages/openstef-models/src/openstef_models/models/forecasting_model.py
@@ -18,7 +18,6 @@ from typing import cast, override
 
 import numpy as np
 import pandas as pd
-
 from pydantic import Field, PrivateAttr
 
 from openstef_beam.evaluation import EvaluationConfig, EvaluationPipeline, SubsetMetric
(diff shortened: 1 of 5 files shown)


egordm (approved):
LGTM!

PR #870, opened 2026-05-06
fix(STEF-3054): exclude __-prefixed columns from feature_names

MvLieshout:
## Problem

The OutlierHandler emits sentinel columns (e.g. `__outlier_nan_load_lag_P7D__`) when it NaN's values. These columns were included in `TimeSeriesDataset.feature_names`, causing feature-aware transforms (Scaler) to fit on them during training. At predict time, when no outliers are detected, the sentinel columns are absent → sklearn's feature-name validation crashes:

```
ValueError: The feature names should match those that were passed during fit.
Feature names seen at fit time, yet now missing:
- __outlier_nan_load_lag_P7D__
```

## Fix

`TimeSeriesDataset` now treats any column whose name starts with `__` as an internal column (same mechanism as `horizon`/`available_at` columns). These are:
- Kept in `data` so transforms can pass them through the pipeline
- Excluded from `feature_names` so feature-aware transforms ignore them

This makes the sentinel column approach work correctly: they flow through the pipeline untouched, are consumed by `restore_target` at the end, and never interfere with the Scaler or model.

## Changes

- **openstef-core/timeseries_dataset.py**: Initialize `_internal_columns` with any `__`-prefixed columns; also apply in the non-versioned branch
- **test_outlier_handler.py**: Assert sentinel columns are excluded from `feature_names`

### DIFF
--- a/packages/openstef-core/src/openstef_core/datasets/timeseries_dataset.py
+++ b/packages/openstef-core/src/openstef_core/datasets/timeseries_dataset.py
@@ -49,7 +49,7 @@ class TimeSeriesDataset(TimeSeriesMixin, DatasetMixin):  # noqa: PLR0904 - impor
     Columns whose names start with a double underscore (``__``) are treated as
     internal/system columns: they are kept in ``data`` so transforms can pass
     them along, but are excluded from ``feature_names`` so feature-aware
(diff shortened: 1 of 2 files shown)


egordm (approved):
LGTM!

PR #872, opened 2026-05-06
Keep custom target column when calling `forecaster.predict()`

### LINKED ISSUE #871: [OpenSTEF 4.0] Custom target column within v4 workflow presets
When setting a custom target column other than "load" for a workflow (by setting `target_column` in a `ForecastingWorkflowConfig` and then calling `create_forecasting_workflow` on that config)  and then calling the workflow's `fit()` method, we get the following error during postprocessing:
```
ValueError: Input data must contain target series for error computation.
```
The reason for that seems to be that in all forecasters' `predict()` methods, the target column from the input data is not carried over to the output `ForecastDataset`, so that the target column of the method's output is always the default "load" value, irrespective of what has been set in the original workflow config. This leads to errors in the downstream postprocessing in case the target column is named differently.

don-alejandrino:
Fix: When calling a forecaster's `predict(input_data)` method, the target column from `input_data` was not carried over to the output `ForecastDataset`, so that the target column of the method's output is always the default "load" value. This led to errors in the downstream postprocessing.

Should fix #871.

### DIFF
--- a/packages/openstef-models/src/openstef_models/models/forecasting/base_case_forecaster.py
+++ b/packages/openstef-models/src/openstef_models/models/forecasting/base_case_forecaster.py
@@ -157,6 +157,7 @@ class BaseCaseForecaster(Forecaster, ExplainableForecaster, ContributionsMixin):
                 index=forecast_index,
             ),
             sample_interval=data.sample_interval,
+            target_column=data.target_column,
         )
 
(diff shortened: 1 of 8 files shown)


don-alejandrino:
Closing due to inconsistent commit sign-off, new PR is #873.

PR #878, opened 2026-05-08
Fix standard lag generation when working with multi-horizon data

don-alejandrino:
When adding standard (trivial) lags to a multi-horizon dataset, currently the longest horizon determines the smallest available lag. As an example, assume we have two horizons, 1 hour and 1 day. Using this logic, no lag smaller than 1 day is added to the dataset. This means that for the forecasting model, the data for BOTH horizons looks exactly like the data for the 1-day horizon.

Instead, the shortest horizon should determine the smallest available lag. This means that in the above example, the following behavior should be observed:
 - All lags smaller than 1 hour are never available and therefore not added to the dataset.
 - All lags larger than 1 hour but smaller than 1 day are available for the 1-hour horizon, but not for the 1-day horizon. Therefore, the corresponding lag columns are filled partially, depending on the horizon.
 - All lags larger than 1 day are always available, so the corresponding columns are filled completely (except of course any transient behavior at the beginning).

This PR proposes to basically just exchange `max_horizon` by `min_horizon` as a simple fix.
I also added a test to verify the expected behavior end-to-end.

### DIFF
--- a/packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
+++ b/packages/openstef-models/src/openstef_models/transforms/time_domain/lags_adder.py
@@ -93,9 +93,9 @@ class LagsAdder(BaseConfig, TimeSeriesTransform):
         return self._lags
 
     @property
-    def max_horizon(self) -> timedelta:
-        """Longest forecast horizon to determine minimum lag requirements."""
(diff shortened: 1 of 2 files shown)


egordm (approved):
The fix is correct and makes sense. LGTM! Thanks!

PR #903, opened 2026-05-22
fix issue with clashing features

JanMaartenvanDoorn:
- is_sunday is generated for both DatetimeFeaturesAdder and HolidayFeatureAdder in the case of Swedish localisation
- Reason is that Sweden treats sunday as a holiday (https://github.com/vacanza/holidays/blob/daadab11a2c4c0968daef8c67191de22808fc8dd/holidays/countries/sweden.py#L42 ).
- To prevent conflict I added exclusion of common DatetimeFeaturesAdder names from holiday features to avoid clashes.


packages/openstef-models/src/openstef_models/transforms/time_domain/holiday_features_adder.py
```diff
@@ -25,6 +25,10 @@
 from openstef_core.datasets import TimeSeriesDataset
 from openstef_core.transforms import TimeSeriesTransform
 
+# Holiday names (after sanitization) that collide with features
+# produced by DatetimeFeaturesAdder and must be excluded.
+_RESERVED_DATETIME_FEATURES: frozenset[str] = frozenset({"sunday", "week_day", "weekend_day"})
+
```
egordm: Could you move this constant as a default field on HolidayFeatureAdder? This way if someone needs to reserve more features, then they can configure it without making changes to openstef.

egordm (changes requested):
Very interesting how it's treated as a holiday. Changes look good, just a small adjustment is needed. Thanks!

packages/openstef-models/src/openstef_models/transforms/time_domain/holiday_features_adder.py
JanMaartenvanDoorn: Good point, I changed this.

egordm (approved):
LGTM! Thanks for the changes!

egordm:
Merged it into main (we no longer use v4 branch)

PR #917, opened 2026-05-27
chore(deps): use uv ecosystem and add cooldown for supply-chain hardening

egordm:
## Summary

Follow-up to #913. Two related changes to the Dependabot config:

### 1. Switch `package-ecosystem` from `pip` to `uv`

Dependabot has a dedicated `uv` ecosystem for projects with `uv.lock` files. The previous config used `pip` (the umbrella Python ecosystem), which worked as a fallback but isn't the canonical value.

Per Astral's docs: <https://docs.astral.sh/uv/guides/integration/dependabot/>

### 2. Add `cooldown` blocks for supply-chain hardening

Brand-new releases now wait a few days before Dependabot proposes a bump, giving the community time to surface malicious or broken releases.

This is the main defence against supply-chain attacks like:
- **xz-utils** (March 2024) — malicious release caught within days of publication
- **tj-actions/changed-files** (March 2025) — compromised action detected fast by the community

Cooldown values applied to all three ecosystems (uv published, uv tooling, github-actions):

| Update type | Days delayed |
|---|---|
| patch | 2 |
| minor | 5 |
| major | 7 |
| default (e.g. pre-releases) | 3 |

### Important: security updates bypass cooldown

Per [GitHub docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference): "The `cooldown` option is only available for *version* updates, not *security* updates."

So CVE patches still land within ~24h on the daily cadence. Cooldown only delays routine version bumps.

## Test plan

- [ ] Dependabot accepts the new config (check Insights → Dependency Graph → Dependabot tab; any config errors surface there).
- [ ] First daily run after merge: confirm PRs are still being created (cooldown of 2 days could mean the first bundle is smaller than usual if recent bumps are still within the window).
- [ ] Spot check: a security PR fires immediately when an alert is published, not delayed by cooldown.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,8 +7,12 @@
 #  - Reduce PR volume (was ~1 PR per package per advisory) by grouping
 #    patch+minor bumps into daily bundles.
 #  - Keep security visible and fast: separate `applies-to: security-updates`
-#    groups land within ~24h on the daily cadence rather than waiting
-#    up to a week.
+#    groups land within ~24h on the daily cadence. `cooldown` does NOT
+#    apply to security updates, so CVE fixes are never throttled.
+#  - Supply-chain hardening via `cooldown`: a brand-new release waits a
+#    few days before Dependabot proposes a bump, which gives the
+#    community time to surface malicious or broken releases (e.g.
+#    xz-utils 2024, tj-actions/changed-files 2025).
 #  - Major bumps stay individual — they often need code changes.
 #
 # Block layout:
@@ -26,7 +30,7 @@ updates:
   # ---------------------------------------------------------------------
   # Block 1 — Published packages (runtime + dev tooling at root)
   # ---------------------------------------------------------------------
-  - package-ecosystem: pip
+  - package-ecosystem: uv
     directories:
       - "/"
       - "/packages/openstef-beam"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -111,6 +127,14 @@ updates:
     labels:
       - dependencies
       - github-actions
+    # Actions are SHA-pinned but cooldown still helps catch incidents
+    # like tj-actions/changed-files (March 2025) where a compromised
+    # action was published and detected within days.
```
bartpleiter: I do not think we need to have examples to describe the use case here

.github/dependabot.yml
```diff
@@ -7,8 +7,12 @@
...
 #  - Keep security visible and fast: separate `applies-to: security-updates`
-#    groups land within ~24h on the daily cadence rather than waiting
-#    up to a week.
+#    groups land within ~24h on the daily cadence. `cooldown` does NOT
+#    apply to security updates, so CVE fixes are never throttled.
+#  - Supply-chain hardening via `cooldown`: a brand-new release waits a
+#    few days before Dependabot proposes a bump, which gives the
+#    community time to surface malicious or broken releases (e.g.
+#    xz-utils 2024, tj-actions/changed-files 2025).
 #  - Major bumps stay individual — they often need code changes.
```
bartpleiter: Can we make this less of an agentic coding implementation log and just describe the general goal of this file instead?

bartpleiter (dismissed):
Approved, but please look at the two comments first.

.github/dependabot.yml
egordm: Agreed. Too verbose. Will fix.

bartpleiter (approved):
Much better!

PR #931, opened 2026-05-29
fix: resolved issue where test_openstef4.py tests loading stale MLflo…

bartpleiter:
…w artifacts from the repo's mlflow dir (which is created when running examples or benchmarks).


egordm (approved):
Looks very useful. We can even disable mlflow storage altogether for these tests

lschilders:
Not necessary after https://github.com/OpenSTEF/openstef/pull/975

Issue #938, opened 2026-05-29
chore: adopt kaleido 1.3+ (fix Timestamp serialization in plotter)


egordm:
**As** a maintainer
**I want to** remove the `kaleido<1.3` cap and adopt the current kaleido
**So that** we stay on the maintained image-export backend.

🌍 **Background**

#939 capped `kaleido>=1.0.0,<1.3` in `docs/pyproject.toml` because kaleido 1.3.0 raises `TypeError: Type is not JSON serializable: Timestamp` when exporting plotly figures whose band x-values are a Python list of pandas `Timestamp` objects — the quantile-fill pattern in `ForecastTimeSeriesPlotter`. This broke the docs deployment (static PNG export of executed tutorials).

Minimal repro:
```python
import pandas as pd, plotly.graph_objects as go, plotly.io as pio
idx = pd.date_range("2024-01-01", periods=5, freq="h")
fig = go.Figure(go.Scatter(x=list(idx)+list(idx[::-1]), y=[1,2,3,4,5,5,4,3,2,1], fill="toself"))
pio.to_image(fig, format="png")  # OK on kaleido 1.2.0, TypeError on 1.3.0
```

❗ **Priority**: Low. kaleido 1.2.0 works; this is staying current.

✅ **Acceptance criteria**
- [ ] Either: make `ForecastTimeSeriesPlotter` pass band x-values as a `DatetimeIndex` / pre-converted datetimes instead of a raw list of `Timestamp` (more robust, likely the real fix), **or** confirm kaleido >1.3.x fixed the regression upstream.
- [ ] Remove the `<1.3` cap in `docs/pyproject.toml`.
- [ ] `poe docs` builds cleanly.

📄 **Notes**: The plotter-side fix is preferable — it removes the fragility regardless of kaleido version.

egordm:
Reopening — the underlying bug in kaleido is still present, and adopting v1.x for the docs build is currently not viable. Summary of findings while working on PR #942:

**Bug location (kaleido 1.0.0–1.3.0):** `kaleido/_kaleido_tab/_tab.py::_calc_fig` calls `orjson.dumps(spec, default=_orjson_default, option=OPT_SERIALIZE_NUMPY)`, where `_orjson_default` only handles `tolist()` and raises `TypeError: Type is not JSON serializable: Timestamp` for `pd.Timestamp` instances in figure specs (scatter `x`, `add_vrect` bounds, shape annotations, layout ranges).

[...]

PR #942, opened 2026-06-01
Fix docs build for backtesting plot export

egordm:
Changed pins for kaleido. To be picked up later as part of https://github.com/OpenSTEF/openstef/issues/938 to do a proper upgrade.


MvLieshout (approved):
LGTM!

PR #957, opened 2026-06-11
fix: validate forecast horizons against the dataset, not themselves

Valyrian-Code:
## What

`validate_horizons_present` built its comparison set from the requested `horizons` argument and then checked those same horizons against it:

```python
required_horizons = set(horizons or [])
missing_horizons = [h for h in horizons if h not in required_horizons]  # always empty
```

`missing_horizons` is therefore always empty and the guard can never raise — the dataset's actual horizons (`dataset.horizons`) are never consulted.

## Impact

The check is a no-op everywhere it is used, silently disabling horizon validation in:

- `ForecastingModel.predict` (`forecasting_model.py:483`)
- `EnsembleForecastingModel` (`ensemble_forecasting_model.py:407`)
- `LagsAdder` (`lags_adder.py:151`)

A dataset that is missing a requested forecast horizon passes validation and proceeds instead of raising.

## Fix

Build the comparison set from `dataset.horizons`. The non-versioned single-horizon early return is unchanged.

```python
available_horizons = set(dataset.horizons or [])
missing_horizons = [h for h in horizons if h not in available_horizons]
```

## Tests

Added regression tests in `test_versioned_timeseries_dataset.py`:

- a requested horizon absent from the dataset now raises `TimeSeriesValidationError` (this test fails on the previous code and passes with the fix);
- requesting only horizons that exist does not raise;
- a non-versioned dataset still satisfies a single-horizon request.

`pyright`, `ruff check` and `ruff format --check` are clean on the changed files.

### DIFF
--- a/packages/openstef-core/src/openstef_core/datasets/timeseries_dataset.py
+++ b/packages/openstef-core/src/openstef_core/datasets/timeseries_dataset.py
@@ -512,8 +512,8 @@ def validate_horizons_present(dataset: TimeSeriesDataset, horizons: list[LeadTim
     if dataset.horizons is None and len(horizons) == 1:
         return  # Non-versioned dataset can satisfy single-horizon requests
 
-    required_horizons = set(horizons or [])
-    missing_horizons = [h for h in horizons if h not in required_horizons]
+    available_horizons = set(dataset.horizons or [])
+    missing_horizons = [h for h in horizons if h not in available_horizons]
     if missing_horizons:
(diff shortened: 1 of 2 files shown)


egordm (approved):
Nice find and fix. Approved! Thanks for your contribution.

Valyrian-Code:
> Nice find and fix. Approved! Thanks for your contribution.

Thanks for the quick review and merge, @egordm! ✌️

Issue #961, opened 2026-06-12
CI: PR Labeler workflow fails on fork PRs (read-only token; action also on deprecated Node 20)


Valyrian-Code:
### Summary

The **PR Labeler** workflow (`.github/workflows/pr-labeler.yaml`) fails on every PR opened from a fork with:

```
Error: HttpError: Resource not accessible by integration
```

Example: the check on #957 ([run](https://github.com/OpenSTEF/openstef/actions/runs/27376723851/job/80975981651)).

### Root cause

The workflow triggers on `pull_request` and requests `permissions: pull-requests: write`. For PRs **from a fork**, GitHub forces the `GITHUB_TOKEN` to read-only regardless of the requested permissions, so `TimonVS/pr-labeler-action` cannot add labels and errors out.

### Impact

Not merge-blocking (labeling is a convenience, not a required check), but it shows a red ❌ on every external contributor's PR, which reads as "something is broken."

### Proposed fix

Switch the trigger to `pull_request_target`:

```yaml
on:
  pull_request_target:
    types: [opened]
```

This is safe here because the workflow has **no checkout step** — it only reads `.github/pr-labeler.yml` from the base repository and applies labels based on the branch name, so no untrusted PR code is executed. (`pull_request_target` is the trigger recommended in the action's own docs for exactly this reason.)

### Secondary: deprecated Node 20

`TimonVS/pr-labeler-action` is pinned at `v5.0.0` (the latest release, Feb 2024) and runs on Node 20, which GitHub is deprecating (workflows forced to Node 24 from 2026-06-16). Since the action appears unmaintained, migrating to [`actions/labeler`](https://github.com/actions/labeler) may be worth considering longer-term — happy to scope that separately.

### Offer

Happy to open a small PR for the `pull_request_target` fix if that's welcome.

Awaneesh03:
Hi @Valyrian-Code, is this still up for grabs? I'd like to take a crack at the `pull_request_target` fix if no one's working on it.

Valyrian-Code:
Hi @Awaneesh03, go for it! I raised this but I'm focused on another change right now. The fix is just switching the trigger to `pull_request_target`, which is safe here since the workflow has no checkout step. Happy to review your PR.

Awaneesh03:
Thanks @Valyrian-Code! PR #979 is already up — also fixed a typo in the config path comment while I was there. Appreciate the detailed writeup in the issue, made it easy to implement.

[...]

PR #965, opened 2026-06-13
fix: RollingAggregatesAdder.features_added() inconsistent with multi-horizon no-op

Valyrian-Code:
## What

`RollingAggregatesAdder.transform()` returns the input unchanged when more than one horizon is configured (multi-horizon isn't supported yet — it logs a warning and no-ops). But `features_added()` still returned the rolling-aggregate column names in that case, so it advertised columns that `transform()` never adds — a contract mismatch that can mislead any consumer relying on it (e.g. feature selection / column validation).

## Fix

Mirror the `transform()` guard in `features_added()`: return `[]` when `len(self.horizons) > 1`. This keeps the existing graceful no-op behaviour and only makes the reported features consistent with what is actually added. (The empty-`aggregation_functions` case was already consistent.)

## Tests

Added regression tests in `test_rolling_aggregates_adder.py`:
- single horizon → `features_added()` lists the columns;
- multiple horizons → `transform()` leaves columns unchanged and `features_added()` returns `[]` (this test fails on the previous code, passes with the fix).

`ty check`, `ruff check` and `ruff format --check` are clean on the changed files; the full `openstef-models` suite passes (425).

## Note

I kept the existing "graceful no-op + warning" behaviour for multi-horizon rather than raising, since that looked intentional. If you'd prefer `transform()` to raise instead so misconfigurations fail loudly, I'm happy to switch — just let me know.


Valyrian-Code:
Same note as my other PR: the `pr-labeler` failure is just the fork PR token limitation (#961), everything else is green.

One open question from the description: I kept the existing no-op for the multi-horizon case rather than raising, since that looked intentional. Happy to switch to a hard error instead if you think misconfigs should fail loudly.

Valyrian-Code:
Hi! @lschilders  I've opened this PR and would appreciate a review when you have time. Any feedback or suggestions would be very welcome.

egordm (approved):
Verified locally, although rare, the bug is real and good to close the hole.

I think returning empty makes sense cause for this one we have already chosen the warn (not error strategy). So keeping it this way for now.

PR Looks good, approving. Thanks for the contribution.

PR #977, opened 2026-06-19
fix: support non-integer quantiles in ForecastTimeSeriesPlotter (#764)

### LINKED ISSUE #764: [OpenSTEF 4.0]  More extreme quantiles break time series plotting function
In evaluation, use these quantiles [0.001, 0.01, 0.05, 0.10, 0.50, 0.90, 0.95, 0.99, 0.999]. However, this throws an error when plotting the time series because line 247 in forecast_time_series_plotter.py assumes integers.

To fix. Use formatting from openstef-core instead.

BRGOVIND:
Closes #764.

Plotting extreme quantiles blows up. With a set like `[0.001, 0.01, 0.05, 0.10, 0.50, 0.90, 0.95, 0.99, 0.999]` you get columns such as `quantile_P0.1` and `quantile_P99.9`, and the plotter chokes on them:

```
ValueError: invalid literal for int() with base 10: '0.1'
```

The cause is in `_prepare_quantile_bands`: it parses the percentile out of the column name with `int(col.split("P")[1])` and rebuilds it with `f"quantile_P{p:02d}"`, and `BandData` keeps the percentile as an `int`. None of that survives a non-integer percentile.

The issue suggested reusing the formatting from openstef-core, which is exactly the right call — there's already a `Quantile` type that round-trips these correctly. So this PR:

- parses/rebuilds the column names with `Quantile.parse()` / `.format()` and stores percentiles as `float` in `BandData`;
- adds a small `_format_percentile` helper so whole numbers still show as `10%` (no stray `.0`) while non-integers show as `2.5%` / `99.9%`.

Existing integer-percentile labels (`10%-90%`) are unchanged.

Tests added:
- `test_plot_with_non_integer_quantiles` — the regression case, extreme quantiles now plot and label correctly;
- `test_band_stores_non_integer_percentiles` — `2.5` / `97.5` stay floats instead of being truncated;
- `test_format_percentile` — label formatting across whole and fractional percentiles.

Full plotter suite passes locally, and `ruff check` / `ruff format` / `pyright` are clean for the changed file.

_Disclosure: prepared with AI assistance; I've reviewed, run, and verified everything myself._


egordm (approved):
Thanks for the PR/contribution @BRGOVIND! Nice fix, and awesome that you could re-use existing way to format quantiles.

I have added a commit to reformat after recent deps updates.

Approved. Will merge shortly.

BRGOVIND:
> Thanks for the PR/contribution @BRGOVIND! Nice fix, and awesome that you could re-use existing way to format quantiles.
> 
> I have added a commit to reformat after recent deps updates.
> 
> Approved. Will merge shortly.

Thanks for the review and approval! Appreciate the feedback and the formatting update. Looking forward to picking up more issues and contributing further.

PR #979, opened 2026-06-19
fix: use pull_request_target in PR labeler to support fork PRs

### LINKED ISSUE #961: CI: PR Labeler workflow fails on fork PRs (read-only token; action also on deprecated Node 20)

Awaneesh03:
Closes #961

## Summary
The PR Labeler workflow fails on every fork PR with:
`Error: HttpError: Resource not accessible by integration`

## Root cause
The workflow triggers on `pull_request`, which gives fork PRs a read-only 
`GITHUB_TOKEN`. The labeler action needs write access to add labels.

## Fix
Changed trigger from `pull_request` to `pull_request_target`. This is safe 
because the workflow has no checkout step — it only reads the base repo's 
`.github/pr-labeler.yml` config and applies labels based on branch names, 
so no untrusted fork code is executed.

Credit to @Valyrian-Code for the detailed root cause analysis in #961.

### DIFF
--- a/.github/workflows/pr-labeler.yaml
+++ b/.github/workflows/pr-labeler.yaml
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2017-2022 Contributors to the OpenSTEF project <openstef@lfenergy.org>
 #
 # SPDX-License-Identifier: MPL-2.0
-# Automatically label PRs, config in ../pr-labler.yml
+# Automatically label PRs, config in .github/pr-labeler.yml
 name: PR Labeler
 on:
   pull_request_target:
--- a/.github/workflows/pr-labeler.yaml
+++ b/.github/workflows/pr-labeler.yaml
@@ -4,7 +4,7 @@
 # Automatically label PRs, config in ../pr-labler.yml
 name: PR Labeler
 on:
-  pull_request:
+  pull_request_target:
     types: [opened]
 
 jobs:


Valyrian-Code:
Thanks for the quick turnaround @Awaneesh03! The `pull_request_target` switch is the right mechanism for letting the labeler write labels on fork PRs.

One heads-up: on #961 @lschilders mentioned they'd rather retire this unmaintained action and move to an official one like `actions/labeler`, instead of patching it. Might be worth aligning with them before this goes further, since this PR keeps the deprecated action. Happy to help either way.

Awaneesh03:
Thanks for the heads-up! I saw @lschilders' comment on #961. I'll look into migrating to `actions/labeler` instead of continuing to patch the existing action.

PR #980, opened 2026-06-19
fix: DatetimeFeaturesAdder one-hot columns unstable and inconsistent with features_added()

Valyrian-Code:
## What

With `onehot_encode=True`, `DatetimeFeaturesAdder` had two related problems:

1. `transform()` built the month/quarter dummies with `pd.get_dummies(data.index.month, ...)`, which only emits columns for the values present in the data. So the column set was **data-dependent**: training spanning Jan-Mar produced `month_1..3`, while inference in Apr produced `month_4`. Training and inference could therefore yield different feature columns.
2. `features_added()` always returned the non-one-hot names (`month_of_year`, `quarter_of_year`), which the one-hot branch never produces, while omitting the actual `month_N`/`quarter_N` columns.

## Fix

- Build the dummies from a fixed category set (`pd.Categorical(..., categories=1..12 / 1..4)`) so the one-hot columns are always the full, stable set regardless of the data span.
- Branch `features_added()` on `onehot_encode` so it reports exactly the columns `transform()` adds.

## Tests

Added regression tests asserting `set(features_added()) == set(added columns)` for both branches, and that the one-hot columns are the full stable set even when the data spans only two months (the one-hot test fails on the previous code, passes with the fix). The existing one-hot test still passes.

`ty check`, `ruff check`, `ruff format --check`, the module doctest, and the full `openstef-models` suite all pass.

## Note

This makes one-hot output always span all 12 months / 4 quarters. I believe that's the desired behaviour (a stable feature space across train/inference), but flagging it in case you'd prefer to keep it data-dependent.


lschilders (changes requested):
The comments in the method and the docstring in the test are way too verbose. Only add such comments when absolutely necessary. Fix and logic look good.

Valyrian-Code:
Done, trimmed the comment in the one-hot branch and the test docstring. Thanks for the review!

Valyrian-Code:
@lschilders, any suggestion on the changes?

egordm (approved):
Looks great! Good fix. We indeed want the full category values not just the ones available in data.

Thanks for your contribution!

lschilders (approved):

PR #981, opened 2026-06-19
fix: import scikit-learn lazily in openstef-beam r2 metric

### LINKED ISSUE #805: [OpenSTEF 4.0]  Fix optional dependencies stef-beam
As
I want to 
So that I

🌍 Background

Currently if you import anything from stef beam it requires sklearn installed to import metrics. But we might import it dynamically to avoid making this a hard dependency.

> [!IMPORTANT]

❗Priority (What if we don't do this?/Are there any deadlines? etc.)

Definition of Done:

✅ Acceptance criteria

- [ ] 

📄 Documentation criteria:

🧪 Test criteria:

Unit tests written and passed

⌛ Dependencies:

N/A

🚀 Releasing:

N/A

Other information:

Valyrian-Code:
## What

`scikit-learn` is an optional dependency of `openstef-beam` (it is not declared in the package's dependencies), but `metrics_deterministic.py` imported `r2_score` from `sklearn.metrics` at module top. As a result, importing `openstef_beam.metrics` (or anything that pulls it in) failed with `ModuleNotFoundError` when scikit-learn was not installed. Resolves #805.

## Fix

Move the `from sklearn.metrics import r2_score` import inside the `r2()` function, its only use site. Importing the module no longer requires scikit-learn; only calling `r2()` does.

## Tests

- `test_r2_score_is_imported_lazily`: asserts `r2_score` is not bound at module level (locks the lazy import).
- `test_r2_still_computes_with_sklearn_available`: confirms `r2()` still computes correctly.

`ty check`, `ruff check`, `ruff format --check`, the module doctests, and the full beam metrics suite all pass.


egordm (changes requested):
Not sure about this one. R2 is used as default in many openstef components. Making the import lazy would notify users only at runtime that they are missing a dependency.

I don't think making it lazy is the right approach.

It is really tempting to inline R2 though, which would remove openstef-beam's dependency on sklearn.

Valyrian-Code:
Good call, thanks! I've inlined the R2 computation in numpy, so openstef-beam no longer depends on scikit-learn at all rather than just deferring the import to runtime. It matches scikit-learn's `r2_score` including sample weights and the constant-`y_true` convention (1.0 for a perfect fit, else 0.0), and I added value tests to lock that down. Also merged latest main.

Valyrian-Code:
Updated per your suggestion, @egordm: R2 is now inlined in numpy so openstef-beam no longer depends on scikit-learn. I verified it matches `sklearn.r2_score` exactly across 2000 random cases (with and without sample weights) and the constant-`y_true` / negative-R2 cases. I also made it return `NaN` for fewer than two samples, matching scikit-learn's behaviour there. Ready for another look when you have a moment.

egordm (approved):
Looks great now. I added handing of additional edge cases so we don't throw errors on data issues, and throw errors on structure / logic issues.

PR #987, opened 2026-06-22
fix: issues with Sonar quality gate

lschilders:
This pull request introduces several important codebase improvements, that were highlighted in the Sonarcloud quality gate. Summary of the changes below:

* Replaced all usage of the deprecated `pytz` library with the standard library's `zoneinfo` for timezone management in `openstef_core` and associated tests. Updated both code and documentation, including the `AvailableAt` type and related test cases, to use `ZoneInfo` instead of `pytz`. Exception handling and parameterization in tests were also updated accordingly.

* Replaced deprecated `np.where` calls with `np.nonzero` for boolean indexing in array processing, improving clarity and future compatibility. 

Besides this, SonarCloud badge links in the `README.md` to point to the correct summary pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,12 +20,12 @@ SPDX-License-Identifier: MPL-2.0
 ![GitHub Release](https://img.shields.io/github/v/release/openstef/openstef)
 
 
-[![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=OpenSTEF_openstef&metric=alert_status)](https://sonarcloud.io/summary/new_code?id=OpenSTEF_openstef)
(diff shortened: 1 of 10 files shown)


egordm (changes requested):
Hmm... back then we had explicitly chosen not to have extra layers on base forecasters for XGBoost and LightGBM, because the models are different anyway and changes in one can impact the other — plus it adds another layer to the onion.

Also, I'd probably lean more toward composition over inheritance. If it's mainly about parameter duplication, we could just ignore it. Because one model can also have different defaults than another, etc. Unfortunately, Sonar doesn't understand that.

So changes approved for all other files. But lightgbm we probably shouldn't touch or discuss for alternatives first.

egordm (approved):

PR #991, opened 2026-06-23
chore(ci): isolate linter/formatter bumps in their own dependabot group

egordm:
## What

Adds a `python-linters` dependabot group for `ruff`, `ty`, and `pyproject-fmt`, and excludes those from the general `python-versions` group.

## Why

These tools regularly add new rules/checks (ruff, ty) or restyle files (pyproject-fmt restyles every `pyproject.toml` on a major; see the version cap in the root `pyproject.toml`). When bundled into the grouped `python-versions` PR, a single linter bump fails the quality gate and blocks all the unrelated library and security updates in that PR.

Isolating them means a linter bump lands in its own PR (which may need a reformat) and never holds up everything else. They are not security critical.

Mirrors the established `openstef-checkpoints` dependabot config.

## Scope

Config only; no dependency versions change here. Version bumps and the Airflow/security work are a separate PR.


lschilders (approved):
Approved in advance, but one question: should `pytest` also be added to this group?

egordm:
no. pytest doesnt change its lint rules every update. so it's stable

PR #996, opened 2026-06-23
fix(deps): resolve security advisories and move the airflow example to airflow 3

egordm:
## What

Resolves all 33 open Dependabot advisories (1 critical, 4 high, 20 moderate, 8 low). Every affected package is a dev, example, or docs dependency; none are in the shipped runtime packages.

## Changes

- Upgrade the deployment example to apache-airflow 3.2 (`>=2.10,<3` to `>=3.2,<4`). Airflow 2.11 was capping the whole workspace lock, so this also lifts shared deps (sqlalchemy 2.0, starlette 1.3.1, flask 3.1.3, python-multipart 0.0.32, attrs 26, rich 15) and drops the Flask-AppBuilder UI stack that airflow 3 replaced with FastAPI.
- Migrate the example to airflow 3: the DAGs use `airflow.sdk` imports, the example runs under `SimpleAuthManager` (so it does not need the Flask-AppBuilder provider), and the obsolete airflow 2-to-3 transition flag and gunicorn/setproctitle workaround are removed. All three commands (train, forecast, standalone UI) run end to end locally.
- Refresh the transitive dev and docs tooling: jupyter-server 2.18 to 2.20 (the critical), jupyterlab, bleach, tornado, cryptography, aiohttp.
- Skip our own workspace packages in licensecheck: airflow 3 requires licensecheck 2026, whose resolver cannot parse the editable workspace entries.

## Why Dependabot did not fix these

- The airflow advisories need a major bump outside the `<3` cap, and `versioning-strategy: lockfile-only` never widens a pyproject constraint. The new `<4` cap lets 3.x security minors flow automatically.
- The rest are transitive dependencies, and Dependabot security updates do not open PRs for undeclared transitive Python deps.

## Validation

- Full test suite green (980 passed, 89% coverage).
- The airflow deployment example runs end to end: training and forecasting via the CLI, and the standalone UI with both DAGs serialized and visible.


lschilders (approved):
LGTM!

PR #963, opened 2026-06-12
fix(#881): make forecasted timestamps consistent over forecasters

lschilders:
This pull request updates the handling of input data for forecasting models to ensure that predictions are only made for the specified forecast horizon (fixes #881). It also adjusts the test configurations to use longer and more realistic forecast horizons. The main changes are as follows:

**Core logic improvements:**

* The `input_data` method in `ValidatedDatasets` now accepts an optional `horizon` parameter, which limits the returned data to timestamps up to `forecast_start + horizon`. [[1]](diffhunk://#diff-d6a52b438eacfa60132a47e31a0b1e96cec8a15dddf32d0dc3bf058eb63a6478L138-R145) [[2]](diffhunk://#diff-d6a52b438eacfa60132a47e31a0b1e96cec8a15dddf32d0dc3bf058eb63a6478R157-R159)
* All forecasting model classes (`GBLinearForecaster`, `LGBMForecaster`, `LGBMLinearForecaster`, `MedianForecaster`, `XGBoostForecaster`) now pass the `horizon` argument to `input_data`, ensuring predictions are restricted to the intended forecast window. [[1]](diffhunk://#diff-1725b6c0037fdfb4a0cffae2fcdfa37f98516045a969b09bf8f29b909f2fd9a5L278-R278) [[2]](diffhunk://#diff-237e7fc65148daa0569f27d1736adc21a52eee89dda40268392aa44a934f2a95L285-R285) [[3]](diffhunk://#diff-106fc52e42e012a02e624721227d247ee2438c8480ef6fb83ae7856dc067ece3L286-R286) [[4]](diffhunk://#diff-2d013a85bd35ce85a0bd39750e4a4dccd369bfd625bf6029322dc8ccd3ba0a1cL188-R188) [[5]](diffhunk://#diff-a933e569b6f94894ab18b8d097fdf9804d98dbfb372019d35ee3781624c165a3L352-R352)

**Test adjustments:**

[...]

PR #1001, opened 2026-06-24
fix: revert PR #963

lschilders:
Decided to revert #963, since it breaks when using test mode (`.score()` or `openstef-beam`). 

Kept (adapted) `test_validated_datasets.py` since it lacked before that PR.


egordm (approved):
LGTM!

PR #1013, opened 2026-06-29
chore: Warnings strict tests

MvLieshout:
## What
Promote Python warnings to errors when running tests on developer machines, while keeping CI lenient, and fix the underlying warnings at source so almost no exclusions remain.

## Why
Linting (`ruff` `ALL`) and typing (`ty` `all = "error"`) already fail on findings, but pytest previously printed warnings without failing. This closes that gap locally without risking flaky CI.

## Changes
- pytest warnings → errors by default (`filterwarnings = ["error", …]` in `pyproject.toml`); each remaining `ignore` has a justification.
- CI stays lenient: `tests`/`doctests` gain a `--relaxed` flag (`-W default`); the quality workflow passes `--relaxed`.
- Fixed at source (exclusions dropped):
  - pydantic `batch_size` shadow — mixin field made annotation-only.
  - R² `<2 samples` — enlarged tiny fixtures / disabled splitting in mlflow + ensemble tests.
  - numpy All-NaN — asserted via `pytest.warns` in the median all-missing test.
- Kept (external, documented): macOS fork deprecation, xgboost `gblinear` deprecation, ctypes unraisable; MAPE divide stays a deliberate `pytest.warns`.
- Sorted `filterwarnings` so `pyproject-fmt --check` passes.

## Verification
`poe tests` (strict): 1084 passed, 7 skipped. `poe format-pyproject --check` clean.


packages/openstef-models/tests/unit/models/forecasting/test_median_forecaster.py
```diff
@@ -169,7 +169,8 @@ def test_median_handles_all_missing_data():
 
     # Act
     model.fit(training_input_data)
-    result = model.predict(training_input_data)
+    with pytest.warns(RuntimeWarning, match="All-NaN slice encountered"):
+        result = model.predict(training_input_data)
```
lschilders: Can we not fix it in a different way? This looks a bit hacky

pyproject.toml
```diff
@@ -181,6 +181,15 @@ ini_options.markers = [
...
 ]
 ini_options.filterwarnings = [
+  # Promote warnings to errors locally so regressions surface early; CI relaxes
+  # this via `poe tests --relaxed` (-W default). Add a scoped `ignore` below only
+  # when an external warning cannot be fixed in our code, with a justification.
+  "error",
+  # XGBoost upstream: `booster=gblinear` deprecation is emitted by the bundled C++
+  # library; we cannot suppress it at source and still need the linear booster.
+  "ignore:.*booster=gblinear.*is deprecated.*:UserWarning",
+  # ctypes callback unraisable noise from a native dependency during teardown.
```
lschilders: No idea what this means, can you clarify this doctoring?

examples/deployment/pyproject.toml
lschilders, line 1: Maybe unrelated to the goal of this PR, but `pyproject-fmt` unfortunately does not really make this more readable here. Maybe add our own headers to `poe` tasks for the deployment examples to better distinguish them. And maybe a separate `poe_tasks.toml` in this dir?

MvLieshout: Good point, moved to separate `poe_tasks.toml`.

packages/openstef-models/tests/unit/models/forecasting/test_median_forecaster.py
MvLieshout: Agreed, fixed in code instead of test.

pyproject.toml
MvLieshout: Should make more sense now. Caused by ONNX runtime, should only happen in tests though.

lschilders (approved):

PR #1023, opened 2026-07-04
fix: LagsAdder.features_added() over-reports lags outside the horizon window

Valyrian-Code:
## What

`LagsAdder.transform()` only materializes lags that fall within a horizon's valid window (`>=` the horizon and `<=` `history_available`) — a configured lag below the shortest horizon, or beyond the available history, is never added. But `features_added()` returned a name for **every** configured lag (`self._lags`), so it advertised columns that `transform()` never creates.

That is a contract mismatch: any consumer relying on `features_added()` (feature selection, column validation) is told about columns that never materialize. Same defect shape as the recently merged #965 (`RollingAggregatesAdder`) and #980 (`DatetimeFeaturesAdder`).

Concretely, with `history_available=10d`, `horizons=[2D]`, `custom_lags=[1D, 3D, 30D]`, `add_trivial_lags=False`:
- `transform()` adds only `load_lag_P3D` (1D is below the 2D horizon, 30D exceeds the 10D history);
- `features_added()` previously returned `load_lag_P30D`, `load_lag_P3D`, `load_lag_P1D`.

## Fix

Return only the lags `transform()` actually produces — the union of the per-horizon valid lags (`self._horizon_lags`), which mirrors the columns built in `_transform_single_horizon` / `_transform_versioned`:

```python
materialized_lags = sorted({lag for lags in self._horizon_lags.values() for lag in lags}, reverse=True)
return [self._lag_feature(lag) for lag in materialized_lags]
```

No change to `transform()`, the `lags`/`horizon_lags` properties, or any public field.

## Tests

Added regression tests (both fail on the previous code, pass with the fix):
- single-horizon: `features_added()` equals the columns `transform()` adds, with the out-of-window `custom_lags` dropped;
- multi-horizon: `features_added()` equals the union of columns created across horizons.

`ruff check`, `ruff format --check`, and `ty check` are clean on the changed files. The `openstef-models` time-domain transforms suite (64) and the `openstef-meta` presets suite (29 — the one internal consumer, `forecasting_workflow.py`, uses this as an *exclude* set where the change is behaviour-preserving) both pass.


LeandervdBijl:
Dear @Valyrian-Code,

Thank you for your contribution.

We'd like to clarify compliance with our contribution policy (https://github.com/OpenSTEF/.github/blob/main/CONTRIBUTING.md#ai-assisted-contributions).

This project requires contributors to disclose any use of AI-assisted tools during the development of a pull request.

Could you please confirm whether AI tools were involved in producing any part of this change set? This is a routine policy check and not a judgment of the contribution itself.

Valyrian-Code:
Hi @LeandervdBijl, yes, same as I described on #1017: https://github.com/OpenSTEF/openstef/pull/1017#issuecomment-4895403199. Same workflow here. I found the contract mismatch, used AI assistance while drafting the fix and tests, reviewed everything, and verified it locally (fail before, pass after, plus ruff, ty and the full transform suites). I take full responsibility for the change.

bartpleiter (approved):
Nice improvement! Thank you for your contribution.

PR #1027, opened 2026-07-06
ci: add CodeQL code scanning (security-and-quality)

egordm:
## What

Add GitHub CodeQL code scanning with the `security-and-quality` query suite for Python.

## Why

Closes the OpenSSF Silver criterion `static_analysis_common_vulnerabilities`: at least one static-analysis tool should carry rules for common vulnerabilities. CodeQL runs natively in Actions with no secrets, so unlike SonarCloud it also scans pull requests from forks. SonarCloud stays in place and continues to report hotspots.

## How

- New workflow `.github/workflows/codeql.yaml`, triggered on push to `main`, pull requests, and a weekly schedule.
- `build-mode: none` (Python is interpreted; no build step).
- Least-privilege permissions: job-scoped `security-events: write`, otherwise read-only.
- Actions SHA-pinned per repo convention (`actions/checkout` v7.0.0, `github/codeql-action` v3).

## Note for reviewer

If the repository currently uses code-scanning **Default** setup, it should be switched to **Advanced** so this workflow is the scanning source (the two can conflict). Requires repo admin.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,43 @@
+# SPDX-FileCopyrightText: 2026 Contributors to the OpenSTEF project <openstef@lfenergy.org>
+# SPDX-License-Identifier: MPL-2.0
+
+name: CodeQL
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches: ['main']
+  pull_request:
+    types: [opened, synchronize, reopened]
+  schedule:
+    # Weekly full scan so advisories that land after a merge still surface.
+    - cron: '30 4 * * 1'
+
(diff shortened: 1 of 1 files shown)


lschilders (commented):
Good one to make this work for forked PRs. Question, what extra value does Sonarcloud add next to CodeQL? Note/reminder, make sure to add CodeQL as required status check and remove Sonarcloud Analysis from there.

egordm:
From my research CodeQL is more security focused (fixes a real gap). While Sonarcloud tries to be everything (basically ruff + codeql + more).  Unfortunately it's tooling is a bit awkward to incorporate into external PR's as you saw.

Let's evaluate if after this there is any value in sonarcloud or if it's too duplicate on the tooling we already have.

lschilders (approved):

PR #1049, opened 2026-07-27
chore: migrate to pandas 3.0 (Copy-on-Write) (#935)

### LINKED ISSUE #935: chore: migrate to pandas 3.0 (Copy-on-Write)
**As** a maintainer
**I want to** upgrade the pandas constraint from `<3` to allow pandas 3.0
**So that** we stay current and stop pinning a major version behind.

🌍 **Background**

Dependency refresh (#934) adopted every other major (numpy 2.4, scipy 1.17, xgboost 3.2, pyarrow 24, pydantic 2.13, …) but held pandas at `<3`. pandas 3.0 enables Copy-on-Write by default, which surfaced real failures — arrays extracted from DataFrames are now read-only, breaking in-place numpy operations.

Reproduced with `pandas>=2.3.1,<4` locked to 3.0.3, **9 test failures**:
- `openstef-models` median forecaster (×6): `ValueError: underlying array is read-only` from an in-place diagonal fill on a DataFrame-derived array.
- `openstef-beam` timeseries visualization (×2) and `openstef-core` dataset `available_at`/lead-time equivalence (×1).

`pandas-stubs` is pinned `<3` to match and should be unpinned together.

❗ **Priority**: Low/medium. Not blocking — pandas 2.3.x is current and secure. Do before pandas 2.x EOL.

✅ **Acceptance criteria**
- [ ] Bump `pandas>=2.3.1,<3` → `<4` in `packages/openstef-core/pyproject.toml`.
- [ ] Unpin `pandas-stubs` (`<3` → none) in root `pyproject.toml`.
- [ ] Audit `.to_numpy()` / `.values` usages mutated in place; add `.copy()` where CoW makes them read-only (start with the median forecaster's diagonal fill).
- [ ] `poe all` green on pandas 3.0.

🧪 **Test criteria**: the 9 failing tests above pass; full suite green.

⌛ **Dependencies**: merge #934 first.

oaksprout:
**You set the work.** #935: migrate to pandas 3.0, with nine named tests failing under Copy-on-Write.

**It's done.** All nine reproduced and fixed at their root causes, plus two more pandas-3.0 breakages the migration surfaced that weren't in your list.

**Here's the evidence.**

- Fresh clone at `e9ed827e`, patch applied, dependencies installed, then the network was disconnected.
- Full suite in a clean `python:3.12-bookworm` container: **1090 pass, 7 skip, 0 fail** — coverage 89.91% against an 80% floor.
- Beyond the container run: `poe doctests` (109 pass), typecheck, lint, `poe lock` in sync, licence checks, notebooks — all clean.
- All nine originally-failing tests individually re-verified passing.

<details>
<summary><b>Audit trail</b> — an independently checkable record that these checks ran, in this order, before this PR existed</summary>

| | |
|---|---|
| Base commit | `e9ed827e0acaa76340ebefd7cb21d17ec01f1127` |
| Container | `python:3.12-bookworm@sha256:9bed8554…` (linux/arm64), network off during tests |
| The patch, content-addressed | [record](https://gateway.autonolas.tech/ipfs/f01551220d69d3bd0c2f57b7d7703c8f4ee9b0507c37847fce697e1e373ed5aaf947ee882) |
| The verification result, signed by a second key | [record](https://gateway.autonolas.tech/ipfs/f015512202e60478657f20a4c8d78358e9d5e8faf322ae23038782f9f515d473361049b1b) |
| Timestamped sequence | [work assigned](https://sepolia.basescan.org/tx/0x3f32218fe48aabc931cedd63a27b21e4a653e1363477f8601d77eb4d1b3e79d6) → [work delivered](https://sepolia.basescan.org/tx/0x533321e3287e0dab8d2ac6dcbe8fa7f0cfd1aa9db4406acccadb5bc389b0b046) → [checks passed](https://sepolia.basescan.org/tx/0x128b07700842a1ec0f840c53e588f1d6058a953a950c0a9554d01d86459bc316) |

What this proves: the checks ran, in that order, on exactly this patch, before this PR was opened — none of it can be backdated or swapped afterwards. What it doesn't prove: that the fix is *right*. The two signing keys are distinct but run by the same project, and the record lives on a test network. Correctness is your judgement, which is the point.

</details>

Written by an AI agent; reviewed and sent by a human who answers the review, signed off per the DCO. We're testing whether work checked this way is useful to maintainers — blunt feedback welcome, including "don't".

---

## Everything below is written by Claude

Fixes #935.

Dependency changes as prescribed: pandas `>=2.3.1,<3` → `<4` in `packages/openstef-core/pyproject.toml`, `pandas-stubs` unpinned at root, `uv.lock` regenerated (resolves pandas 3.0.5; the lock diff also carries mechanical uv-resolver marker refactors, not hand edits).

The three breakages from the issue, all reproduced first:

- **Median forecaster diagonal fill (×6 failures):** `lag_df.to_numpy()` can return a read-only Copy-on-Write view, and the autoregressive fill mutates it in place → `ValueError: underlying array is read-only`. Fixed with `.to_numpy(copy=True)` in `median_forecaster.py`.
- **Beam visualization (×2):** CoW removed pandas' `df[col]` object-identity cache, so `assert_called_once_with` fell through from an identity shortcut to `Series.__eq__` ambiguity. Fixed with the repo's existing `IsSamePandas` matcher (already used elsewhere in the same file).
- **Core dataset equivalence (×1):** pandas 3.0 infers the coarsest sufficient timedelta resolution per code path (`us` from subtraction vs `s` from `pd.to_timedelta(unit="h")`) — values equal, dtype not. Fixed with `check_dtype=False`, matching the existing `check_names=False` on the same assertion.

Two further pandas-3.0 breakages surfaced only after those were unblocked, both fixed: lowercase `freq="1d"` deprecation in three test fixtures (→ `"1D"`), and an empty `pd.DatetimeIndex([], freq="h")` in `test_median_handles_all_missing_data` resolving to a different default resolution than the pipeline's `date_range`-derived index (fixed by slicing the real index, `index[:0]`).

### Notes for the reviewer

- The `check_dtype=False` and `IsSamePandas` changes are test-side accommodations of pandas 3.0's changed dtype-resolution/identity-caching. Values are genuinely equivalent, but if you'd rather normalise resolution inside `TimeSeriesDataset` itself, happy to rework that way.
- Verification ran against pandas 3.0.5 (what `uv lock --upgrade-package pandas` resolves today) rather than the issue's 3.0.3; the constraint is `<4` so CI floats within 3.x regardless.
- The two extra fixes (`"1d"` → `"1D"`, `index[:0]`) weren't named in the issue — worth a second look that they match your intended shape.

### DIFF
--- a/packages/openstef-beam/tests/unit/analysis/visualizations/test_timeseries_visualization.py
+++ b/packages/openstef-beam/tests/unit/analysis/visualizations/test_timeseries_visualization.py
@@ -14,6 +14,7 @@ from openstef_beam.analysis.visualizations import TimeSeriesVisualization
 from openstef_beam.analysis.visualizations.base import AnalysisAggregation, ReportTuple, RunName
 from openstef_beam.evaluation.models.report import EvaluationSubsetReport
 from openstef_core.testing import IsSamePandas
+from openstef_core.utils import not_none
 from tests.utils.mocks import MockFigure
 
 
@@ -79,7 +80,9 @@ def test_create_by_none_generates_single_target_time_series(
 
     # Assert
     # Verify all expected calls are made with correct parameters
-    mock_add_measurements.assert_called_once_with(sample_evaluation_report.subset.target_series)
+    mock_add_measurements.assert_called_once_with(
+        IsSamePandas(pandas_obj=not_none(sample_evaluation_report.subset.target_series))
+    )
     mock_add_model.assert_called_once_with(
         model_name="TestRun", quantiles=IsSamePandas(pandas_obj=sample_evaluation_report.subset.quantiles_data)
     )
@@ -122,7 +125,9 @@ def test_create_by_run_generates_multi_run_comparison(
 
     # Assert
     # Verify measurements are shared and limits use first target's value
-    mock_add_measurements.assert_called_once_with(sample_evaluation_report.subset.target_series)
+    mock_add_measurements.assert_called_once_with(
+        IsSamePandas(pandas_obj=not_none(sample_evaluation_report.subset.target_series))
+    )
     mock_add_limit.assert_has_calls(
         [
             call(value=100.0, name="Upper Limit"),  # First target's limit, not 150.0
--- a/packages/openstef-core/pyproject.toml
+++ b/packages/openstef-core/pyproject.toml
@@ -27,8 +27,7 @@ classifiers = [
 dependencies = [
   "joblib>=1,<2",
   "numpy>=2.3.2,<3",
-  # Held at <3 pending the pandas 3.0 Copy-on-Write migration (see migration issue).
-  "pandas>=2.3.1,<3",
+  "pandas>=2.3.1,<4",
   "pyarrow>=21",
   "pydantic>=2.12.4,<3",
   "pydantic-extra-types>=2.10.5,<3",
--- a/packages/openstef-core/tests/unit/datasets/test_timeseries_dataset.py
+++ b/packages/openstef-core/tests/unit/datasets/test_timeseries_dataset.py
@@ -249,6 +249,11 @@ def test_available_at_and_lead_time_series_equivalence():
         available_at_dataset.lead_time_series,
         horizon_dataset.lead_time_series,
         check_names=False,
+        # available_at_dataset derives lead_time_series from a datetime64 subtraction while
+        # horizon_dataset stores it directly as the raw (pandas-inferred) timedelta64 column;
+        # pandas independently picks the coarsest sufficient resolution for each ("us" vs
+        # "s"), so the values are equivalent even though the exact dtype differs.
+        check_dtype=False,
     )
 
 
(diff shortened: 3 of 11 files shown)


bartpleiter:
Thank you for your contribution! I reviewed the code and it looks good to me, the comments are also clear. I did a follow-up commit to fix two tiny issues I found with your PR: remove a leftover type ignore comment, and also bump the pandas-stubs to match the pandas version.
To validate we do not have any hidden data bugs that are hard to capture in the unit tests, I ran a subset of the benchmark script on main with pandas 2 and on this PR with pandas 3, and both results are practically indentical. I also validated with the constant quantile forecaster, and checked if the unit tests from this PR still work when synced to pandas 2 (to validate the backwards compatibility). All looks good! Will ask another maintainer for a final review as I now contributed myself too to this PR.

bartpleiter (approved):

egordm (approved):
LGTM!

PR #1062, opened 2026-08-15
fix: use validation data and support N-forecaster ensembles in WeightsCombiner

majidkhoshrou:
## What does this PR do?

Fixes three bugs in `WeightsCombiner` (learned-weights ensemble combiner) and one in
`StackingCombiner` that limit the ensemble machinery to small (2-model), always-trainable
ensembles, even though `EnsembleForecastingModel.forecasters` is already a generic
`dict[str, Forecaster]` intended to support any number of base forecasters:

1. **`WeightsCombiner.fit()` ignored `data_val`.** The method accepted a `data_val`
   parameter but never referenced it, so the per-quantile "who wins" classifier was always
   trained on in-sample training predictions. In-sample predictions from boosted/tree
   forecasters are systematically over-optimistic, biasing the combiner toward whichever
   forecaster overfits the training window hardest. `fit()` now trains on `data_val` when
   supplied, falling back to `data` otherwise.

2. **The degenerate-label guard didn't generalize past 2 forecasters.**
   `_validate_labels()` only fell back to a `DummyClassifier` when exactly one forecaster
   ever won (the only degenerate case possible with 2 base models). With 3+ forecasters
   it's common for a *subset* to never win, which crashed `predict_proba` with a
   `ValueError` (column count mismatch) instead of falling back gracefully. The guard now
   triggers whenever fewer forecasters win than are registered, so ensembles of arbitrary
   size no longer crash.

3. **The `DummyClassifier` fallback silently discarded its own answer.**
   `_predict_weights()` hardcoded weight 1.0 on `self._label_encoder.classes_[0]` — the
   alphabetically-*first* registered forecaster name — instead of asking the fitted
   `DummyClassifier(strategy="most_frequent")` which forecaster actually wins most often.
   This is easy to miss with 2 models (right about half the time by chance) but becomes a
   frequent, silent mis-attribution with 3+ models. The fallback now uses the recorded
   majority winner.

4. **`StackingCombiner.fit()` also discarded `data_val`.** It accepted `data_val` but
   always passed `data_val=None` to the underlying meta-forecaster's `fit()`. It now passes
   validation data through so meta-forecasters that support early stopping (e.g. LGBM,
   XGBoost) can use it.

Together these make the ensemble machinery robust for ensembles of arbitrary size (not
just the built-in 2-model `lgbm`/`gblinear` preset default) and make the learned-weights
combiner actually use held-out validation data as its name/docstring already implied.

No public API changes — all fixes are internal to `fit()`/`_validate_labels()`/
`_predict_weights()` behavior.

Closes #

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore / CI

## Breaking changes checklist

N/A — internal fitting behavior only; no public API, config schema, or serialized object
changes.

## AI disclosure

- [x] AI assistance was used — tool(s): GitHub Copilot (Claude Sonnet 4.5)
  - [x] This is disclosed in a commit message (`Assisted-by: GitHub Copilot (Claude Sonnet 4.5)`)

## Checklist

- [x] `poe all --check` passes locally
- [ ] Documentation updated (docstrings, user guide, examples) if needed
- [x] Commits are signed off per our DCO (`git commit -s`)

### DIFF
--- a/packages/openstef-meta/src/openstef_meta/models/forecast_combiners/learned_weights_combiner.py
+++ b/packages/openstef-meta/src/openstef_meta/models/forecast_combiners/learned_weights_combiner.py
@@ -160,6 +160,12 @@ class WeightsCombiner(ForecastCombiner):
     A classifier predicts per-timestep model weights.  Depending on ``hard_selection``,
     the combiner either picks the best forecaster (hard) or blends using predicted
     probabilities (soft).
+
+    When ``data_val`` is supplied to :meth:`fit`, the classifier is trained on those
+    held-out predictions instead of the in-sample training predictions. In-sample
+    predictions from boosted/tree forecasters are systematically over-optimistic, which
+    would otherwise bias the combiner towards whichever forecaster overfits the training
+    data hardest, regardless of true out-of-sample skill.
     """
 
     hyperparams: HyperParams = Field(
@@ -181,6 +187,7 @@ class WeightsCombiner(ForecastCombiner):
     _is_fitted: bool = PrivateAttr(default=False)
     _feature_names: list[str] = PrivateAttr(default_factory=list[str])
     _models: dict[Quantile, ClassifierMixin] = PrivateAttr(default_factory=dict[Quantile, ClassifierMixin])
+    _dummy_fallback_label: dict[Quantile, str] = PrivateAttr(default_factory=dict[Quantile, str])
 
     @override
     def model_post_init(self, _context: object, /) -> None:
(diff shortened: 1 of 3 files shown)


egordm:
The goals are worthwhile, but this implementation needs some runtime, evaluation, and pickle-compatibility improvements.

- Align validation features with validation predictions (learned_weights_combiner.py:224–238). When validation data is supplied, the combiner switches to it but keeps the training-side additional_features. The normal ensemble pipeline passes features_train here, so the default inner join can produce no rows; fitting then fails. Please pass validation-aligned features and add a regression test through the ensemble workflow.

- Don’t report fit data as held-out validation (learned_weights_combiner.py:224). The combiner now trains on data_val, which the ensemble pipeline then uses to calculate validation metrics. Those metrics are no longer independent. Please use separate calibration or out-of-fold predictions, or otherwise keep the reported validation score genuinely held out.

- Keep stacking validation inputs consistent with training (stacking_combiner.py:97). Training combines additional_features; validation input does not. With an extra predictor feature, I reproduced an XGBoost feature-count error. In the built-in workflow, validation sample weights are also omitted. Pass the matching validation features and test this path.

- Preserve inference for older pickled combiners (learned_weights_combiner.py:190, 294). Older pickles won’t contain _dummy_fallback_label. For a previously saved combiner using DummyClassifier, unpickling succeeds but predict() raises AttributeError. Derive the fallback winner from the persisted classifier, or provide a migration that actually runs for this combiner, with a legacy-pickle test.

The migration hook used by Stateful won’t apply automatically here: WeightsCombiner inherits through ForecastCombiner from Pydantic BaseConfig, not Stateful. I’d merge the underlying fixes after these issues are addressed.

PR #1068, opened 2026-08-19
feat: add regression coverage score metric

### LINKED ISSUE #1064: Add RCS for Probabilistic Forecast Evaluation
## Is your feature request related to a problem?
As a forecast developer, I’m missing key metrics to evaluate quantile calibration:

- **Regression Coverage Score (RCS):** fraction of true outcomes within the prediction interval. Although a calibration error metric already exists and provides similar information, RCS is preferred because it is more intuitive to interpret as a percentage, similar to accuracy.
- **Regression Mean Width Score (RMWS):** average width of the prediction interval. This is already covered by the rIQD (relative Inter Quantile Distance) metric.

These are important KPIs used within Stedin for evaluating probabilistic forecasts.

> [!TIP]
It may also be worthwhile to define a uniform metric for evaluating probabilistic forecasts across the sector, similar to rMAE (relative Mean Absolute Error) for deterministic forecasts.

## Describe the solution you'd like
Add RCS metric to `openstef-beam`.

Rather than introducing MAPIE as a dependency solely for these metrics, implement the metric calculations directly in OpenSTEF.

## Describe alternatives you've considered
Use MAPIE to compute the metrics.

## Affected package(s)

- [ ] openstef-core
- [ ] openstef-models
- [x] openstef-beam
- [ ] openstef-foundation-models
- [ ] openstef-meta
- [ ] examples / deployment
- [ ] docs
- [ ] not sure

## Would this require breaking changes to the public API?

- [x] No
- [ ] Yes
- [ ] Not sure

## Additional context

Coverage and interval width should be evaluated together: high coverage alone can be achieved with overly wide prediction intervals.

MaxPolak97:
## What does this PR do?

Adds the Regression Coverage Score (RCS) metric to `openstef-beam` for evaluating probabilistic forecast interval calibration.

RCS measures the fraction of observed values that fall within a predicted quantile interval, such as P10-P90. The metric is implemented as `rcs(...)` in the deterministic metrics module, exported through `openstef_beam.metrics`, and exposed through `RCSProvider` for evaluation pipelines. The provider follows the same symmetric quantile-pair logic as `RIQDProvider`.

Closes #1064

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Refactor / chore / CI

## Breaking changes checklist

- [ ] If yes: see the [breaking changes guide](https://openstef.github.io/openstef/contribute/development_workflow.html#breaking-changes) for the pickle/`_migrate_state` migration pattern

Migration path for existing users (e.g. "old pickled `XScaler` objects auto-migrate on load", "users must now pass `X` explicitly"):

## AI disclosure

- [x] AI assistance was used — tool(s): 
  - [ ] This is disclosed in a commit message (e.g. `Assisted-by: <tool name>`)

## Checklist

- [x] `poe all --check` passes locally
- [x] Documentation updated (docstrings, user guide, examples) if needed
- [x] Commits are signed off per our [DCO](https://openstef.github.io/openstef/contribute/contributing_guide.html#signing-the-developer-certificate-of-origin-dco) (`git commit -s`)
- [x] PR title follows [Conventional Commits](https://www.conventionalcommits.org/) (e.g. `feat: ...`, `fix: ...`, `feat!: ...` for breaking changes)


packages/openstef-beam/src/openstef_beam/evaluation/metric_providers.py
```diff
@@ -699,6 +700,78 @@ def compute_probabilistic(
...
+        Returns:
+            QuantileMetricsDict containing RCS metrics for each processable quantile.
+        """
+        metrics: QuantileMetricsDict = {}
+
+        for i, quantile in enumerate(quantiles):
+            if self.quantiles is not None and quantile not in self.quantiles:
+                continue
+
+            symmetric_quantile = 1.0 - quantile
```
lschilders: Have a look at the [`Quantile` type](https://github.com/OpenSTEF/openstef/blob/f332eb272904071cb90b7cf34d591812876ed7e9/packages/openstef-core/src/openstef_core/types.py#L300), which has some convenient methods for this kind of things. It can simplify parts of this implementation.

I see you copied a lot from the `RIQDProvider`, maybe we can introduce a helper method that can compute metrics based on symmetric quantiles to be more DRY.

lschilders (changes requested):

packages/openstef-beam/src/openstef_beam/evaluation/metric_providers.py
MaxPolak97: Yes agree, I will add this in `metrics_helpers.py`!

MaxPolak97: @lschilders Done!

egordm (approved):
Code looks great now! Thanks for your contribution.
Merging.

lschilders (approved):
Nice improvement, LGTM!

PR #1089, opened 2026-09-14
fix: use recent median for flatliner fallback

egordm:
## What does this PR do?

Fixes the flatliner fallback using a median calculated over the complete training history.

When a nonzero flatline is detected, the fallback now estimates its value from the configured recent flatline window. The existing `flatliner_threshold` is used for both detection and estimation.

This prevents the fallback from changing over multiple retraining cycles as more flatlined samples enter the training history.

Closes #

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor / chore / CI

## Breaking changes checklist

- [ ] If yes: see the breaking changes guide for the pickle/`_migrate_state` migration pattern

Migration path for existing users:

Not applicable. The new `median_window` option defaults to `None`, preserving direct `FlatlinerForecaster` behavior. Configured workflows use the existing 24-hour `flatliner_threshold`.

## AI disclosure

- [ ] No AI assistance was used
- [x] AI assistance was used — tool(s): GitHub Copilot
  - [ ] This is disclosed in a commit message

## Checklist

- [ ] `poe all --check` passes locally
- [x] Documentation updated through model and field docstrings
- [x] Commits are signed off per the DCO


packages/openstef-models/tests/unit/models/forecasting/test_flatliner_forecaster.py
```diff
@@ -64,3 +65,43 @@ def test_predict_returns_median_when_predict_median_is_true(sample_forecast_inpu
     assert isinstance(result.data, pd.DataFrame)
     assert (result.data == expected_median).all().all()
     assert set(result.data.columns) == {q.format() for q in forecaster.quantiles}
+
+
+def test_predict_returns_recent_median_when_median_window_is_set() -> None:
```
MvLieshout: I would like to also have a test where the value is the median of two different values inside the window. Can maybe be parametrized.

MvLieshout (commented):
Just a small request for an extra test. Code looks good.

packages/openstef-models/tests/unit/models/forecasting/test_flatliner_forecaster.py
egordm: Added an extra test.

MvLieshout (dismissed):
LGTM!

MvLieshout (approved):

PR #1092, opened 2026-09-17
chore(deps): adopt kaleido 1.4 for docs and tutorials

### LINKED ISSUE #938: chore: adopt kaleido 1.3+ (fix Timestamp serialization in plotter)

Blizzeq:
## What does this PR do?

Lifts the kaleido pin in `docs/` and `examples/` from the deprecated 0.2.x line to `>=1.4`, and refreshes `uv.lock`.

Closes #938

The issue's acceptance criteria offered two routes: fix the plotter, or "confirm kaleido >1.3.x fixed the regression upstream". The second one is now met.

@egordm's analysis pinned the bug to `_orjson_default` in `kaleido/_kaleido_tab/_tab.py`, which only handled `.tolist()` and raised on `pd.Timestamp`, and noted that no upstream PR existed yet. One does now: plotly/Kaleido#461 by @gaoflow adds an `isoformat()` fallback for datetime-like objects, shipped in kaleido v1.4.0 on 2026-08-31. The fix itself is theirs; what this PR adds is the verification and the lift.

Because it sits in the central serializer rather than at a call site, it covers the paths the plotter-side workaround could not reach, which is why #942 reverted `_normalize_x_data` and constrained the dependency instead.

### Verification

**1. The call sites, 1.3.0 versus 1.4.0.** Same script, same plotly 7.1.0 and pandas 3.0.5, only kaleido differs:

| Figure content | kaleido 1.3.0 | kaleido 1.4.0 |
| --- | --- | --- |
| quantile-fill band (the repro in #938) | `TypeError: Type is not JSON serializable: Timestamp` | exports |
| `add_vrect` with Timestamp bounds | same `TypeError` | exports |
| `layout.xaxis.range` with Timestamps | same `TypeError` | exports |
| annotation anchored at a Timestamp | same `TypeError` | exports |

The last three are the call sites #938 lists as not covered by the workaround.

**2. The real plotter.** `ForecastTimeSeriesPlotter` with measurements, a forecast and P10/P50/P90 bands, exported through `pio.to_image`, succeeds on 1.4.0 (60054 bytes). It also succeeds on 1.3.0, because `_add_single_quantile_polygon` already keeps `x` as a `DatetimeIndex`. That workaround is left untouched: it costs nothing and keeps the export robust against future regressions, as the issue describes.

**3. The docs build.** `docs-check` run on my fork against this branch: `uv sync --frozen` then `poe docs`, green on ubuntu-latest, 24.7 MB of HTML uploaded. `nb_execution_mode` is `cache` and a fresh runner has no cache, so the tutorials were executed and their figures exported through kaleido 1.4.0 rather than replayed.

### Why `>=1.4` and not `>=1.0`

1.0 through 1.3 still carry the Timestamp bug, so the floor has to be the release containing plotly/Kaleido#461.

This also resolves the other two reasons the pin mentioned: v1 is the maintained line, so the deprecated v0 goes away, and 1.4.0 ships a single `py3-none-any` wheel, so the missing macOS arm64 wheel that ruled out `0.2.1.post1` is no longer a factor. Note that v1 does not bundle a browser and uses a system Chrome or Chromium, which the CI runner already has, as the green docs build shows.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Refactor / chore / CI

## Breaking changes checklist

Not applicable: this only moves a docs/tutorial build dependency. No library code changes.

## AI disclosure

- [x] AI assistance was used — tool(s): Claude Opus 5, as a research and drafting aid under my direction
  - [x] This is disclosed in a commit message (`Assisted-by: Claude Opus 5`)

## Checklist

- [ ] `poe all --check` passes locally
- [ ] Tests added/updated for the change
- [ ] Documentation updated (docstrings, user guide, examples) if needed
- [x] Commits are signed off per our DCO (`git commit -s`)

On the two unchecked boxes: there is no code change to test, and the upstream behaviour is already covered by `test_orjson_encoder.py` in plotly/Kaleido#461. I ran the docs build rather than `poe all --check`, since that is the job this pin affects; happy to run the full check if you would like it on the record.


docs/pyproject.toml
```diff
@@ -23,13 +23,10 @@ dependencies = [
...
-  # Constrained to v0.2.x because kaleido >=1.0 fails to JSON-serialize
-  # pandas Timestamp values in figure specs (orjson default raises TypeError),
-  # 0.4.x is incompatible with the current plotly scope API, and 0.2.1.post1
-  # only ships a linux/armv7 wheel. v0 is deprecated upstream but functional.
-  # See OpenSTEF/openstef#938; lift the constraint once kaleido v1 handles
-  # pandas Timestamp natively.
-  "kaleido>=0.2.1,<0.3,!=0.2.1.post1",
+  # kaleido 1.4.0 serializes pandas Timestamp values in figure specs
+  # (plotly/Kaleido#461), which is what kept us on the deprecated v0.2.x.
+  # See OpenSTEF/openstef#938.
```
egordm: If the issue has been fixed, we no longer need these comments. We can delete them.

examples/pyproject.toml
```diff
@@ -21,10 +21,9 @@ dependencies = [
 ]
 optional-dependencies.tutorials = [
   "jupyter>=1.1.1",
-  # Constrained to v0.2.x; kaleido >=1.0 cannot serialize pandas Timestamp
-  # via orjson, 0.4.x breaks the plotly scope API, and 0.2.1.post1 ships only
-  # a linux/armv7 wheel. See OpenSTEF/openstef#938.
-  "kaleido>=0.2.1,<0.3,!=0.2.1.post1",
+  # kaleido 1.4.0 serializes pandas Timestamp values in figure specs
+  # (plotly/Kaleido#461). See OpenSTEF/openstef#938.
```
egordm: Same here.

egordm (commented):
Nice find! Overall looks great.

Just a note to remove the comments that are no longer necessary since we no longer pin a specific version.

I will run some final checks tomorrow and merge.

docs/pyproject.toml
Blizzeq: Done in 9dd7ee6d, both comment blocks are gone. The reasoning stays in #938 and in the commit message, so the dependency line reads like any other one now.

examples/pyproject.toml
Blizzeq: Removed here too, in the same commit.

egordm (approved):
Looks great. Approved. Thanks for your contribution!

